PR #2, opened 2025-09-19
Revamp CSS for light-dark() function, fixing Safari rendering in the process

mdo:


src/pierre-js/style.css
```diff
@@ -1,19 +1,21 @@
...
-  --pjs-bg: white;
-  --pjs-fg: black;
+  --pjs-surface: #f5f5f7;
+  --pjs-color: #111;
 }
 
 [data-pjs] {
-  /* These colors inherit from themed colors provided from the Shiki theme, and
-   * then build up colors off of them */
-  --pjs-bg: var(--pjs-light-bg, var(--pjs-bg));
```
amadeus, line 9: So this isn't quite right. We need to use `pjs-light-bg` and `pjs-dark-bg` and `pjs-light` and `pjs-dark` because those are the styles that are set from Shiki

Image

We can change the `wrapped` colors `--pjs-bg` and `--pjs-fg` but the underlying stuff would require us doing a lot more work on shiki itself.

Basically there's a few different formats from shiki that get set through all of this, whether you have a single theme applied, or whether you provide a dark/light theme

amadeus: I'm realizing there's also some font-weight stuff that i'll need to make sure I apply too

mdo: I reworked this now. I _think_ it matches the behavior on the deployed preview site? This 1) more easily responds to color modes with less code, and 2) works in Safari without issue.

src/style.css
```diff
@@ -1,15 +1,18 @@
 :root {
```
mdo, line 1: I'm going to re-audit the style changes here, but do let me know if anything seems off. My understanding from our call Friday is that these styles are basically our placeholder for whatever an end user would do if they built with PierreJS. They would have some globals already defined, and PierreJS would (could?) inherit that (by regular CSS inheritance, applying inline styles/Tailwind, etc etc.

Is that right @amadeus?

src/pierre-js/DiffRenderer.ts
```diff
@@ -208,8 +194,9 @@ export class DiffRenderer {
     if ('themes' in this.options) {
       return {
         themes: this.options.themes,
+        cssVariablePrefix: formatCSSVariablePrefix(),
         lang: this.options.lang ?? ('text' as BundledLanguage),
-        defaultColor: this.options.defaultColor,
+        defaultColor: this.options.defaultColor ?? false,
```
amadeus, line 199: These changes were mostly to ensure some consistency in the default settings of CodeRenderer and DiffRenderer. Takes a bit of an opinionated stance on how `defaultColor` is applied which i think fits better with our autotheme-ing infrastructure.

src/pierre-js/style.css
```diff
@@ -30,27 +31,45 @@
...
+  }
+}
+
+[data-pjs][data-theme='light'] {
+  color-scheme: light;
+
+  [data-column-content] span {
+    color: var(--pjs-light);
+    font-weight: var(--pjs-light-font-weight);
+    font-style: var(--pjs-light-font-style);
```
amadeus, line 59: This `span` stuff was necessary for the `defaultColor` changes I made above

src/pierre-js/style.css
```diff
@@ -30,27 +31,45 @@
...
+      font-style: var(--pjs-dark-font-style);
+    }
+  }
+}
+
+[data-pjs][data-theme='light'] {
+  color-scheme: light;
+
+  [data-column-content] span {
+    color: var(--pjs-light);
```
amadeus: these variables are set via inline styles

src/test_files/index.ts
```diff
@@ -37,18 +37,13 @@ export function toggleTheme() {
...
+        : 'light');
 
-    if (currentTheme === 'light') {
-      pre.dataset.theme = 'dark';
-    } else if (currentTheme === 'dark') {
-      delete pre.dataset.theme;
-    } else {
-      pre.dataset.theme = systemDark ? 'light' : 'dark';
-    }
+    pre.dataset.theme = currentTheme === 'dark' ? 'light' : 'dark';
```
amadeus, line 43: This was some code that I originally had converted by AI and it completely derped lol. So this should now properly work and not require phantom clicks when toggling themes lol

src/main.ts
```diff
@@ -66,7 +66,6 @@ function renderDiff() {
   for (const file of parsedPatch.files) {
     const decorations = DIFF_DECORATIONS[file.name];
     const pre = document.createElement('pre');
-    pre.dataset.theme = 'dark';
```
amadeus, line 69: To improve the default case, making the initial render inherit your preferred theme from the browser, before toggleTheme takes over and applies an override

src/style.css
amadeus: Yeah, this file is just styles for the test page I built, so it can really just be whatever w think looks good aesthetically.

src/style.css
```diff
@@ -144,24 +143,24 @@ code {
 
 [data-pjs] {
   min-height: 0;
-  padding-block: 4px;
+  /* padding-block: 4px; mdo-do: do we need this? bg color doesn't blend which sucks */
```
amadeus: nope, not required, just some styles I did to make things look decent-ish. I'll remove it in a commit

src/test_files/index.ts
mdo: Ayyy was going to ask about this today, so thanks!

PR #4, opened 2025-09-24
Move to individual packages in a bun workspace

SlexAxton:
Should be relatively no code changes. Cleaned up deps. Broke things apart. More can be done to make it good.

- `bun run clean`
- `bun run format`
- `bun run --filter @pierre/js-demo-app dev`
- `bun run --filter @pierre/diff-ui build`

### DIFF
--- a/package.json
+++ b/package.json
@@ -22,6 +22,7 @@
     }
   },
   "scripts": {
+    "prod": "bun run --filter @pierre/js-demo-app prod",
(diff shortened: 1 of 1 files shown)


tsconfig.options.json
```diff
@@ -0,0 +1,35 @@
+{
+  "compilerOptions": {
+    "allowJs": true,
+    "allowSyntheticDefaultImports": true,
```
amadeus, line 4: Is this important?

SlexAxton: copied from monorepo base tsconfig, feel free to modify

PR #5, opened 2025-09-24
Add `bun run dev` & `bun run build` aliases

amadeus:
@SlexAxton adding some aliases

### DIFF
--- a/package.json
+++ b/package.json
@@ -22,9 +22,9 @@
     }
   },
   "scripts": {
-    "prod": "bun run --filter @pierre/js-demo-app prod",
-    "dev": "bun run --filter @pierre/js-demo-app dev",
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -23,6 +23,8 @@
   },
   "scripts": {
     "prod": "bun run --filter @pierre/js-demo-app prod",
+    "dev": "bun run --filter @pierre/js-demo-app dev",
+    "build": "bun run --filter @pierre/js-demo-app build",
```
SlexAxton: Not sure this is what I’d expect build does at the root.

maybe we should just do: `demo:build` and `diff-ui:build` in the fullness if time. And then `build would do them all.

amadeus: Yeah that's a good point, i'll fix!

PR #13, opened 2025-09-26
Per Word Diff 

amadeus:
This PR adds shiki decorations on deletion/addition lines to highlight specific word-based changes:

CleanShot 2025-09-26 at 16 00 30

Because of the way I do this, I've decided to disable the custom shiki decoration API because technically it won't be compatible given that shiki doesn't allow overlapping decorations.

Anyways, at a high level this PR adds `jsdiff` as a dependency and when we iterate over to render we figure out the word-based spans to add highlights for specific changes within a line. I initially used `fast-diff` to do this but it only could diff based on characters which I found a bit more annoying in practice. I think word based is much easier.

I also applied some additional logic to look for cases where there would be single white-space characters between spans and use that to join the spans together to avoid a sort of gap-tooth'd aesthetic.

I think we still need to iterate on what the spans look like and what colors they should adopt, but this is definitely a good foundation for us to build on.

### DIFF
--- a/packages/diff-ui/src/DiffRenderer.ts
+++ b/packages/diff-ui/src/DiffRenderer.ts
@@ -301,6 +301,9 @@ export class DiffRenderer {
     function addToChangeGroup(type: 'addition' | 'deletion', line: string) {
       if (currentChangeGroup == null) {
         currentChangeGroup = {
+          // In unified layout, deletionLineIndex and additionLineIndex won't
+          // be usable, and we will have to compute start indexes as we are
+          // iterating
           deletionStartIndex: unified ? -1 : deletionLineIndex - 1,
           additionStartIndex: unified ? -1 : additionLineIndex - 1,
           deletionLines: [],
@@ -679,6 +682,11 @@ function createTransformerWithState(): {
   };
 }
 
+// For diff decoration spans, we want to be sure that if there is a single
+// white-space gap between diffs that we join them together into a longer diff span.
+// Spans are basically just a tuple - 1 means the content should be
+// highlighted, 0 means it should not, we still need to the span data to figure
+// out span positions
 function pushOrJoinSpan(
   item: ChangeObject<string>,
   arr: [0 | 1, string][],


packages/diff-ui/src/DiffRenderer.ts
```diff
@@ -101,17 +107,11 @@ export class DiffRenderer {
...
-    | [FileMetadata, HTMLPreElement, DiffDecorationItem[] | undefined]
-    | undefined;
+  private queuedRenderArgs: [FileMetadata, HTMLPreElement] | undefined;
 
-  async render(
-    _diff: FileMetadata,
-    _wrapper: HTMLPreElement,
-    _decorations?: DiffDecorationItem[]
-  ) {
+  async render(_diff: FileMetadata, _wrapper: HTMLPreElement) {
```
amadeus, line 112: The changes within the `render` function` here are mostly just to remove the decorations API

packages/diff-ui/src/DiffRenderer.ts
```diff
@@ -232,33 +215,120 @@ export class DiffRenderer {
...
     state,
     transformer,
-    additionDecorations,
-    deletionDecorations,
     codeAdditions,
     codeDeletions,
     codeUnified,
   }: RenderHunkProps) {
     if (hunk.hunkContent == null) return;
+    const { additions, deletions, unified } = this.processLines(hunk);
```
amadeus, line 223: I kinda refactored the layout of `renderHunks` to some sub-functions, but the core logic of rendering into the code elements is largely unchanged.

packages/diff-ui/src/DiffRenderer.ts
```diff
@@ -350,51 +423,92 @@ export class DiffRenderer {
...
+        const additionLine = group.additionLines[i];
+        if (deletionLine == null || additionLine == null) {
+          break;
+        }
+        // Lets skep running diffs on super long lines because it's probably
+        // expensive and hard to follow
+        if (deletionLine.length > 1000 || additionLine.length > 1000) {
+          continue;
+        }
+        const lineDiff = diffWordsWithSpace(deletionLine, additionLine);
```
amadeus, line 473: This is where most of the magic happens -- `diffGroups` is essentially a data structure where we figure out arrays of additions and deletions without the context lines and figure out the line index relative to each group that can then be passed through to the decorations API

packages/diff-ui/src/DiffRenderer.ts
```diff
@@ -434,6 +548,25 @@ export class DiffRenderer {
...
+function createDiffSpanDecoration({
+  line,
+  spanStart,
+  spanLength,
+}: CreateDiffSpanDecorationProps): DecorationItem {
+  return {
+    start: { line, character: spanStart },
+    end: { line, character: spanStart + spanLength },
+    properties: { 'data-diff-span': '' },
+    alwaysWrap: true,
```
amadeus, line 569: This property is important to make sure we never highlight the `whole` line and instead JUST the text even if all the content of the line changed (the decorations need to feel like specific text wrappers, not full line wrappers

packages/demo-app/src/mocks/index.ts
```diff
@@ -65,49 +65,6 @@ export const DIFF_CONTENT_FORMATS: Record<string, BundledLanguage | undefined> =
     patch: 'diff',
   };
 
-export const DIFF_DECORATIONS: Record<string, DiffDecorationItem[]> = {
```
amadeus, line 68: This API is now kill... we can't really support it and highlighting diffs at the same time... which I think is ok in practice

PR #18, opened 2025-10-02
Performance and diff2html API Rework

amadeus:
This PR encapsulates a a bunch of changes that are connected yet disparate.

* Fixed a few bugs with some of the diff file parsing based on Ed's monster PR patch file
* A bunch of performance optimizations for pathologically large diffs
* Started a round of proper configurability to diff rendering
* Started work on a class for handling an entire file (that then wraps the DiffHunksRenderer)

### DIFF
--- a/apps/demo/src/main.ts
+++ b/apps/demo/src/main.ts
@@ -15,9 +15,16 @@ import { CodeConfigs, FILE_NEW, FILE_OLD, toggleTheme } from './mocks/';
 import './style.css';
 import { createFakeContentStream } from './utils/createFakeContentStream';
 
+let loadingPatch: Promise<string> | undefined;
 async function loadPatchContent() {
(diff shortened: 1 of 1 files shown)


apps/demo/src/mocks/index.ts
```diff
@@ -53,25 +45,3 @@ export function toggleTheme() {
 
 export const FILE_OLD = fileOld;
 export const FILE_NEW = fileNew;
-
-export const DIFF_CONTENT = diffContent;
-export const DIFF_CONTENT_2 = diffContent2;
-export const DIFF_CONTENT_3 = diffContent3;
-export const DIFF_CONTENT_4 = diffContent4;
-export const DIFF_CONTENT_5 = diffContent5;
```
amadeus, line 61: Undoing the sins of my past here. No reason to bundle these in with the JS, should `import()` them, which now I do

PR #35, opened 2025-10-10
First attempt at a raw-loader for docs

amadeus:
This seems to work. And it can use the same import in both places.

### DIFF
--- a/packages/diff-ui/package.json
+++ b/packages/diff-ui/package.json
@@ -4,7 +4,7 @@
   "version": "0.0.0",
   "type": "module",
   "scripts": {
-    "build": "tsc --build --force",
+    "build": "tsc --build",
     "dev": "tsc --build --watch",
     "tsc": "tsc"
   },
--- a/packages/shiki-stream/package.json
+++ b/packages/shiki-stream/package.json
@@ -4,7 +4,7 @@
   "version": "0.0.0",
   "type": "module",
   "scripts": {
-    "build": "tsc --build --force",
+    "build": "tsc --build",
     "dev": "tsc --build --watch"
   },
   "exports": {
--- a/packages/sync-ui-shadcn/package.json
+++ b/packages/sync-ui-shadcn/package.json
@@ -5,7 +5,7 @@
   "type": "module",
   "scripts": {
     "tsc": "tsc --build",
-    "build": "tsc --build --force"
+    "build": "tsc --build"
   },
   "dependencies": {
     "@radix-ui/react-dialog": "catalog:",


apps/docs/app/api/code-storage/repo/route.ts
```diff
@@ -1,13 +1,15 @@
...
-});
+export const dynamic = 'force-dynamic';
 
 export async function POST(request: NextRequest) {
   try {
+    const store = new GitStorage({
+      name: 'pierre',
+      key: process.env.CODE_STORAGE_SYNC_PRIVATE_KEY || '',
+    });
+
```
amadeus, line 10: the docs build command was failing with this at the module level, unclear if that's a safe change to make or not, but I figured it would be fine to do?

packages/diff-ui/package.json
```diff
@@ -4,7 +4,7 @@
   "version": "0.0.0",
   "type": "module",
   "scripts": {
-    "build": "tsc --build",
+    "build": "tsc --build --force",
```
amadeus: Mostly added these because I would notice myself getting into weird incremental build states

apps/docs/next.config.mjs
```diff
@@ -14,6 +14,36 @@ const nextConfig = {
...
+        loaders: ['raw-loader'],
+        as: '*.js',
+      },
+      '../../packages/diff-ui/src/style.css': {
+        loaders: ['raw-loader'],
+        as: '*.js',
+      },
+    },
+  },
+  // NOTE(amadeus): Pretty sure we don't need this, but keeping it around just
```
amadeus, line 29: This is some slop that claude added, keeping it around in case something required it, but pretty sure it wasn't needed

apps/docs/app/api/code-storage/repo/route.ts
```diff
@@ -1,13 +1,15 @@
 import { GitStorage } from '@pierre/storage';
 import { NextRequest, NextResponse } from 'next/server';
 
-const store = new GitStorage({
-  name: 'pierre',
-  key: process.env.CODE_STORAGE_SYNC_PRIVATE_KEY || '',
-});
+export const dynamic = 'force-dynamic';
```
SlexAxton: Interesting that this helps.

apps/docs/app/api/code-storage/repo/route.ts
SlexAxton: we should more generally ban Jacob style `|| ''` but i think we agree on that

apps/docs/app/api/code-storage/repo/route.ts
amadeus: honestly -- not sure if it's needed either lol. it might have been claude's first try to fix, which didn't work. I can try undoing it

apps/docs/app/api/code-storage/repo/route.ts
SlexAxton: I wonder if it’s like a top level async problem

packages/diff-ui/package.json
SlexAxton: hm, maybe check the various project references and run a clean. but seems ok since it’s a check and not a real build

apps/docs/app/api/code-storage/repo/route.ts
amadeus: it did not matter for building, so worth not mucking around with other stuff to get builds to pass

apps/docs/app/api/code-storage/repo/route.ts
amadeus: I would be very happy to add some stricter typescript aware eslint rules to this project :feelsgood:

(obvi for another PR tho)

PR #37, opened 2025-10-10
Added an example embedded diff

amadeus:
A quick hack to make a FileDiff component that's essentially a hacked wrapper around the diff rendering logic.

This is not my finest work, but it's functional to allow us to move forward.


apps/docs/app/diff-examples/SplitUnified.tsx
```diff
@@ -0,0 +1,82 @@
...
+        </Button>
+        <Button
+          variant={diffStyle === 'unified' ? 'active' : undefined}
+          onClick={() => setDiffStyle('unified')}
+        >
+          <IconDiffBlended />
+          Stacked
+        </Button>
+      </ButtonGroup>
+      <FileDiff
```
amadeus, line 71: @mdo @fat this is basically an example of how to use this component right now

PR #58, opened 2025-10-16
SSR Implementation with Declarative Shadow Dom

SlexAxton:


packages/precision-diffs/src/ssr/FileDiffElement.ts
```diff
@@ -0,0 +1,39 @@
...
+      super();
+    }
+  }
+
+  if (customElements.get('file-diff') === undefined) {
+    customElements.define('file-diff', FileDiffElement);
+  }
+}
+
+export const FileDiffElementHasLoaded = true;
```
amadeus, line 13: Does this actually do anything? or is this something specialized for lit or something?

packages/precision-diffs/src/ssr/FileDiffElement.ts
```diff
@@ -0,0 +1,39 @@
+if (typeof HTMLElement !== 'undefined') {
+  class FileDiffElement extends HTMLElement {
+    constructor() {
+      super();
+    }
+  }
+
+  if (customElements.get('file-diff') === undefined) {
```
amadeus, line 8: nit, but can be `== null` if you want less verbosity 

I assume this was to get around some of the double registration stuff you were worried about earlier?

packages/precision-diffs/src/ssr/FileDiffElement.ts
SlexAxton: I will check, but for a bit it was mad at me that `FileDiffElement` was `not a module` (likely because the module definition is in an if-statement.

SlexAxton: (and so it wouldn't let me import the file, which has desirable side effects)

packages/precision-diffs/src/ssr/FileDiffElement.ts
SlexAxton: wouldn't that be against your goals here? `customElements` is a map that is either `undefined` or has a custom element in it. We could handle the case where it's nullable… but at that point why not just check truthiness?

apps/docs/tsconfig.json
```diff
@@ -12,11 +12,14 @@
...
       "path": "../../packages/precision-diffs"
+    },
+    {
+      "path": "../../packages/storage-elements-next"
     }
   ],
   "compilerOptions": {
     "target": "ES2017",
-    "lib": ["dom", "dom.iterable", "esnext"],
+    "lib": ["ES2023", "DOM", "DOM.Iterable"],
```
SlexAxton: i could probably kill most of these things, so i will.

packages/precision-diffs/src/DiffHunksRenderer.ts
```diff
@@ -248,7 +301,8 @@ export class DiffHunksRenderer<LAnnotation = undefined> {
 
   private renderDiff(
     fileDiff: FileDiffMetadata,
-    highlighter: PJSHighlighter
+    highlighter: PJSHighlighter,
+    shouldUseClasses: boolean = false
```
amadeus, line 305: I can do this refactor merge, but probably worth making this part of the options for the class

packages/precision-diffs/src/DiffHunksRenderer.ts
```diff
@@ -101,9 +102,54 @@ export interface HunksRenderResult {
   additionsAST: ElementContent[] | undefined;
   deletionsAST: ElementContent[] | undefined;
   unifiedAST: ElementContent[] | undefined;
+  css: string;
   preNode: Element;
 }
 
+// Create a transformer that converts token color/fontStyle to htmlStyle
+// This needs to run BEFORE transformerStyleToClass
+const tokenStyleNormalizer: ShikiTransformer = {
```
amadeus, line 111: Also something I can refactor after merge -- shove this into the function that creates the line transformer

packages/precision-diffs/src/ssr/FileDiffServer.ts
```diff
@@ -0,0 +1,726 @@
...
+
+  return {
+    suppressHydrationWarning: true,
+    dangerouslySetInnerHTML: {
+      __html: prerenderedHTML,
+    },
+  } satisfies PreloadedFileDiffResult;
+}
+
+const rawStyles = `:host {
```
amadeus, line 66: Should we make this import the raw text from style.css so we don't have to worry about keeping 2 things in sync?

packages/precision-diffs/src/ssr/FileDiffServer.ts
```diff
@@ -0,0 +1,726 @@
...
+}: PreloadFileDiffOptions<LAnnotation>) {
+  const fileDiff = parseDiffFromFile(oldFile, newFile);
+  const diffHunksRenderer = new DiffHunksRenderer<LAnnotation>(options);
+  const result = await diffHunksRenderer.render(fileDiff, true);
+  if (result == null) {
+    throw new Error('Failed to render file diff');
+  }
+  const html = diffHunksRenderer.renderFullHTML(result);
+
+  const prerenderedHTML = `<template shadowrootmode="open">
```
amadeus, line 44: I can probably make a version of this that uses the hast stuff, so then you can take the result and wrap everything with this and then have that render everything out. I figured we should lean into hast stuff for html generation since it's probably safest that way

packages/precision-diffs/src/ssr/FileDiffElement.ts
amadeus: Oh I see, makes sense!

packages/precision-diffs/src/ssr/FileDiffElement.ts
amadeus: the eslint thing I put in allows == or != checks for existence (I.E. it allows coercion of null/undefined-ness, but no other). I just use == null or != null to check existence generally (it won't convert an empty array or the number 0). Ultimately just a nit tho, nbd

packages/precision-diffs/src/ssr/FileDiffServer.ts
SlexAxton: yea, if we can make it work. it started getting mad at me and so i just did this as a stop gap to not block on figuring out the tooling yet.

packages/precision-diffs/src/ssr/FileDiffServer.ts
SlexAxton: sg

PR #78, opened 2025-10-20
Add Vibrant themes, make default

mdo:
Replacing the dark/light with vibrant versions that take advantage of P3 color space in CSS.


apps/docs/app/diff-examples/ShikiThemes.tsx
```diff
@@ -82,6 +83,7 @@ const LIGHT_THEMES = [
 
 const DARK_THEMES = [
   'pierre-dark',
+  'pierre-dark-vibrant',
```
amadeus, line 86: total nit, but if they are default, mb move them to the top of the array?

PR #81, opened 2025-10-21
comment ui

fat:


apps/docs/app/diff-examples/Annotations.tsx
```diff
@@ -42,24 +44,190 @@ export default function Home() {
...
+  const handleCancelComment = useCallback(() => {
+    setActiveCommentForm(null);
+  }, []);
+
+  const annotations = [
+    ...Array.from(commentThreads).map((key) => ({
+      key,
+      side: key.split('-')[0] as 'additions' | 'deletions',
+      lineNumber: parseInt(key.split('-')[1], 10),
+      isThread: true,
```
amadeus: `isThread` and `key` should be stuffed into the `metadata`

apps/docs/app/diff-examples/Annotations.tsx
```diff
@@ -42,24 +44,190 @@ export default function Home() {
 `,
 };
 
+type LineKey = `${string}-${number}`;
+
 export function Annotations() {
+  const [commentThreads, setCommentThreads] = useState<Set<LineKey>>(
```
amadeus: Why not do something like: 

```ts
interface AnnotationMetadata {
  isThread: boolean;
  // any other data you might want to stuff in here
}

const [annotations, setAnnotation] = useState<DiffLineAnnotation<AnnotationMetadata>>(() => [{
  side: 'additions',
  lineNumber: 8,
  metadata: {
    isThread: true,
    // not sure if you need a key, also you can have multiple annotations that
    // target the same line technically, so `additions-8` is not technically a key
  }
}]);
```
Then for hover state, you'll get side and line number as data, so just hold onto that directly and just push them the array via setAnnotation on the click handler

apps/docs/app/diff-examples/Annotations.tsx
```diff
@@ -42,24 +44,190 @@ export default function Home() {
...
+  );
+  const [activeCommentForm, setActiveCommentForm] = useState<LineKey | null>(null);
+  const [buttonPosition, setButtonPosition] = useState<{
+    top: number;
+    left: number;
+  } | null>(null);
+  const [hoveredLineKey, setHoveredLineKey] = useState<LineKey | null>(null);
+  const containerRef = useRef<HTMLDivElement>(null);
+
+  const handleLineEnter = useCallback((props: any) => {
```
amadeus: I do have some ideas for this, but haven't gotten around to it.  I think ideally it would be some API like `renderMouseOverDecoration` or something, and you'd just get a render function to render a component like a button, and in that render function you'd get an object of the line number and side. and so then automatically on mouse over we'd call that render function and render that wee lad

packages/precision-diffs/src/FileDiff.ts
```diff
@@ -310,11 +310,45 @@ export class FileDiff<LAnnotation = undefined> {
   }
 
   spriteSVG: SVGElement | undefined;
+  private eventListenersAttached = false;
+
+  private attachEventListeners() {
```
amadeus, line 319: Ideally this wouldn't be necessary, I might have to poke around and fix this

PR #83, opened 2025-10-21
Misc Fixes and Improvements

amadeus:
* Fixed an annotation rendering bug /cc @SlexAxton 
* Various SSR API improvements
  * DiffHunkRenderer now has more flexibility in what it can do with results
* Some prep for using `hast` on the server to render out some of the other things we'll need
* Ensure annotations are at least 1lh by default

### DIFF
--- a/packages/precision-diffs/src/style.css
+++ b/packages/precision-diffs/src/style.css
@@ -13,6 +13,7 @@
 
   --pjs-mixer: light-dark(black, white);
 
+  --pjs-annotation-min-height: 1lh;
   /*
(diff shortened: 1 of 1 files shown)


packages/precision-diffs/src/DiffHunksRenderer.ts
```diff
@@ -704,6 +713,7 @@ export class DiffHunksRenderer<LAnnotation = undefined> {
           );
         }
       }
+      resolveUnresolvedSpans();
```
amadeus, line 716: This should be the fix for annotations

PR #84, opened 2025-10-21
SSR Hax

amadeus:
Made some tweaks to some of the SSR stuff to better utilize hast for rendering stuff.

### DIFF
--- a/apps/docs/app/ssr/page.tsx
+++ b/apps/docs/app/ssr/page.tsx
@@ -1,4 +1,7 @@
-import { type FileContents } from '@pierre/precision-diffs';
+import {
+  type DiffLineAnnotation,
+  type FileContents,
+} from '@pierre/precision-diffs';
 import { preloadFileDiff } from '@pierre/precision-diffs/ssr';
 
 import { SsrPage } from './SsrPage';
@@ -41,9 +44,9 @@ pub fn main() !void {
 };
 
 // Define annotation positions once in the server component
-const annotationPositions = [
+const annotationPositions: DiffLineAnnotation[] = [
   {
-    side: 'additions' as const,
+    side: 'additions',
     lineNumber: 8,
   },
 ];


packages/precision-diffs/src/sprite.ts
```diff
@@ -1,4 +1,4 @@
-<svg data-icon-sprite aria-hidden="true" width="0" height="0">
+export const SVGSpriteSheet = `<svg data-icon-sprite aria-hidden="true" width="0" height="0">
   <symbol id="pjs-icon-diff-split">
```
SlexAxton, line 2: fyi, `id` usage in a re-used svg is a little bit of a recipe for very confusing bugs down the line. much like ids in html, they're supposed to be globally unique, so when you apply a gradient to a `#pjs-icon-diff-split` or something, it ends up applying to some random other instance of the icon on the page, internally from some other svg.

But seems like we aren't using this ID, so should be simple to just kill it?

amadeus: this is rendered in the shadow dom, so ID's should be encapsulated? that said, there's a valid question of whether multiple instances of the diffs have access to these same IDs. however in practice I think it was a non-issue since if they are shared across the same container instances, then they'd just be a duplication and the icons that reference these IDs would ultimately be referring to the same svg shapes

SlexAxton: this particular svg isn't rendered more than once in the same diff?

SlexAxton: yea im not saying it'd be a problem now, but later if we like… applied a filter to one svg directly, it may apply to some other first rendered svg on the page. usually when you're building svgs for this kind of stuff you'll run all svgs through something that can guarantee uniqueness - this is default on in svgo for instance https://svgo.dev/docs/plugins/cleanupIds/

amadeus: It should not be rendered more than once since it's supposed to be like a library of icons, and it's always hidden, as it's simply a reference for _other_ SVGs to render out whatever that symbol is based on the id.

In other words, it should only be 1 of these rendered per `<pjs-container>` or `<file-diff>` shadow dom

PR #91, opened 2025-10-23
SSR Continuance

amadeus:
Mostly the intention of this branch was to get the FileDiff.tsx working with serverside rendering from @SlexAxton.

### Part 1

The first thing I focused on in this PR was getting the FileDiff.tsx properly hydrating from the server. The key thing I was trying to make sure we could do is manage a normie react lifecycle for things like annotations, metadata and future updates to the options of the component. Mostly that involved some trickery with server rendered state and the initial load on client.

After that we hydrate the FileDiffUI class instance with the rendered content, which involves setting up element references, hooking up event listeners and resize observers.

I added both a changeable option (diffStyle) and the ability to dynamically add annotations by clicking on the line.

### Part 2

After I got this all working, next step was to converge the differing custom components under one (I liked `file-diff` over `pjs-container` anyways) and then fixed up types. There were also a series of small bugs with regards to icons and icon colors that I tweaked into shape.

### Part 3

Finally I took a pass on the SSR test page and brought back the original static demo that Alex build and converged APIs and annotations to match. I also added headers and icons into the static version and fixed up some out of date CSS.

Anyways, I think this version is ready to merge as a foundation we can both work off of as we continue to flesh out the components.

### DIFF
--- a/apps/docs/app/ssr/page.tsx
+++ b/apps/docs/app/ssr/page.tsx
@@ -5,7 +5,7 @@ import {
 import { preloadFileDiff } from '@pierre/precision-diffs/ssr';
 
 import { SsrPage } from './SsrPage';
-import type { AnnotationMetadata } from './ssr_types.ts';
+import type { AnnotationMetadata } from './ssr_types';
 
 const OLD_FILE: FileContents = {
   name: 'main.zig',


apps/docs/app/diff-examples/Annotations.tsx
```diff
@@ -199,7 +194,6 @@ export function Annotations() {
             theme: 'pierre-dark',
             diffStyle: 'unified',
             onLineEnter: handleLineEnter,
-            onLineLeave: handleLineLeave,
```
amadeus, line 202: I still need to take a pass at this API, after I tweaked things up, I needed to remove this API to prevent the icon from flickering. Part of the actual issue here is that because the element isn't slotted in, it fires a mouse leave event...

apps/docs/app/ssr/SsrPage.tsx
```diff
@@ -92,15 +84,66 @@ export function SsrPage({
...
+      <div className="flex flex-col gap-20">
+        <div>
+          <h2 className="text-2xl font-medium tracking-tight md:text-2xl">
+            Static Test
+          </h2>
+          <FileDiffSsr<AnnotationMetadata>
+            prerenderedHTML={preloadedFileDiff.prerenderedHTML}
+            className="overflow-hidden rounded-lg border"
+            annotations={preloadedFileDiff.annotations}
+            renderAnnotation={renderAnnotation}
```
amadeus, line 99: Now both the SSR version and dynamic version both use the same APIs for annotation data and rendering annotations

apps/docs/app/ssr/SsrPage.tsx
```diff
@@ -92,15 +84,66 @@ export function SsrPage({
...
-        className="overflow-hidden rounded-lg border"
-        annotations={annotations}
-      />
+      <div className="flex flex-col gap-20">
+        <div>
+          <h2 className="text-2xl font-medium tracking-tight md:text-2xl">
+            Static Test
+          </h2>
+          <FileDiffSsr<AnnotationMetadata>
+            prerenderedHTML={preloadedFileDiff.prerenderedHTML}
```
amadeus, line 96: I've simplified this prop into just a string. I think it's less susceptible to blowing memoization, and just feels simpler to orchestrate if you're doing your own thing with prerendering

apps/docs/app/ssr/SsrPage.tsx
```diff
@@ -92,15 +84,66 @@ export function SsrPage({
...
+              onClick={() =>
+                setDiffStyle(diffStyle === 'split' ? 'unified' : 'split')
+              }
+            >
+              Toggle Diff Style
+            </button>
+          </div>
+          <FileDiff<AnnotationMetadata>
+            {...preloadedFileDiff}
+            // @ts-expect-error lol
```
amadeus, line 119: I need to probably think about this a bit more. Shiki is pretty strict about how you pass a theme property to it. Either it's `theme: themeName` or `themes: {dark/light}`. This makes stuff like this far less flexible. Wondering if I should just make the component more permissible and just resolve things internally

apps/docs/components/diff-ui/FileDiff.tsx
```diff
@@ -39,25 +40,39 @@ export function FileDiff<LAnnotation = undefined>({
...
     const forceRender = !deepEqual(instanceRef.current.options, options);
     instanceRef.current.setOptions(options);
-    void instanceRef.current.render({
-      forceRender,
-      oldFile,
-      newFile,
-      fileContainer: ref.current ?? undefined,
-      lineAnnotations: annotations,
-    });
+    if (firstRender && prerenderedHTML != null) {
```
amadeus, line 57: This is now the hydration step for FileDiff. It requires similar data, but then ends up setting up everything for further re-renders.

This also opens up the possibility of me removing FileDiffUI completely and just coordinating some classes or a custom class specifically for react (and not mux it with the vanilla api

apps/docs/components/diff-ui/FileDiff.tsx
```diff
@@ -67,18 +82,36 @@ export function FileDiff<LAnnotation = undefined>({
...
+    </>
+  );
+  return (
+    <file-diff ref={ref} className={className} style={style}>
+      {templateRender(children, prerenderedHTML)}
+    </file-diff>
   );
 }
+
+function templateRender(children: ReactNode, __html?: string) {
```
amadeus, line 100: This is really the secret sauce of this component. It elegantly handles serverside rendering with custom shadowdom html, and cleanly allows hydration of annotations/metadata etc. It's probably worth porting some of these ideas over to the SSR FileDiff component. The main thing is that it doesn't need to instantiate the FileDiffUI class.

it also has the added benefit of no longer needing to suppress hydration warnings, it's that clean

apps/docs/components/diff-ui/FileDiff.tsx
```diff
@@ -67,18 +82,36 @@ export function FileDiff<LAnnotation = undefined>({
...
+  if (typeof window === 'undefined' && __html != null) {
+    return (
+      <>
+        <template
+          // @ts-expect-error unclear how to fix this
+          shadowrootmode="open"
```
amadeus, line 106: I just don't know how to fix this error unfortunately. React does not like this prop and I couldn't figure out how to override intrinsic definitions

packages/precision-diffs/src/custom-components/Container.ts
```diff
@@ -2,27 +2,33 @@ import styles from '../style.css?raw';
...
+      const shadowRoot = this.attachShadow({ mode: 'open' });
       if (sheet == null) {
         sheet = new CSSStyleSheet();
         sheet.replaceSync(styles);
       }
       // TODO(amadeus): Figure out how to adapt user generated styles into here?
-      if (this.shadowRoot?.adoptedStyleSheets != null) {
-        this.shadowRoot.adoptedStyleSheets = [sheet];
-      }
+      shadowRoot.adoptedStyleSheets = [sheet];
```
amadeus, line 25: We sorta have 2 versions of this component -- one for SSR and one here. Wondering if there's a correct way we can converge them.

This gets a bit into how to handle the CSS question. Current SSR just inlines raw text and does some stuff, but it would be great to have both pulling from the same data source as a couple bugs I fixed where due to outdated CSS.

packages/precision-diffs/src/ssr/FileDiffReact.tsx
```diff
@@ -75,18 +78,20 @@ export function FileDiffSsr({
...
               // Render the component with React markers for proper hydration
-              const content = renderToString(render());
+              const content = renderToString(renderAnnotation(annotation));
               return `<div slot="${slotName}">${content}</div>`;
             })
             .join('')
         : '';
 
-    const fullHTML = `<file-diff${classAttr}${styleAttr}>${preloadedFileDiff.dangerouslySetInnerHTML.__html}${annotationSlots}</file-diff>`;
+    const fullHTML = `<file-diff${classAttr}${styleAttr}><template shadowrootmode="open">${prerenderedHTML}</template>${annotationSlots}</file-diff>`;
```
amadeus, line 94: I have the feeling we could probably simplify this component to work a bit like the other FileDiff.tsx just without instantiating the JS that includes shiki and co. and then have similar rendering paths that won't require manual hydration.

packages/precision-diffs/src/ssr/FileDiffServer.ts
```diff
@@ -759,6 +756,7 @@ code {
 }
 
 [data-change-icon] {
+  fill: currentColor;
```
amadeus, line 761: This was def a cheeky bug for me, I thought something was broken with the CSS on the SSR version, turned out it was just out of date from some of our icon generation stuff, so definitely something we should figure out -- how to make sure there's 1 true source of CSS for all versions

apps/docs/app/ssr/page.tsx
```diff
@@ -5,6 +5,7 @@ import {
 import { preloadFileDiff } from '@pierre/precision-diffs/ssr';
 
 import { SsrPage } from './SsrPage';
+import type { AnnotationMetadata } from './ssr_types.ts';
```
SlexAxton: probably dont need the .ts

packages/precision-diffs/src/ssr/FileDiffReact.tsx
```diff
@@ -108,19 +113,19 @@ export function FileDiffSsr({
...
-          const annotation = annotations.find(({ line, side }) => {
-            return slotName === `annotation-${side}-${line}`;
+          const annotation = annotations.find((annotation) => {
+            return slotName === getLineAnnotationId(annotation);
           });
 
-          if (annotation !== undefined) {
-            hydrateRoot(slotElement, annotation.render());
+          if (annotation !== undefined && renderAnnotation != null) {
+            hydrateRoot(slotElement, renderAnnotation(annotation));
```
amadeus, line 121: The main change I made in this file was simply to have `renderAnnotation` api's be identical between this and FileDiff

apps/docs/app/ssr/page.tsx
amadeus: oof, autocomplete derp probably, will fix

packages/precision-diffs/src/ssr/FileDiffServer.ts
SlexAxton: course

PR #95, opened 2025-10-24
Move react components into precision diffs library

amadeus:
Trying to osmose some @SlexAxton here, I think it's working and proper

### DIFF
--- a/packages/precision-diffs/package.json
+++ b/packages/precision-diffs/package.json
@@ -3,7 +3,7 @@
   "version": "0.0.2-alpha.1-1",
   "type": "module",
   "scripts": {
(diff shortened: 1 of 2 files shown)


packages/precision-diffs/tsconfig.json
```diff
@@ -7,7 +7,7 @@
     "baseUrl": ".",
     "rootDir": "src",
     "lib": ["ES2023", "DOM", "DOM.Iterable"],
-    "jsx": "jsx-react",
+    "jsx": "react-jsx",
```
SlexAxton, line 10: Lol sorry

amadeus: the classic ruse

PR #97, opened 2025-10-25
Make `bun run publish` do all the publish work in precision-diffs

SlexAxton:
also added protection from accidentally running `bun publish`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,47 +17,10 @@ bun run dev
 
 ## Publishing precision diffs
 
-Note that publishing precision diffs is a bit artisinal.
-
-**This is dangerous.**
-
-First run `bun run clean` at root then `cd` into the package directory
-
-```bash
-bun run clean
+```sh
 cd packages/precision-diffs
+bun run publish
 ```
-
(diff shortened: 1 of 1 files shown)


packages/precision-diffs/package.json
```diff
@@ -1,11 +1,13 @@
...
+  "version": "0.0.2-alpha.1-4",
   "type": "module",
   "scripts": {
-    "build": "tsc --build && node post-build.js",
+    "build": "tsc --build && bun post-build.js",
     "dev": "tsc --build --watch",
-    "tsc": "tsc"
+    "tsc": "tsc",
+    "prepublishOnly": "echo 'Error: Do not run bun publish directly. Use: bun run publish' && exit 1",
+    "publish": "bun run build && bun prepare-publish.js && cd dist && bun publish"
```
amadeus, line 10: what's the `cd dist` do with this command?

SlexAxton: it publishes from the dist folder not the root.

PR #99, opened 2025-10-26
Fat/cursor rip

fat:


packages/precision-diffs/src/style.css
```diff
@@ -13,7 +13,7 @@
 
   --pjs-mixer: light-dark(black, white);
 
-  --pjs-annotation-min-height: 1lh;
+  --pjs-annotation-min-height: 0;
```
amadeus, line 16: Do we even need this variable anymore?

packages/precision-diffs/src/types.ts
```diff
@@ -110,6 +110,7 @@ export interface BaseDiffProps extends BaseCodeProps {
   diffIndicators?: 'classic' | 'bars' | 'none'; // bars is default
   disableBackground?: boolean;
   hunkSeparators?: HunkSeparators; // line-info is default
+  collapseUnchanged?: boolean; // true collapses unchanged hunks
```
amadeus: Stylistically I'm not a fan of a boolean that defaults to true. Can we invert this and just call it `expandUnchanged` or something? I don't like APIs where NOT filling it out results in a `truth`y value.

It also matches the other booleans that we have in this API like `disableBackgrounds` that default to `false`

packages/precision-diffs/src/DiffHunksRenderer.ts
```diff
@@ -136,6 +136,17 @@ export class DiffHunksRenderer<LAnnotation = undefined> {
     this.expandedHunks.add(index);
   }
 
+  private shouldCollapseUnchanged() {
+    return this.getOptionsWithDefaults().collapseUnchanged;
+  }
```
amadeus: Since this has been added `getOptionsWithDefaults`, can we just remove this function and deconstruct it it from `getOptionsWithDefaults` wherever we need it?

packages/precision-diffs/src/DiffHunksRenderer.ts
```diff
@@ -306,6 +320,37 @@ export class DiffHunksRenderer<LAnnotation = undefined> {
...
+      this.diff?.newLines != null &&
+      this.diff.newLines.length > 0
+    ) {
+      const lineCount = this.diff.newLines.length;
+      const syntheticHunk: Hunk = {
+        additionCount: lineCount,
+        additionStart: 1,
+        deletedCount: lineCount,
+        deletedStart: 1,
+        hunkContent: this.diff.newLines.map((line) => ` ${line}`),
```
amadeus: We we even need this map? can we just pass this.diff.newLines directly?

packages/precision-diffs/src/style.css
amadeus: We do in fact need this variable and it is important as I found a bug that we removed it from being used elsewhere. So disregard this.

packages/precision-diffs/src/types.ts
amadeus: Fixed in my follow commits

packages/precision-diffs/src/DiffHunksRenderer.ts
amadeus: Fixed in my follow-up commits

packages/precision-diffs/src/DiffHunksRenderer.ts
amadeus: oh, we do need this cause of the space, ugh

amadeus: Nevermind, i figured out a way around this, see 6fd001db42d7d0509eca12de54798039869f4b38

PR #105, opened 2025-10-28
Allow some layout css variables to be overridden

amadeus:
Kindof annoyingly, we just can't set these in host, and then have to manually allow fallbacks in all the call sites... blech.

Basically this PR is in response to Anomaly asking to be able to customize the `--pjs-gap` value.

I ended up getting rid of gap and instead having `--pjs-gap-block` and `--pjs-gap-inline` to keep them separate.

### DIFF
--- a/apps/docs/app/docs/Styling/constants.ts
+++ b/apps/docs/app/docs/Styling/constants.ts
@@ -28,6 +28,15 @@ export const STYLING_CODE_GLOBAL: PreloadFileOptions<undefined> = {
   --pjs-deletion-color-override: orange;
   --pjs-addition-color-override: yellow;
   --pjs-modified-color-override: purple;
+
+  /* Some basic variables for tweaking the layouts of some of the built in
+   * components */
+  --pjs-gap-inline: 8px;
+  --pjs-gap-block: 8px;
+
+  /* If you'd like to adjust the size of tab characters,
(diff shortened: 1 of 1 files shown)


packages/precision-diffs/src/style.css
```diff
@@ -1,9 +1,6 @@
 :host {
   --pjs-bg: #fff;
   --pjs-fg: #000;
-  --pjs-gap: 8px;
```
SlexAxton, line 4: seems like you could rename these `--pjs-bg-fallback: #fff` and then use them as the fallbacks instead of the literals below?

amadeus: yeah I went back and forth in my head about this... not strong feelings either way... but easy enough to add and then we don't have to worry about the 8px getting out of sync for whatever reason

amadeus: Oki, adding in the fallback, fixed a few more places that needed updates and also added to the docs

SlexAxton: ah to be clear, i didn't feel strongly about doing it, so much as I wanted to point out that it would resolve the 'fallback' variable behavior you described wanting in the pr description. seems fine either way to me in practice unless there's a lot of duplication in the future.

PR #106, opened 2025-10-28
Improve builds and publishing

SlexAxton:


packages/precision-diffs/src/react/utils/templateRender.tsx
```diff
@@ -3,7 +3,7 @@ import type { ReactNode } from 'react';
 export function templateRender(
   children: ReactNode,
   __html: string | undefined
-) {
+): React.JSX.Element {
```
amadeus, line 6: Wonder if we should just make this `ReactNode` instead? mb doesn't matter tho. But technically this is not supposed to be a component, rather a render function

packages/precision-diffs/src/custom-components/Container.ts
```diff
@@ -1,4 +1,4 @@
-import styles from '../style.css?raw';
+import styles from '../style.css';
```
amadeus, line 1: I assume you setup something different for importing this as text?

Running `demo:dev` seems to imply that it still imports properly

packages/precision-diffs/src/DiffHunksRenderer.ts
```diff
@@ -98,6 +98,10 @@ export type DiffHunksRendererOptions =
   | DiffHunkRendererThemeOptions
   | DiffHunkRendererThemesOptions;
 
+type OptionsWithDefaults =
+  | Required<Omit<DiffHunkRendererThemeOptions, 'theme'>>
+  | Required<Omit<DiffHunkRendererThemesOptions, 'themes'>>;
```
amadeus, line 103: This gonna get real thrashy with my code, but I can probably resolve it

packages/precision-diffs/src/DiffHunksRenderer.ts
```diff
@@ -121,24 +125,24 @@ export class DiffHunksRenderer<LAnnotation = undefined> {
...
-  expandHunk(index: number) {
+  expandHunk(index: number): void {
     this.expandedHunks.add(index);
   }
 
-  private isHunkExpanded(index: number) {
+  private isHunkExpanded(index: number): boolean {
     return (
-      this.getOptionsWithDefaults().expandUnchanged ||
+      (this.getOptionsWithDefaults().expandUnchanged ?? false) ||
```
amadeus: this `?? false` shouldn't be needed now?

packages/precision-diffs/src/DiffHunksRenderer.ts
```diff
@@ -275,24 +279,23 @@ export class DiffHunksRenderer<LAnnotation = undefined> {
...
-    } = this.getOptionsWithDefaults();
+    } = options;
 
     this.diff = fileDiff;
     const additionsAST: ElementContent[] = [];
     const deletionsAST: ElementContent[] = [];
     const unifiedAST: ElementContent[] = [];
     const { state, transformers, toClass } = createTransformerWithState({
-      disableLineNumbers,
+      disableLineNumbers: disableLineNumbers ?? false,
```
amadeus: Shouldn't need this `?? false` either cause of the type?

packages/precision-diffs/src/DiffHunksRenderer.ts
```diff
@@ -304,7 +307,7 @@ export class DiffHunksRenderer<LAnnotation = undefined> {
         return fileDiff.hunks;
       }
       if (
-        expandUnchanged &&
+        (expandUnchanged ?? false) &&
```
amadeus: Similar sitch, shouldn't need `?? false`

packages/precision-diffs/src/DiffHunksRenderer.ts
SlexAxton: Yea, I was just codifying the existing behavior, but there's definitely room to just make it better.

packages/precision-diffs/src/DiffHunksRenderer.ts
SlexAxton: yea, that is correct, i think ai did this before i fixed the types and i guess it doesnt complain in reverse

SlexAxton: pushing a fix for the 3 affected places

packages/precision-diffs/src/custom-components/Container.ts
SlexAxton: yea, it uses `tsdown` to handle

packages/precision-diffs/src/react/utils/templateRender.tsx
SlexAxton: easy switch to whatever you wanna do, the type gods were satisfied with this is all

packages/precision-diffs/src/DiffHunksRenderer.ts
amadeus: Not your fault, mostly just a comment as a note for myself

PR #115, opened 2025-10-30
Add cursor pointer to clickable lines

amadeus:
lil request from OpenAI


packages/precision-diffs/src/File.ts
```diff
@@ -328,10 +328,14 @@ export class File<LAnnotation = undefined> {
     pre.removeEventListener('click', this.handleMouseClick);
     pre.removeEventListener('mousemove', this.handleMouseMove);
     pre.removeEventListener('mouseout', this.handleMouseLeave);
+    delete pre.dataset.interactiveLines;
```
amadeus, line 331: I realize this isn't the same as in FileDiff, but this will be properly resolved in my other PR #114

PR #119, opened 2025-10-30
Actually Fix `disableLineNumbers`

amadeus:
Alright, turned out I was trying to be more clever than I needed to be.

Most of this change is just adding a hidden toggle to play with line number changes


packages/precision-diffs/src/style.css
```diff
@@ -233,6 +233,14 @@ code {
   }
 }
 
+[data-column-number] {
+  grid-column: 1 / 2;
+}
+
+[data-column-content] {
+  grid-column: 2 / 3;
+}
```
amadeus, line 242: This was the only required change.

PR #127, opened 2025-11-05
Smarter Line Number Column Defaults

amadeus:
a quick sprint on making the number column minimum width based on number of lines the diff can show


packages/precision-diffs/src/style.css
```diff
@@ -550,15 +550,18 @@ code {
 }
 
 [data-column-number] {
+  box-sizing: content-box;
```
amadeus, line 553: Decided to make the variable a bit more ergonomic, so it matches the character count exactly

PR #129, opened 2025-11-05
gap-style

fat:
Allow someone to override the gap border styles

### DIFF
--- a/packages/precision-diffs/src/style.css
+++ b/packages/precision-diffs/src/style.css
@@ -10,7 +10,6 @@
 
   --pjs-mixer: light-dark(black, white);
   --pjs-gap-fallback: 8px;
-  --pjs-gap-style: 1px solid var(--pjs-bg);
 
   /*
     // Available CSS Color Overrides


packages/precision-diffs/src/style.css
```diff
@@ -5,11 +5,12 @@
...
     'Courier New', monospace;
   --pjs-header-font-fallback:
-    system-ui, -apple-system, 'Segoe UI', Roboto, 'Helvetica Neue', 'Noto Sans',
-    'Liberation Sans', Arial, sans-serif;
+    -apple-system, BlinkMacSystemFont, 'Segoe UI', Roboto, Oxygen, Ubuntu,
+    Cantarell, 'Fira Sans', 'Droid Sans', 'Helvetica Neue', sans-serif;
 
   --pjs-mixer: light-dark(black, white);
   --pjs-gap-fallback: 8px;
+  --pjs-gap-style: 1px solid var(--pjs-bg);
```
amadeus: Don't define it here, otherwise it's not overrideable

PR #130, opened 2025-11-06
Allow unsafe css to be injected into the shadow root stylesheet

SlexAxton:
Documented it behind a url flag for now.

image

### DIFF
--- a/apps/docs/app/docs/Styling/Styling.tsx
+++ b/apps/docs/app/docs/Styling/Styling.tsx
@@ -3,6 +3,7 @@
 import { IconCiWarning } from '@/components/icons';
 import type { PreloadedFileResult } from '@pierre/precision-diffs/ssr';
 import { useSearchParams } from 'next/navigation';
+import { Suspense } from 'react';
 
 import { DocsCodeExample } from '../DocsCodeExample';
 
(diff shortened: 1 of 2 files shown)


packages/precision-diffs/src/types.ts
```diff
@@ -99,6 +99,9 @@ export interface BaseCodeOptions {
   lang?: SupportedLanguages;
   preferWasmHighlighter?: boolean;
   useCSSClasses?: boolean;
+
+  // Custom CSS injection
+  unsafeCss?: string;
```
amadeus: can we do `unsafeCSS` to match `useCSSClasses`?

packages/precision-diffs/src/FileDiff.ts
```diff
@@ -518,6 +519,30 @@ export class FileDiff<LAnnotation = undefined> {
     });
   }
 
+  private injectUnsafeCss(): void {
```
amadeus: Similar note on CSS casing

packages/precision-diffs/src/FileDiff.ts
amadeus, line 243: We'd probably want to call `injectUnsafeCSS` here on hydration

packages/precision-diffs/src/types.ts
SlexAxton: maybe `useCSSClasses` is the one who's wrong

PR #134, opened 2025-11-06
Batched Hunk Expansion

amadeus:
Implementing the ability to expand hunks in batches. Can help potentially with performance and stuff...

As part of this, I will also need to figure out ways to re-use existing code better... since this most definitely exposes some issues with incremental rendering...


apps/docs/app/docs/ReactAPI/constants.ts
```diff
@@ -333,15 +337,17 @@ export function SingleDiff() {
 export const REACT_API_FILE_DIFF: PreloadFileOptions<undefined> = {
   file: {
     name: 'file_diff.tsx',
-    contents: '// coming soon',
+    contents:
+      '// documentation coming soon, check typescript types for more info',
```
amadeus, line 341: Wanted to clarify some of this -- the components actually exist if people want to use them, so just pointing them to introspect the source code if they want to use the feature

PR #140, opened 2025-11-08
Add GH Action for CI, fix Stylelint errors

mdo:
- Runs all the things in one big Action workflow
- Caches bun install on first job, so others are faster

TBD on needing all this, but might be worth one final check if something's off with local editor extensions/settings.

### DIFF
--- a/stylelint.config.js
+++ b/stylelint.config.js
@@ -20,13 +20,11 @@ export default {
     ],
     'selector-class-pattern': null,
     'no-descending-specificity': null,
-    // 'no-duplicate-selectors': null,
 
     'property-no-vendor-prefix': null,
     'value-no-vendor-prefix': null,
     'declaration-property-value-keyword-no-deprecated': null,
 
-    // === Style Flexibility (not errors, just preferences) ===
     // Whitespace & Formatting
     'custom-property-empty-line-before': null,
     'comment-empty-line-before': null,
@@ -44,10 +42,5 @@ export default {
     // Value Formatting
     'value-keyword-case': null,
     'import-notation': null,
-    // 'shorthand-property-no-redundant-values': null,
-    // 'declaration-block-no-redundant-longhand-properties': null,
-
-    // Modern CSS Features
-    // 'media-feature-range-notation': null,
   },
 };


apps/demo/src/style.css
```diff
@@ -45,7 +45,6 @@ body,
   display: block;
   margin: 0;
   padding: 0;
-  min-height: 100vh;
```
mdo, line 48: Unsure if this was intentional?

apps/demo/src/style.css
```diff
@@ -61,7 +60,7 @@ textarea {
   appearance: none;
   border-radius: 0;
   border: 1px solid var(--fg);
-  padding: 4px 4px;
+  padding: 4px;
```
mdo, line 64: Shorthand

apps/demo/src/style.css
```diff
@@ -163,7 +162,7 @@ code {
 .files-input {
   position: fixed;
   z-index: 20;
-  inset: 50px 0 0 0;
+  inset: 50px 0 0;
```
mdo, line 166: Another shorthand

apps/docs/app/globals.css
```diff
@@ -112,19 +113,14 @@
   --sidebar-accent-foreground: oklch(0.985 0 0);
   --sidebar-border: oklch(1 0 0 / 10%);
   --sidebar-ring: oklch(0.556 0 0);
+  color-scheme: dark;
```
mdo, line 116: This and the above one dedupe some selectors.

packages/precision-diffs/src/style.css
mdo, line 1: Changes here are all to combine and reduce # of selectors—so just some light moving around.

apps/demo/src/style.css
amadeus: yeah, was intentional iirc

amadeus: maaay be worth ignoring this file in general cause it's really just for hacking shit together

apps/docs/app/layout.tsx
```diff
@@ -12,7 +12,6 @@ import {
 import localFont from 'next/font/local';
 
 import './globals.css';
-// import '../styles/tokens.css';
```
amadeus, line 15: Should we yeet this comment?

packages/precision-diffs/src/style.css
```diff
@@ -555,6 +549,7 @@ code {
 }
 
 [data-column-number] {
+  grid-column: 1 / 2;
```
amadeus, line 556: have you tested this change with unified vs split, and with scrollable vs not? it might break things?

stylelint.config.js
```diff
@@ -1,80 +1,53 @@
...
-    'selector-nested-pattern': null,
-
-    // Disable some overly strict rules
+    'selector-class-pattern': null,
     'no-descending-specificity': null,
-    'no-duplicate-selectors': null,
-    'declaration-block-no-duplicate-properties': null, // Allow duplicates for fallbacks
-    'declaration-block-no-redundant-longhand-properties': null,
-    'shorthand-property-no-redundant-values': null,
+    // 'no-duplicate-selectors': null,
```
amadeus: Should this be deleted? or was it inadvertently commented out?

stylelint.config.js
```diff
@@ -1,80 +1,53 @@
...
+    // Value Formatting
+    'value-keyword-case': null,
+    'import-notation': null,
+    // 'shorthand-property-no-redundant-values': null,
+    // 'declaration-block-no-redundant-longhand-properties': null,
 
-    // Allow deprecated keywords (sometimes needed for compatibility)
-    'declaration-property-value-keyword-no-deprecated': null,
+    // Modern CSS Features
+    // 'media-feature-range-notation': null,
```
amadeus: Similar question about these lads being commented out

apps/docs/app/layout.tsx
mdo: Already yeeted—that's a deletion

packages/precision-diffs/src/style.css
mdo: I did. The change here is just to combine existing selectors, so the specificity is the same, only some slight placement in the order of styles are different. However, since specificity hasn't changed, there's no difference here.

PR #141, opened 2025-11-08
Fat/selection

fat:
This add a rich selection api to our diff and file components as well as implements a multi line comment ui in the comments and annotation example.

https://github.com/user-attachments/assets/c1ee82c0-413c-4ebe-9eff-4fe2fb3cc6de

### DIFF
--- a/apps/docs/app/docs/ReactAPI/constants.ts
+++ b/apps/docs/app/docs/ReactAPI/constants.ts
@@ -55,6 +55,16 @@ export function SingleDiff() {
         // for them
         return <CommentThread threadId={annotation.metadata.threadId} />;
       }}
+
+      // Programmatically control which lines are selected. This
+      // allows two-way binding with state. Selections should be 
+      // stable across 'split' and 'unified' diff styles.
+      // 'start' and 'end' map to the visual line numbers you see in the
+      // number column. 'side' and 'endSide' are considered optional.
+      // 'side' will default to the 'additions' side. 'endSide' will
+      // default to whatever 'side' is unless you specify otherwise.
+      selectedLines={{ start: 3, side: 'additions', end: 4, endSide: 'deletions' }}
+
       // Here's every property you can pass to options, with their
       // default values if not specified.
       options={{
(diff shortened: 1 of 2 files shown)


apps/docs/app/diff-examples/Annotations/Annotations.tsx
```diff
@@ -144,6 +168,7 @@ export function Annotations({ prerenderedDiff }: AnnotationsProps) {
               backgroundColor: '#1a76d4',
               transition: 'none',
               cursor: 'pointer',
+              pointerEvents: 'none',
```
fat, line 168: I changed the logic here so that this is really like a "fake" button and clicking it really clicks through to the line number which is what ultimately triggers the comment adding.

Number of benefits here imo (particularly fixes annoying thing with mouse over / highlight state, etc.

apps/docs/app/diff-examples/Annotations/Annotations.tsx
```diff
@@ -155,6 +180,9 @@ export function Annotations({ prerenderedDiff }: AnnotationsProps) {
           options={{
             ...prerenderedDiff.options,
             onLineEnter: handleLineEnter,
+            enableLineSelection: !hasOpenCommentForm,
```
fat, line 181: @amadeus i think this is the right place for this stuff to live, but lmk if you'd rather it be top level

apps/docs/app/diff-examples/Annotations/Annotations.tsx
```diff
@@ -204,44 +232,70 @@ function CommentForm({
 
   return (
     <div
-      className="max-w-[95%] sm:max-w-[70%]"
       style={{
-        whiteSpace: 'normal',
-        margin: 20,
-        fontFamily: 'Geist',
+        marginLeft: -43,
```
fat: some tom foolery for this component, to make it simulate the continuation of the left gutter – like this:

Image

packages/precision-diffs/src/LineSelectionManager.ts
```diff
@@ -0,0 +1,454 @@
+import type { AnnotationSide } from './types';
```
fat, line 3: I mostly tried to crib the logic from pierre.co for this guy – but follow existing code patterns. One thing to note is i ended up ripping out the url management… and just adding more event handers (assume people could wire that up + scroll to) themselves

packages/precision-diffs/src/react/utils/useFileDiffInstance.ts
```diff
@@ -60,15 +60,34 @@ export function useFileDiffInstance<LAnnotation>({
...
-    instanceRef.current.setOptions(options);
-    void instanceRef.current.render({
-      forceRender,
-      fileDiff,
-      oldFile,
-      newFile,
-      lineAnnotations: lineAnnotations,
-    });
+    const instance = instanceRef.current;
+    const forceRender = !deepEqual(instance.options, options);
```
amadeus, line 67: so the downside of this is that forceRender will become true when the selection updates, which will trickle down into the render step which is probably something we don't want.

We need to come up with a better way for these side-effecty things.

fat: yah that makes sense to me…

PR #144, opened 2025-11-11
Add `.browserslistrc` and setup Autoprefixer

mdo:
- **Adds `.browserslistrc`.** Used the minimum version of the main browsers for now based on what CSS support we currently need. This could easily get to oler verisons if we decided to avoid `light-dark()` and some color functions.

- **Generates updated `dist/style.css` with prefixes** via PostCSS. LightningCSS wasn't an option because it reformats and strips comments, and I'd think we want to keep those intact for developer convenience.

- ~~**Adds new `/css-comparison` page** with with our own diffs to show the changes from Autoprefixer, mostly as a local manual verification step for myself for now.~~

I could use some review on the browsers list and the `tsdown.config.ts` changes.

### DIFF
--- a/.browserslistrc
+++ b/.browserslistrc
@@ -3,7 +3,7 @@
 #
 # Used by:
 # - apps/docs (Next.js + Tailwind CSS v4)
-# - packages/precision-diffs (tsdown + lightningcss)
+# - packages/precision-diffs (tsdown + postcss)
 
 chrome >= 123
 edge >= 123


packages/precision-diffs/package.json
```diff
@@ -37,6 +37,8 @@
     "@types/hast": "catalog:",
     "@types/react": "catalog:",
     "@types/react-dom": "catalog:",
+    "autoprefixer": "^10.4.22",
+    "postcss": "^8.5.6",
```
amadeus: The should be set in the root package.json catalog, and then set here via `catalog:`

Also if when added to the root catalog, can you remove the `^` and just set the version directly?

packages/precision-diffs/tsdown.config.ts
```diff
@@ -12,6 +16,24 @@ const config: UserConfig = defineConfig([
     },
     unbundle: true,
     platform: 'neutral',
+    plugins: [
```
amadeus, line 19: This seems fine to me, but I am not in any way familiar with the right way to do builds and stuff, so i would defer to @SlexAxton on whether this is the right way to do this

SlexAxton: good enough for a single file for sure.

PR #148, opened 2025-11-12
Another pass at selection colors

mdo:
- Uses new theme blue
- Reduces selectors with nesting
- Simplifies # of colors needed with still the same effect

Needs more time to bake.

### DIFF
--- a/packages/precision-diffs/src/style.css
+++ b/packages/precision-diffs/src/style.css
@@ -229,27 +229,6 @@ code {
     )
   );
 
-  /* --pjs-selection-base: var(
-    --pjs-selection-color-override,
-    light-dark(rgb(37, 99, 235), rgb(147, 197, 253))
-  );
-  --pjs-bg-selection: var(
-    --pjs-bg-selection-override,
-    light-dark(rgb(147, 197, 253, 0.28), rgb(96, 165, 250, 0.32))
-  );
-  --pjs-bg-selection-number: var(
-    --pjs-bg-selection-number-override,
-    light-dark(rgb(96, 165, 250, 0.55), rgb(59, 130, 246, 0.6))
-  );
-    --pjs-bg-selection-background: var(
-    --pjs-bg-selection-background-override,
-    light-dark(rgb(96, 165, 250, 0.2), rgb(59, 130, 246, 0.26))
-  );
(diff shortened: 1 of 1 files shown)


apps/docs/app/diff-examples/LineSelection/LineSelection.tsx
```diff
@@ -19,64 +20,103 @@ export function LineSelection({ prerenderedDiff }: LineSelectionProps) {
     last: number;
   } | null>(null);
 
+  // Store the current theme
+  const [theme, setTheme] = useState<'pierre-dark' | 'pierre-light'>(
+    'pierre-dark'
```
amadeus, line 25: this is a bit tricky here because of the nature of pre-rendered diffs. it works right now, but if someone changes the theme in the pre-render, then we'll have a mismatch.

Normally you `should` do something like:

```ts
prerenderedDiff.options?.theme ?? 'pierre-dark'
```

however i have the feeling this will have typescript issues, but maybe try?

packages/precision-diffs/src/style.css
```diff
@@ -241,14 +241,48 @@ code {
...
+      rgb(from var(--pjs-selection-base) r g b / 0.26)
+    )
+  ); */
+  /* --pjs-bg-selection-number-background: var(
+    --pjs-bg-selection-number-background-override,
+    light-dark(
+      rgb(from var(--pjs-selection-base) r g b / 0.25),
+      rgb(from var(--pjs-selection-base) r g b / 0.4)
+    )
+  ); */
```
amadeus: Probs should just delete these meaty comments?

packages/precision-diffs/src/style.css
```diff
@@ -241,14 +241,48 @@ code {
...
   );
-  --pjs-bg-selection-background: var(
+    --pjs-bg-selection-background: var(
     --pjs-bg-selection-background-override,
     light-dark(rgb(96, 165, 250, 0.2), rgb(59, 130, 246, 0.26))
   );
   --pjs-bg-selection-number-background: var(
     --pjs-bg-selection-number-background-override,
     light-dark(rgb(59, 130, 246, 0.4), rgb(37, 99, 235, 0.45))
+  ); */
```
amadeus: This one too

PR #154, opened 2025-11-13
Adding addition/deleted line counts to Hunk

amadeus:
Came about through a conversation with openai guys, would make some things easier for them to compute in custom headers.

I tested this on a really big diff and it didn't appear to add any realistic overhead, which seems good. It probably can in certain extreme cases.

### DIFF
--- a/packages/precision-diffs/src/utils/parseLineType.ts
+++ b/packages/precision-diffs/src/utils/parseLineType.ts
@@ -10,7 +10,7 @@ export function parseLineType(
   line: string,
   maxLineLength: number
 ): ParseLineTypeReturn {
-  const firstChar = line.substring(0, 1);
+  const firstChar = line[0];
   if (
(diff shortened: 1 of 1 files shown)


packages/precision-diffs/src/utils/parsePatchFiles.ts
```diff
@@ -124,13 +126,23 @@ function processPatch(data: string): ParsedPatch {
...
+          } else if (line.startsWith('-')) {
+            deletedLines++;
+          }
+        }
       }
       const hunkData: Hunk = {
         collapsedBefore: 0,
         additionCount: parseInt(match[4] ?? '1'),
         additionStart: parseInt(match[3]),
+        additionLines,
```
amadeus, line 142: I really need to rename this to `added` prefix, but a change for another day

PR #172, opened 2025-11-25
Fix selection colors

mdo:
Update with latest from 0ba7073:

- Additions, deletions, and context-expanded lines set background via CSS variable reassignment. `--pjs-bg: suchandsuch`.
- This allows us to have one big selector group for selected lines in all three contexts using the same styles, remixing different `--pjs-bg` values the same way across their contextual background colors.

I still need to see if I can get rid of some existing CSS variables, or use those here instead. I think I can't do the latter because the values of the variable at that time are already computed or whatever. Will continue testing.

### DIFF
--- a/packages/precision-diffs/src/style.css
+++ b/packages/precision-diffs/src/style.css
@@ -358,6 +358,9 @@ code {
 
   &::before {
     content: '';
+    position: sticky;
+    top: 0;
+    left: 0;
     display: block;
     border-right: var(--pjs-gap-style, 1px solid var(--pjs-bg));
     background-color: var(--pjs-bg-selection-number);


packages/precision-diffs/src/style.css
```diff
@@ -358,6 +358,9 @@ code {
 
   &::before {
     content: '';
+    position: sticky;
+    top: 0;
+    left: 0;
```
amadeus, line 363: Small issues i found that wasn't visible before, but these lads need to be sticky on horizontal scroll

packages/precision-diffs/src/style.css
```diff
@@ -577,6 +580,7 @@ code {
 [data-line-annotation] {
   min-height: var(--pjs-annotation-min-height, 0);
   background-color: var(--pjs-bg-context);
+  z-index: 3;
```
amadeus, line 583: Quick fix since we are CSSin in here for that hover bug with the accept/reject annotation api

packages/precision-diffs/src/style.css
```diff
@@ -873,6 +876,25 @@ code {
...
+
+[data-line-type='context-expanded'][data-selected-line] [data-column-number],
+[data-line-type='change-addition'][data-selected-line] [data-column-number],
+[data-line-type='change-deletion'][data-selected-line] [data-column-number] {
+  color: var(--pjs-selection-number-fg);
+  background-color: light-dark(
+    color-mix(in lab, var(--pjs-bg) 75%, var(--pjs-selection-base)),
+    color-mix(in lab, var(--pjs-bg) 60%, var(--pjs-selection-base))
+  );
+}
```
amadeus, line 896: clever boi

PR #175, opened 2025-11-26
Possible fix for split gutter on wrap

amadeus:
Not sure why I didn't do this originally, I do worry there could be some issue afoot but in my quick visual tests, it appears to be fine

### DIFF
--- a/packages/precision-diffs/src/style.css
+++ b/packages/precision-diffs/src/style.css
@@ -375,16 +375,6 @@ code {
   cursor: pointer;
 }
 
-[data-type='split'][data-overflow='wrap'] [data-additions] [data-line],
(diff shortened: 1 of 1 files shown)


packages/precision-diffs/src/style.css
```diff
@@ -377,10 +377,12 @@ code {
 
 [data-type='split'][data-overflow='wrap'] [data-additions] [data-line],
 [data-type='split'][data-overflow='wrap'] [data-additions] [data-separator],
+[data-type='split'][data-overflow='wrap'] [data-additions] [data-buffer],
```
amadeus: Realized certain buffer types were not also accomodating this gutter, so fixed that upp too

PR #176, opened 2025-11-26
Potential bg color mixing fixes

amadeus:


packages/precision-diffs/src/style.css
```diff
@@ -711,26 +715,26 @@ code {
...
     background-color: var(--pjs-bg-deletion-number);
   }
 }
 
 [data-line-type='context-expanded'] {
-  --pjs-bg: var(--pjs-bg-context);
+  --pjs-line-bg: var(--pjs-bg-context);
 
   [data-column-number] {
-    --pjs-bg: var(--pjs-bg-context);
```
amadeus, line 733: I think this was a bug from before?

PR #182, opened 2025-12-02
Add a single snapshot test to diffs

SlexAxton:
Gettin' it goin'

`bun run diffs:test`

### DIFF
--- a/packages/precision-diffs/test/FileRenderer.ast.test.ts
+++ b/packages/precision-diffs/test/FileRenderer.ast.test.ts
@@ -27,9 +27,9 @@ describe('FileRenderer AST Structure', () => {
 
       // Each line should have the correct properties
       expect(lineElement.properties).toBeDefined();
-      expect(lineElement.properties!['data-line']).toBe(i + 1);
-      expect(lineElement.properties!['data-line-type']).toBe('context');
-      expect(lineElement.properties!['data-line-index']).toBe(i);
+      expect(lineElement.properties['data-line']).toBe(i + 1);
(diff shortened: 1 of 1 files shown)


packages/precision-diffs/test/__snapshots__/FileRenderer.test.ts.snap
```diff
@@ -0,0 +1,1558 @@
+// Bun Snapshot v1, https://bun.sh/docs/test/snapshots
```
SlexAxton, line 1: If you're not used to snapshot testing, this file is automatically generated with `--update-snapshots` - it's supposed to be checked in. Then when you run tests in the future it checks them against this snapshot. And if it's different than what comes out it errors. If you intend for the change to occur, then you run `--update-snapshots` again and check in the change. Now all changes to the AST output come with a visible representation of the diff in output.

image

amadeus: oh damn that's sick

AGENTS.md
```diff
@@ -33,3 +33,38 @@ We use project references between each of our packages and apps.
...
+bun run diffs:test
+```
+
+### Updating Snapshots
+
+When test snapshots need to be updated:
+
+```bash
+# From the package directory
+bun test --update-snapshots
```
amadeus, line 59: Should we let the AI know about this? Could that result in them seeing a failing test and then going `oh, well, i'll just update the snapshots` to `fix` the problem?

SlexAxton: these are like an industry standard, so AI should know when to do it and when to not.

PR #188, opened 2025-12-02
More docs design tweaks

mdo:
- Stickied header on scroll w/ bottom border
- Redesigned header to be on a single line and reduced Discord/GH links to icon-only
- Simplified mobile header—no byline in the logo
- Redesigned mobile docs menu
- Tweaked a lot of docs content spacing and sizing (text now smaller on mobile, tighter line height, improved colors in notices, etc)
- Mobile docs sidebar menu has a default active item (first item) on page load

### DIFF
--- a/apps/docs/app/HeaderWrapper.tsx
+++ b/apps/docs/app/HeaderWrapper.tsx
@@ -9,8 +9,13 @@ export function HeaderWrapper() {
   const [isStuck, setIsStuck] = useState(false);
 
   useEffect(() => {
+    let lastStuck: boolean | undefined;
     const handleScroll = () => {
-      setIsStuck(window.scrollY > 0);
+      const isStuck = window.scrollY > 0;
+      if (isStuck !== lastStuck) {
+        lastStuck = isStuck;
+        setIsStuck(isStuck);
+      }
     };
 
     // Check initial state
@@ -35,7 +40,6 @@ export function HeaderWrapper() {
         <Header.NavLink href="/">Home</Header.NavLink>
         <Header.NavLink href="/docs">Docs</Header.NavLink>
         <li className="border-border h-5 w-[1px] items-center border-l" />
-        {/* <li className="ml-auto bg-red-500 md:hidden" /> */}
         <Header.NavLink href="https://discord.gg/pierre" external>
           <IconBrandDiscord />
         </Header.NavLink>
--- a/apps/docs/components/ui/header.tsx
+++ b/apps/docs/components/ui/header.tsx
@@ -103,7 +103,6 @@ function HeaderNavLink({
         {...props}
       >
         {children}
-        {/* {external && <IconArrowUpRight />} */}
       </NavigationMenuLink>
     </NavigationMenuItem>
   );
--- a/apps/docs/components/ui/notice.tsx
+++ b/apps/docs/components/ui/notice.tsx
@@ -26,7 +26,9 @@ export type NoticeProps = React.HTMLAttributes<HTMLDivElement> &
 function Notice({ className, variant, icon, children, ...props }: NoticeProps) {
   return (
     <div className={cn(noticeVariants({ variant, className }))} {...props}>
-      {icon != null && <div className="mt-[2px] flex-shrink-0 md:mt-[4px]">{icon}</div>}
+      {icon != null && (
+        <div className="mt-[2px] flex-shrink-0 md:mt-[4px]">{icon}</div>
+      )}
       <div className="leading-[1.5]">{children}</div>
     </div>
   );


apps/docs/app/docs/DocsSidebar.tsx
```diff
@@ -120,15 +132,18 @@ export function DocsSidebar({
...
+          className="fixed inset-0 z-51 bg-black/50 md:hidden"
           onClick={onMobileClose}
         />
       )}
 
       <aside
-        className={`bg-background border-border fixed top-0 left-0 z-50 h-screen w-72 -translate-x-full transform overflow-y-auto border-r p-5 shadow-xl transition-transform duration-300 ease-in-out md:pointer-events-auto md:relative md:top-auto md:left-auto md:z-auto md:block md:h-auto md:w-auto md:translate-x-0 md:transform-none md:overflow-visible md:border-none md:bg-transparent md:px-0 md:py-7 md:opacity-100 md:shadow-none md:transition-none ${isMobileOpen ? 'translate-x-0' : '-translate-x-full'} `}
+        className={`bg-background border-border fixed top-0 left-0 z-60 h-screen w-72 -translate-x-full transform overflow-y-auto border-r p-5 shadow-xl transition-transform duration-300 ease-in-out md:pointer-events-auto md:relative md:top-auto md:left-auto md:z-auto md:block md:h-auto md:w-auto md:translate-x-0 md:transform-none md:overflow-visible md:border-none md:bg-transparent md:px-0 md:py-7 md:opacity-100 md:shadow-none md:transition-none ${isMobileOpen ? 'translate-x-0' : '-translate-x-full'} `}
```
amadeus, line 141: not your fault, but god i find tailwind obnoxious 🥲

apps/docs/app/docs/HeadingAnchors.tsx
```diff
@@ -0,0 +1,55 @@
...
+          const url = `${window.location.origin}${window.location.pathname}#${heading.id}`;
+
+          void navigator.clipboard.writeText(url).catch((err) => {
+            console.warn('Failed to copy to clipboard:', err);
+          });
+        });
+
+        heading.appendChild(anchor);
+      }
+    }, 150);
```
amadeus, line 44: I haven't looked at this too closely, and i assume this is just moving stuff around or something? but would love to not have to setTimeout on this, and be able to make it so the effect is `useLayoutEffect`. That said tho, i tried tweaking that earlier and it broke a bunch of scroll detection stuff, so yeah, no point in fuxing with it now, just my long yearning thoughts

apps/docs/app/HeaderWrapper.tsx
```diff
@@ -1,34 +1,46 @@
...
 import { Header } from '@/components/ui/header';
+import { useEffect, useRef, useState } from 'react';
 
 export function HeaderWrapper() {
+  const headerRef = useRef<HTMLElement>(null);
+  const [isStuck, setIsStuck] = useState(false);
+
+  useEffect(() => {
+    const handleScroll = () => {
+      setIsStuck(window.scrollY > 0);
```
amadeus: one thing that might be nice, would be to have a cache internally of the boolean value, so that you don't have to trigger react trying to determine whether it it needs to re-render again constantly (scrolling is a super noisy activity, and yes, while react will do logic and and be like, i don't need to re-render, i just figure avoid even going there on your own, just in case (probably an over optimization, but it definitely feels better).

So you'd do something like:

```ts
let lastStuck: boolean | undefined;
const handleScroll = () => {
  const isStuck = window.scrollY > 0;
  if (isStuck !== lastStuck) {
    lastStuck = isStuck;
    setIsStuck(isStuck);
  }
}
```

Basically then you only ever call setState when the value changes

apps/docs/app/HeaderWrapper.tsx
```diff
@@ -1,34 +1,46 @@
...
+        <Header.Logo href="/" subtitle={<>by The Pierre Computer Co.</>}>
           Diffs
         </Header.Logo>
       }
     >
       <Header.Nav>
         <Header.NavLink href="/">Home</Header.NavLink>
         <Header.NavLink href="/docs">Docs</Header.NavLink>
+        <li className="border-border h-5 w-[1px] items-center border-l" />
+        {/* <li className="ml-auto bg-red-500 md:hidden" /> */}
```
amadeus: Can we delete this? maybe this just a WIP so ignore if so

apps/docs/components/ui/header.tsx
```diff
@@ -141,14 +103,36 @@ function HeaderNavLink({
         {...props}
       >
         {children}
-        {external && <IconArrowUpRight />}
+        {/* {external && <IconArrowUpRight />} */}
```
amadeus: Do we still need this?

apps/docs/app/docs/DocsSidebar.tsx
mdo: Yes, I entirely agree lol. We could promote these to global classes and do some `@apply` stuff. For example, from `globals.css`. Still a similar problem, but cleans up our HTML.

```css
[data-slot='header'] {
  @apply border-b border-transparent transition-[border-color] duration-200;
}
```

apps/docs/app/HeaderWrapper.tsx
mdo: Oh that makes a lot of sense. Nice.

apps/docs/app/docs/HeadingAnchors.tsx
mdo: We'd get some of this for free once we get on the MDX train, but until then, yeah, it's a tough call because I don't think Next will always know what the DOM is right off the bat from these components? I'll give MDX a more earnest go next and aim to resolve this stuff.

apps/docs/app/docs/DocsSidebar.tsx
amadeus: We could just like… rebel against Alex and change to something else 👯

PR #193, opened 2025-12-04
Worker Architecture Refactor

amadeus:
The main goal of this was to make it so VSCode extensions can more easily adopt the WorkerPool API (a system that would definitely benefit a more long running process like this).

The ultimate limitation of WebWorkers in VSCode extensions is that they can not dynamically load imports. This sorta left us in a bit of a tough spot. I attempted a quick spring to just make a very large bundle that included all of shiki's languages and themes, however it ballooned the file to around 12mb and took actual seconds to spin up the worker in practice, so this was mostly a no go.

In this process I made a discovery that shiki themes and languages are basically just JSON blobs, so with a bit of a rework, I can make it so the WorkerPoolManager just sends over the appropriate themes and languages on demand since the main thread can import dynamically as needed.

This presented a bit of a challenge because a lot of the existing architecture depended on the built in import() logic to setup. I initially had claude do a moon shot to get this working, and surprisingly in about an hour back and forth, he got it working. We had webworker initialization times under 50ms and no noticeable performance hits.

However when i started to go through claude's code, I quickly realized this wasn't going to be a solution we could ship. So I spent most of today rewriting it. I'm thankful for some of the ideas it gave me, but it had no sense for architecture.

Anyways, I think this is now in a good spot, but it is a big change, so it does sorta reset me a bit in terms of this beta, in terms of feeling confident about the status of it.

Furthermore, OpenAI bro is out for the rest of the week and he's been by far one of the most valuable testers on this (in terms of finding bugs and battle testing stuff).

Anyways, tomorrow, i will look to merge this probably after I go over a few more tests and stuff. 

**NOTE 1:**
One thing that is a bit annoying however, is that I still have to make 2 versions of the worker. One that's fairly modern and whatever, and another that's more self contained. (worker.ts and worker-portable.ts)

If you check the config in tsdown.config you'll see the differences. I was hoping to JUST ship `worker-portable` however it causes very weird warnings when used with NextJS (it thinks it's a commonjs file, and neither myself or claude could figure out how to get rid of those NextJS warnings). Ultimately I think this is a problem perfectly suited for @SlexAxton, whenever he has time (after his trip or whatever), but in the meantime just having 2 scripts seemed pretty harmless, given both are not very big.

**NOTE 2:**
@mdo I added some instructions that might be worth you taking a pass on. One thing I also thought looked nicer, is having all the files render without a header, just feels cleaner. It does mean no `copy` button but like, i feel like that's probably ok for most of these? Or we can just try to figure out how to add another copy button that isn't built into the header...

### DIFF
--- a/packages/precision-diffs/tsconfig.json
+++ b/packages/precision-diffs/tsconfig.json
@@ -1,10 +1,9 @@
 {
   "extends": "../../tsconfig.options.json",
   "include": [
-    "scripts/**/*.ts",
-    "src/**/*.json",
     "src/**/*.ts",
     "src/**/*.tsx",
+    "src/**/*.json",
     "test/**/*.ts",
     "tsdown.config.ts"
   ],


apps/docs/app/docs/WorkerPool/WorkerPool.tsx
```diff
@@ -95,9 +109,45 @@ export function WorkerPool({
         Workers only work in client components. Ensure your function has the{' '}
         <code>'use client'</code> directive if using App Router.
       </Notice>
-
       <DocsCodeExample {...helperNextJS} />
 
+      <h4 data-toc-ignore>VS Code Webview Extension</h4>
+      <p>
+        VS Code webviews have special security restrictions that require a
+        different approach. You'll need to configure both the extension side (to
```
mdo:
```suggestion
        different approach. You’ll need to configure both the extension side (to
```

PR #200, opened 2025-12-05
Add AST Caching to Worker Pool

amadeus:
This adds the ability to cache ast responses from the worker so repeat renders don't pay the render cost... hopefully

https://github.com/user-attachments/assets/5274225f-f9fc-4fff-97a9-8ee26bf2f4ff

### DIFF
--- a/apps/docs/app/docs/WorkerPool/WorkerPool.tsx
+++ b/apps/docs/app/docs/WorkerPool/WorkerPool.tsx
@@ -238,7 +238,7 @@ export function WorkerPool({
       <p>
         The worker pool can cache rendered AST results to avoid redundant
         highlighting work. When enabled, subsequent requests for the same file
-        or diff will return cached results immediately instead of re-processing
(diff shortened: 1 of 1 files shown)


apps/docs/app/docs/WorkerPool/WorkerPool.tsx
```diff
@@ -228,6 +230,29 @@ export function WorkerPool({
...
 
+      <h3>Render Cache</h3>
+      <Notice variant="warning" icon={<IconCiWarningFill />}>
+        This is an experimental feature. It is disabled by default while being
+        validated in production use cases.
+      </Notice>
+      <p>
+        The worker pool can cache rendered AST results to avoid redundant
+        highlighting work. When enabled, subsequent requests for the same file
+        or diff will return cached results immediately instead of re-processing
```
mdo:
```suggestion
        or diff will return cached results immediately instead of reprocessing
```

PR #209, opened 2025-12-08
Simplify docs Header and Sidebar

mdo:
Pulled out of #201.


apps/docs/app/page.tsx
```diff
@@ -37,7 +37,7 @@ import { SPLIT_UNIFIED } from './diff-examples/SplitUnified/constants';
 export default function Home() {
   return (
     <div className="mx-auto min-h-screen max-w-5xl px-5 xl:max-w-[80rem]">
-      <HeaderWrapper />
+      <Header className="-mb-[1px]" />
```
amadeus, line 40: what does the prefixed `-` mean here?

mdo: That means a negative value. So this is `margin-bottom: -1px`. Tailwind is fucking wild.

PR #214, opened 2025-12-08
New prose styles and updated mobile sidebar

mdo:
- Removes the tailwind typography plugin
- Adds new `prose.css` file for the docs page with a subset of typography-related styles
- Removes some wrapping `<section>` elements from content files (this is needed because I use flexbox for the spacing in `.docs-prose`)
- Includes some basic responsive type changes so that content is a little easier to read on mobile. This includes a base font-size reset from 14px to 16px, plus some `clamp()` use for heading font sizes.
- Modified the sidebar sticky styles to smooth out some bouncing, and also tightened up spacing here.
- Changed the mobile menu from a drawer to a popover to avoid some odd styling on new iOS.

### DIFF
--- a/apps/docs/app/globals.css
+++ b/apps/docs/app/globals.css
@@ -208,7 +208,7 @@
 
 /* Docs sidebar */
 .docs-sidebar {
-  /* Mobile: fixed dropdown panel */
+  /* Mobile popover menu */
   @apply bg-background fixed top-16 right-4 left-4 z-[60];
   @apply max-h-[80dvh] overflow-y-auto;
   @apply -translate-y-2;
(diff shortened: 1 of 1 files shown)


apps/docs/app/docs/Overview/Overview.tsx
```diff
@@ -104,41 +104,26 @@ export function Overview({
...
-            className="flex-1 sm:flex-initial"
-            value="patch-file"
-          >
-            Patch file
-          </ButtonGroupItem>
+          <ButtonGroupItem value="single-file">Single file</ButtonGroupItem>
+          <ButtonGroupItem value="patch-file">Patch file</ButtonGroupItem>
         </ButtonGroup>
       </div>
-      <DocsCodeExample {...file} key={`${type}-${example}`} />
```
amadeus, line 141: oh -- you need to maintain this key (maybe this is where some of the perf stuff i saw was stemming from)

mdo: I think Claude may have removed this or something, wasn't intentional.

amadeus: tfw claudster

PR #224, opened 2025-12-10
Update buttons, button group, and HP hero

mdo:
More coming, but…

- Refined buttons
- Simpler button group styles—no extra padding around them
- Updated hero tagline and more

### DIFF
--- a/apps/docs/app/docs/ReactAPI/ComponentTabs.tsx
+++ b/apps/docs/app/docs/ReactAPI/ComponentTabs.tsx
@@ -80,7 +80,7 @@ export function SharedPropTabs({
       <ButtonGroup
         value={sharedProps}
         onValueChange={(value) => setSharedProps(value as SharedPropsTypes)}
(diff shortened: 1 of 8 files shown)


apps/docs/app/docs/ReactAPI/ComponentTabs.tsx
```diff
@@ -80,7 +80,7 @@ export function SharedPropTabs({
       <ButtonGroup
         value={sharedProps}
         onValueChange={(value) => setSharedProps(value as SharedPropsTypes)}
-        className="no-scrollbar max-w-full overflow-x-auto"
+        className="no-scrollbar max-w-full overflow-x-auto md:overflow-visible"
```
amadeus, line 83: how come we need md:overflow-visible here?

mdo: That's me removing the overflow, just in case, when we don't need it on wider viewports. Might not be necessary though.

PR #244, opened 2025-12-19
Virtualization Foundation

amadeus:
This branch started off with me building out a few virtualization test cases, and as I've learned more about the requirements, I realized there are various foundational pieces that needed to be addressed before I can ship a virtualization feature.

This PR technically includes a few Virtualization components, however are not ready for prime time and should not be used at this time.

The goal with this PR however is to get it merged into main in a state where I can potentially do a release with these foundational changes to make sure they are thoroughly tested by people as I continue working on virtualization.

### Revamped `FileDiffMetadata`

The original data structure was mostly built as a convenience wrapper for early diff rendering, however it was not very efficient or useful and required a lot of additional data to be derived as well as doubled up a lot of existing data (i.e. we would include multiple copies of the contents, both as a top level array of strings and in the hunk structure itself).

The new type looks something like this:

```ts
export interface FileDiffMetadata {
  name: string;
  prevName?: string;
  languageOverride?: SupportedLanguages;
  mode?: string;
  prevMode?: string;
  type: ChangeTypes;

  hunks: Hunk[];
  splitLineCount: number;
  unifiedLineCount: number;

  isPartial: boolean;
  deletionLines: string[];
  additionLines: string[];
  cacheKey?: string;
}

export interface Hunk {
  collapsedBefore: number;

  additionStart: number;
  additionCount: number;
  additionLines: number;
  additionLineIndex: number;

  deletionStart: number;
  deletionCount: number;
  deletionLines: number;
  deletionLineIndex: number;

  hunkContent: (ContextContent | ChangeContent)[];
  hunkContext: string | undefined;
  hunkSpecs: string | undefined;

  splitLineStart: number;
  splitLineCount: number;

  unifiedLineStart: number;
  unifiedLineCount: number;

  noEOFCRDeletions: boolean;
  noEOFCRAdditions: boolean;
}

export interface ContextContent {
  type: 'context';
  lines: number;
  additionLineIndex: number;
  deletionLineIndex: number;
}

export interface ChangeContent {
  type: 'change';
  deletions: number;
  deletionLineIndex: number;
  additions: number;
  additionLineIndex: number;
}
```

Content diff content now lives at the top level of the data structure and the nested Hunk/Change/Content structures are simply derived data to work with these top level arrays of strings. This is to enable more efficient calculations, not needlessly duplicate data and require significantly fewer variables when iterating through the data structure.

### New Shared `iterateOverDiff` Helper

Whether it's highlighiting, rendering or eventually computing virtualization data, there are many similar but different use cases that require us to iterate through our `FileDiffMetadata` data structure. As i was building out various virtualization prototypes, it was quite common for me to deal with subtle bugs in how these different usecases would calculate things.  Not to mention it would continue to be a maintenance nightmare as we move forward with adding new features, etc.

Essentially `iterateOverDiff` takes a callback, an instance of `FileDiffMetadata`, and potentially some other helper props and will iterate over each line (or within windowed situations).  It's built to keep overall iterations to a minimum, allowing to skip lines that shouldn't be rendered/computed/highlighted. In this PR it's been fully implemented into `DiffHunksRenderer` and our `renderDiffWithHighlighter` method. 

### Misc Virtualization/DOM Performance Optimizations

Originally these components were built to have a very simple snapshot style rendering. if we needed to update or re-render the diff, we'd just full scale re-render everything. This is something that won't work well with virtualization as we'll be doing many DOM operations while scrolling. This PR does not include all the performance optimizations I need to make, but it does include a decent handful:

* Annotations are now smarter about re-rendering
* `position: sticky` can have a real performance impact and so I've taken a pass only apply sticky in cases where we have `overflow: scroll`
* I've made a few CSS tweaks related to virtualization, to ensure things stay a bit more stable under various windowed scenarios
* Various managers should be and interactivity classes should be smarter about only doing work when they need to
* WorkerPoolManager should be a bit smarter about handling a larger task queue and if components get mounted and then unmounted while in a highlight queue

### Virtualization Pieces

Given the size of this PR, it was going to be too difficult to properly remove these unfinished virtualization components. So I am including them in this merge. They are not fully usable in their current state, but I will be continuing to work on them after merge.

When I do a release with these changes, I will probably make a specialized beta branch that removes the exports for them and does the release to ensure that people don't try using them. There's also a whole bunch of docs code that I had built out for testing monster diffs that I've removed, but will re-vert back in a follow up branch as I continue development.

### Misc

* Removed the `fast-deep-equal` dependency
  * It didn't provide a whole lot, plus I found it more helpful to just create special object comparators for things.
* The `lang` prop on `FileDiffMetadata` and `FileContents` was confusing people as they often assumed it was required. I've renamed it to `languageOverride` to be explicit about it's intent. I've also tweaked the docs that mention it


packages/diffs/src/components/SimpleVirtualizedFileDiff.ts
```diff
@@ -0,0 +1,582 @@
...
+      const deletionRemaining =
+        fileDiff.deletionLines.length -
+        (lastHunk.deletionLineIndex + lastHunk.deletionCount);
+      if (lastHunk != null && additionRemaining !== deletionRemaining) {
+        throw new Error(
+          `LittleVirtualizedFileDiff: trailing context mismatch (additions=${additionRemaining}, deletions=${deletionRemaining}) for ${fileDiff.name}`
+        );
+      }
+      const trailingRangeSize = Math.min(additionRemaining, deletionRemaining);
+      if (lastHunk != null && trailingRangeSize > 0) {
```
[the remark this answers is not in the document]
amadeus: lmao no idea what you're on about here... lastHunk is nullable...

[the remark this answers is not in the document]
amadeus: Lmao, go home, you're drunk.

apps/demo/index.html
```diff
@@ -20,11 +20,11 @@
           <button id="load-diff">Load Large-ish Diff</button>
           <button id="worker-load-diff">Worker Render Diff</button>
           <label>
-            <input id="unified" type="checkbox" />
+            <input id="unified" type="checkbox" checked />
```
amadeus: this is just the demo site, and yeah, made it easier for me to test things

apps/docs/app/docs/CoreTypes/constants.ts
```diff
@@ -33,16 +33,18 @@ interface FileContents {
...
 };
 
 // With explicit language override
 const jsonFile: FileContents = {
-  name: 'config', // No extension, so we specify lang
+  // No extension, so we specify languageOverride
+  name: 'config',
   contents: '{ "key": "value" }',
-  lang: 'json',
+  languageOverride: 'json',
```
fat: i would personally not change this

packages/diffs/src/components/AdvancedVirtualizedFileDiff.ts
```diff
@@ -0,0 +1,228 @@
...
+    // Add header height
+    if (!disableFileHeader) {
+      this.unifiedHeight += DIFF_HEADER_HEIGHT;
+      this.splitHeight += DIFF_HEADER_HEIGHT;
+    } else {
+      this.unifiedHeight += FILE_GAP;
+      this.splitHeight += FILE_GAP;
+    }
+
+    // NOTE(amadeus): I wonder if it's worth shortcutting this? I it might help
```
fat: `I it`

packages/diffs/src/components/AdvancedVirtualizer.ts
```diff
@@ -0,0 +1,297 @@
+import { DEFAULT_THEMES, DIFFS_TAG_NAME, FILE_GAP } from '../constants';
+import { queueRender } from '../managers/UniversalRenderingManager';
+import type { ParsedPatch } from '../types';
+import { createWindowFromScrollPosition } from '../utils/createWindowFromScrollPosition';
+import type { WorkerPoolManager } from '../worker';
+import { AdvancedVirtualizedFileDiff } from './AdvancedVirtualizedFileDiff';
+import type { FileDiffOptions } from './FileDiff';
+
+// FIXME(amadeus): REMOVE ME
```
fat: 🔪

amadeus: I just moved these to a specialized types file since they are somewhat shared between virtualizers, also renamed the types to be a bit less goofy, and share a `__` prefix. While i do want to keep them around for development, in the beta release that I do will remove these

packages/diffs/src/components/AdvancedVirtualizedFileDiff.ts
amadeus: fixed

apps/docs/app/docs/CoreTypes/constants.ts
amadeus: my hot, wont-die-on-the-hill-about-it take:

* it's an easy thing for people to fix
* it should hopefully help train people that they don't really need this prop in practice
* figured i would do it while changing a bunch of types already.

That said, if more of you guys think it's better to stay consistent, i can revert this.

amadeus: @SlexAxton ur turn to weigh in 👯

packages/diffs/src/utils/diffAcceptRejectHunk.ts
```diff
@@ -92,11 +102,38 @@ export function diffAcceptRejectHunk(
...
+        deletionOffset === 0
+      ) {
+        break;
+      }
     } else {
       hunk.splitLineStart += splitOffset;
       hunk.unifiedLineStart += unifiedOffset;
+
       hunk.additionStart += additionOffset;
+      hunk.additionLineIndex += additionOffset;
```
fat, line 120: Is it possible to have offsets that aren't positive?  (e.g., accepting a hunk that removes lines)

packages/diffs/src/components/SimpleVirtualizer.ts
```diff
@@ -0,0 +1,558 @@
...
+      passive: true,
+    });
+    contentContainer ??= this.root.firstElementChild ?? undefined;
+    if (contentContainer != null) {
+      this.resizeObserver?.observe(contentContainer);
+    }
+  }
+
+  cleanUp(): void {
+    this.intersectionObserver?.disconnect();
```
fat, line 133: should this be removing scroll/resize listeners and observer maps too?

amadeus: This component is not finished, so most likely yes. Would need a lot more work

packages/diffs/src/utils/diffAcceptRejectHunk.ts
amadeus: yeah, very much depends on how it works out from:

```
      additionOffset = lineCount - hunk.additionCount;
```

so if it is negative, the += would properly acommodate it

packages/diffs/src/components/File.ts
```diff
@@ -311,48 +328,72 @@ export class File<LAnnotation = undefined> {
...
-      this.fileContainer.appendChild(this.hoverContent);
+    if (this.fileContainer == null || renderHoverUtility == null) {
+      return;
     }
     const element = renderHoverUtility(this.mouseEventManager.getHoveredLine);
-    this.hoverContent.innerHTML = '';
-    if (element != null) {
-      this.hoverContent.appendChild(element);
+    if (element != null && this.hoverContent != null) {
+      return;
```
fat, line 388: im mb reading wrong… but is it possible for renderHoverUtility to return a different element, but for us to show stale element bc element and hoverContent both exist?

packages/diffs/src/utils/diffAcceptRejectHunk.ts
fat: i guess if the change removes lines, and then those offsets would be negative and code on 125 wouldn't run?

Which would mean the  additionLineIndex / deletionLineIndex inside hunk.hunkContent stay at their old positions even though the file got shorter…

packages/diffs/src/components/SimpleVirtualizer.ts
fat: ah ok my b my b

packages/diffs/src/components/File.ts
amadeus: So this is mostly a VanillaJS API thing. The React api doesn't render through this path as it has a top level prop that uses react (and would fundamentally work how React works with regards to the render function and components and how they render).

Basically the problem is as follows -- in virtualized situations we may be re-rendering a lot when scrolling, and when it comes to rendering content in slots, i've found it to be quite expensive to needlessly re-render them. The way I originally designed this API was to try to be optimized and stable (it's why it doesn't re-render on each line you hover, for example).

Anyways, this change was part of a broader set of changes to make a lot of the slotted APIs more stable. If you're using the vanilla JS api and you need more reactivity for stuff, then you should probably do that further down the chain, not at this level (i.e. some other class that ultimately gets rendered that has some sort of lifecycle and reactivity component that you setup separately, if that makes sense).

fat: image

packages/diffs/src/utils/diffAcceptRejectHunk.ts
amadeus: ahh yeah good point, i should compare against 0

amadeus: Good catch

packages/diffs/src/components/File.ts
amadeus: The TL;DR: React doesn't apply here, and that's probably all you care about.

apps/docs/app/docs/CoreTypes/constants.ts
SlexAxton: I don't mind *changing* it, but I might go with something like `forceLanguage` too as another option. I do think it's better since the default behavior is almost always what people want. Presumably we'll call this a breaking change, but if by some magic there are no other breaking changes in this beast, then we could deprecate `lang` for now rather than remove it.

apps/demo/src/utils/createWorkerAPI.ts
```diff
@@ -13,7 +13,7 @@ export function createWorkerAPI(
       workerFactory() {
         return new Worker(WorkerUrl, { type: 'module' });
       },
-      poolSize: 8,
+      poolSize: 3,
```
SlexAxton, line 16: should we detect the best thing here, or is 3 always good?

packages/diffs/src/components/AdvancedVirtualizer.ts
```diff
@@ -0,0 +1,288 @@
...
+    private workerManager?: WorkerPoolManager | undefined
+  ) {
+    this.stickyOffset = document.createElement('div');
+    this.stickyOffset.style.contain = 'layout size';
+    this.stickyContainer = document.createElement('div');
+    this.stickyContainer.style.contain = 'strict';
+    this.stickyContainer.style.position = 'sticky';
+    this.stickyContainer.style.width = '100%';
+    this.stickyContainer.style.contain = 'strict';
+    this.stickyContainer.style.isolation = 'isolate';
```
SlexAxton, line 56: theoretically if we don't care about strong style CSPs, which right now we mostly aren't compatible with, you *could* do `this.stickyContainer.style.cssText = "contain: layout size; …"` - but i don't think you should do this, just interesting comments to show that im reading.

packages/diffs/src/components/File.ts
```diff
@@ -311,48 +328,72 @@ export class File<LAnnotation = undefined> {
...
-      for (const annotation of this.lineAnnotations) {
-        const content = renderAnnotation(annotation);
-        if (content == null) continue;
-        const el = createAnnotationWrapperNode(
-          getLineAnnotationName(annotation)
-        );
-        el.appendChild(content);
-        this.annotationElements.push(el);
-        this.fileContainer.appendChild(el);
+      for (const [index, annotation] of this.lineAnnotations.entries()) {
```
SlexAxton, line 348: reminds me that we should officially decide on our browser support matrix and make sure we're not introducing syntax that isn't transpiled or supported

packages/diffs/src/components/File.ts
```diff
@@ -404,18 +444,22 @@ export class File<LAnnotation = undefined> {
...
-      return;
-    }
-    if (this.headerElement != null) {
-      container.shadowRoot?.replaceChild(newHeader, this.headerElement);
-    } else {
-      container.shadowRoot?.prepend(newHeader);
+    const headerHTML = toHtml(headerAST);
+    if (headerHTML !== this.lastRenderedHeaderHTML) {
+      const tempDiv = document.createElement('div');
+      tempDiv.innerHTML = headerHTML;
```
SlexAxton, line 450: Future: since we will more or less always need to inject raw html into the page, we're also a little scarier for pages with strict CSPs. We could document the `require-trusted-types-for 'script'` CSP directive and then allow users to inject the sanitization method of their choosing.

packages/diffs/src/components/FileDiff.ts
```diff
@@ -232,39 +258,60 @@ export class FileDiff<LAnnotation = undefined> {
...
       this.fileContainer.shadowRoot.innerHTML = '';
     }
     this.fileContainer = undefined;
-    this.pre = undefined;
+    // Manually help garbage collection
+    if (this.pre != null) {
+      this.pre.innerHTML = '';
+      this.pre = undefined;
+    }
+    this.codeUnified = undefined;
```
SlexAxton, line 283: would `delete this.codeUnified` be better? Would somehow storing these things in a weak map instead be better on memory?

packages/diffs/src/components/SimpleVirtualizedFileDiff.ts
```diff
@@ -0,0 +1,582 @@
...
+  LAnnotation = undefined,
+> extends FileDiff<LAnnotation> {
+  override readonly __id: string = `little-virtualized-file-diff:${++instanceId}`;
+
+  public top: number | undefined;
+  public height: number = 0;
+  public override fileDiff: FileDiffMetadata | undefined = undefined;
+  // Sparse map: view-specific line index -> measured height
+  // Only stores lines that differ what is returned from `getLineHeight`
+  private heightCache: Map<number, number> = new Map();
```
SlexAxton, line 37: Is this leaky? should we have a cleanup func for this that clears it? would a weakmap work if we don't want that?

packages/diffs/src/components/SimpleVirtualizedFileDiff.ts
```diff
@@ -0,0 +1,582 @@
...
+    if (lastHunk != null && hasFinalHunk(fileDiff)) {
+      const additionRemaining =
+        fileDiff.additionLines.length -
+        (lastHunk.additionLineIndex + lastHunk.additionCount);
+      const deletionRemaining =
+        fileDiff.deletionLines.length -
+        (lastHunk.deletionLineIndex + lastHunk.deletionCount);
+      if (lastHunk != null && additionRemaining !== deletionRemaining) {
+        throw new Error(
+          `LittleVirtualizedFileDiff: trailing context mismatch (additions=${additionRemaining}, deletions=${deletionRemaining}) for ${fileDiff.name}`
```
SlexAxton: Error message still uses an old name for this.

packages/diffs/src/renderers/DiffHunksRenderer.ts
```diff
@@ -130,11 +107,15 @@ export interface HunksRenderResult {
   baseThemeType: 'light' | 'dark' | undefined;
 }
 
+let instanceId = -1;
```
SlexAxton, line 110: have been scared by this in my code and have been considering swapping to a random number/hash. theoretically its fine if the code is never executed twice, but even on the server vs client boundary it seems a lil dangerous to hope that these don't collide ever.

packages/diffs/src/style.css
```diff
@@ -467,7 +470,29 @@
       var(--diffs-bg-buffer) calc(4px * 1.414)
     );
     min-height: 1lh;
-    width: var(--diffs-column-width, auto);
+    /* lol not sure this is worth it... */
```
SlexAxton: never `...` always `…` (option semicolon)

packages/diffs/src/style.css
```diff
@@ -625,23 +658,35 @@
...
+    margin-left: var(--diffs-gap-inline, var(--diffs-gap-fallback));
     border-top-right-radius: 0;
     border-bottom-right-radius: 0;
   }
 
+  [data-type='file'][data-overflow='scroll'],
+  [data-type='split'][data-overflow='scroll'] {
+    [data-deletions] [data-separator='line-info'] [data-separator-wrapper] {
+      left: var(--diffs-gap-inline, var(--diffs-gap-fallback));
+    }
```
SlexAxton, line 670: do we transpile this?

SlexAxton: nevermind, it looks like support is better than i would have guessed: https://caniuse.com/css-nesting

apps/demo/src/utils/createWorkerAPI.ts
amadeus: This is demo code so it doesn't really matter in a general context, but I definitely found that virtualizing in extreme scenarios, using 8 threads would just crash chrome eventually. So as a general learning it's probably good to document a lower number, 3 was the maximum it would seem to not crash but give you best parallelization.

packages/diffs/src/components/File.ts
amadeus: Oh yeah, i've def been going ham with things like for...of loops in various contexts

packages/diffs/src/components/FileDiff.ts
amadeus: Storing these refs are mostly just improvements for more incremental dom updates where possible. I'm not opposed to moving over to something like a weak map, but yeah, would probably do that as a general purpose pass after things are mostly setup.

As for the delete vs undefined thing... its not really necessary for us to remove the key from the class, and i think technically we don't want to do that, we just want to clear out the reference.

packages/diffs/src/components/SimpleVirtualizedFileDiff.ts
amadeus: It's just because this component is not really fully done yet (and won't be part of the next beta release to test the foundational changes of this code). But yes, i do plan to make sure cleanUp does all the right things (probably in the next big phase PR for this

packages/diffs/src/components/SimpleVirtualizedFileDiff.ts
amadeus: Good catch, will fix

packages/diffs/src/renderers/DiffHunksRenderer.ts
amadeus: These instance id's aren't really used for anything practically. It's really just been for debugging when i need to log lots of instances and can keep track of which is which. So server vs client interop is not something that has been needed with this

packages/diffs/src/style.css
amadeus: Honestly it would probably be good if we did have some sort of CSS transpiler step actually, would love to make this as dumb as CSS as possible... but obvi doesn't need to happen here

packages/diffs/src/style.css
amadeus: fixed

PR #252, opened 2026-01-05
fix(diffs): use catalog reference for @shikijs/core to ensure consistent version resolution

NisargIO:
## Problem

When installing `@pierre/diffs` in a project that has Shiki 3.20.0 defined, the `@shikijs/core` dependency would resolve to version 3.19.0 instead of the expected 3.20.0, causing version mismatches.

## Root Cause

The `@shikijs/core` dependency in `packages/diffs/package.json` was using a semver range (`^3.0.0`) instead of the `catalog:` reference that other shiki packages were using:

```diff
"dependencies": {
-  "@shikijs/core": "^3.0.0",
+  "@shikijs/core": "catalog:",
   "@shikijs/engine-javascript": "catalog:",
   "@shikijs/transformers": "catalog:",
   ...
}
```

This inconsistency caused Bun to resolve `@shikijs/core` to 3.19.0 from the stale lockfile entry, ignoring the catalog-defined version of 3.20.0. Meanwhile, other shiki packages (`@shikijs/engine-javascript`, `@shikijs/transformers`, `shiki`) correctly used `catalog:` and resolved to 3.20.0.

## Fix

- Changed `@shikijs/core` to use `catalog:` reference, consistent with other shiki packages
- Refreshed `bun.lock` to resolve all shiki packages to the catalog-defined version (3.20.0)

## Impact

This ensures downstream consumers get consistent shiki versions across all `@shikijs/*` packages when installing `@pierre/diffs`.

### DIFF
--- a/bun.lock
+++ b/bun.lock
@@ -123,7 +123,7 @@
       "name": "@pierre/diffs",
       "version": "1.0.2",
       "dependencies": {
-        "@shikijs/core": "^3.0.0",
+        "@shikijs/core": "catalog:",
         "@shikijs/engine-javascript": "catalog:",
         "@shikijs/transformers": "catalog:",
         "diff": "catalog:",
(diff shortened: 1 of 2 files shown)


amadeus:
fixed in #253 

thanks for bringing up the issue! 🙏

PR #260, opened 2026-01-07
add visibleRanges support for FileRenderer

BrianHung:
### Description

This PR adds collapse/expand functionality to the File component, mirroring the existing diff hunks UI/UX.

New FileRenderer API:
- `visibleRanges: LineRange[]` - specify which line ranges to show
- `expansionLineCount: number` - lines to reveal per expand click  
- `hunkSeparators: HunkSeparators` - separator style for collapsed regions

I've also added a demo to `apps/demo/index.html`

### Motivation & Context

I would like to have same UI/UX for collapsed or hidden lines, but for FileRenderer.

Screenshot
CleanShot 2026-01-07 at 02 45 31@2x

### Type of changes

- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would change existing functionality)
- [ ] Documentation update

### Checklist

- [ ] I have read the
      [contributing guidelines](https://github.com/pierre-computer/diffs/blob/main/CONTRIBUTING.md)
- [ ] All new and existing tests pass (`bun run diffs:test`)

### DIFF
--- a/apps/demo/index.html
+++ b/apps/demo/index.html
@@ -15,6 +15,7 @@
         <div class="tools">
           <button id="toggle-theme">Toggle Theme</button>
           <button id="render-file">Render File</button>
+          <button id="visible-ranges">Visible Ranges</button>
           <button id="stream-code">Stream Code</button>
           <button id="diff-files">Diff Two Files</button>
           <button id="load-diff">Load Large-ish Diff</button>
--- a/apps/demo/src/main.ts
+++ b/apps/demo/src/main.ts
@@ -559,6 +559,37 @@ if (renderFileButton != null) {
   });
 }
 
+const visibleRangesButton = document.getElementById('visible-ranges');
+if (visibleRangesButton != null) {
+  visibleRangesButton.addEventListener('click', () => {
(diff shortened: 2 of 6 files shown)


amadeus:
This is definitely an interesting API idea, however it's something that we definitely can't accept until I've shipped virtualization, since that will have a major impact on how things are rendered and I don't want to add any complexity to the existing APIs until I have that API shipped.

I'd also want to spend some time thinking about this API a bit more though, I'm not sure an array of expanded regions is the right idea because it muddies the intention of the File component. It adds a significant amount of complexity to the File renderer which I'm not sure I want to burden that class with.

That said, you could probably actually get this functionality today by still using FileDiff and instead generating a custom FileDiffMetadata that only exposes hunks that are context only lines.

Also for new features, we do want to be pretty intentional about what we add, so I added a bit more color to our pull requests template that helps communicate this going forward.

amadeus:
Just note that when i ship virtualization, it will be a largely breaking change for whatever you might come up with now to do this, however, it should be a lot easier to make the new FileDiffMetadata objects AFTER i ship that change since it won't have weird intermixed content

PR #261, opened 2026-01-07
Update NextJS to 16.1.1

amadeus:
I've been having a lot of issues with turbopack recently where it seems to hard crash during development requiring a server restart. Apparently 16.1.1 brings some major updates to turbopack so hoping this helps. Should be a pretty harmless update otherwise.


apps/docs/package.json
```diff
@@ -70,7 +70,7 @@
     "typescript": "catalog:"
   },
   "overrides": {
-    "@types/react": "19.2.2",
-    "@types/react-dom": "19.2.2"
+    "@types/react": "19.2.7",
+    "@types/react-dom": "19.2.3"
```
amadeus, line 74: Kinda wondering why we need these tbh, but not sure... wonder if they could be set to just `catalog:` tho...

PR #267, opened 2026-01-10
Live demo on theme, add docs

mdo:
Big WIP, needs a lot of work on the the docs most of all.

### DIFF
--- a/apps/docs/app/theme/ThemeDemo.tsx
+++ b/apps/docs/app/theme/ThemeDemo.tsx
@@ -2,6 +2,7 @@
 
 import {
   IconCheckCheck,
+  IconChevronsNarrow,
   IconColorDark,
   IconColorLight,
   IconFileCode,
@@ -11,7 +12,7 @@ import { cn } from '@/lib/utils';
 import { parseDiffFromFile, preloadHighlighter } from '@pierre/diffs';
 import { File, FileDiff } from '@pierre/diffs/react';
 import { useTheme } from 'next-themes';
-import { useEffect, useMemo, useState } from 'react';
+import { useEffect, useMemo, useRef, useState } from 'react';
 
 // Preload themes at module level for earliest possible start
 void preloadHighlighter({
@@ -307,6 +308,8 @@ export function ThemeDemo() {
   const [colorMode, setColorMode] = useState<'light' | 'dark'>('dark');
   const [activeTab, setActiveTab] = useState<TabId>('typescript');
   const [mounted, setMounted] = useState(false);
+  const [dropdownOpen, setDropdownOpen] = useState(false);
+  const dropdownRef = useRef<HTMLDivElement>(null);
 
   // Store raw file contents - these get modified as changes are accepted/rejected
   const [workingFiles, setWorkingFiles] = useState<WorkingFile[]>(() =>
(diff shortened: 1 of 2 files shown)


amadeus:
The page looks really dope, the big question i have is around `here's our theme` vs `here's how to build a custom theme`. I worry the discoverability of the custom theme-ing stuff will be tricky

apps/docs/app/docs/Theming/content.mdx
```diff
@@ -0,0 +1,194 @@
...
+
+For Shiki/web use only, you can skip publishing and simply host your theme JSON
+files or bundle them with your application.
+
+---
+
+#### Credit
+
+Thanks again to the Primer team at GitHub for their work on the open source
+[GitHub VS Code themes](https://github.com/primer/github-vscode-theme). We
```
amadeus, line 193: should this be VSCode? I see it in a couple other places too, but ignore if wrong

apps/docs/app/theme/ThemeDemo.tsx
```diff
@@ -0,0 +1,596 @@
...
+        ? 'text-neutral-400 hover:text-neutral-300'
+        : 'text-neutral-500 hover:text-neutral-700',
+      tabIndicator: isDark ? 'bg-blue-400' : 'bg-blue-500',
+      headerText: isDark ? 'text-neutral-300' : 'text-neutral-700',
+      buttonSecondary: isDark
+        ? 'bg-neutral-700 text-neutral-300 hover:bg-neutral-600'
+        : 'bg-neutral-200 text-neutral-700 hover:bg-neutral-300',
+      buttonPrimary: isDark
+        ? 'bg-blue-600 text-white hover:bg-blue-500'
+        : 'bg-blue-500 text-white hover:bg-blue-400',
```
amadeus, line 396: tailwind is wild, lol

apps/docs/app/theme/ThemeDemo.tsx
```diff
@@ -0,0 +1,596 @@
...
+      setColorMode(resolvedTheme);
+    }
+  }, [resolvedTheme]);
+
+  // Preload themes on mount
+  useEffect(() => {
+    void preloadHighlighter({
+      themes: ['pierre-dark', 'pierre-light'],
+      langs: ['tsx', 'html', 'css'],
+    });
```
amadeus: You should probably just do this at the file/module level, not in useEffect to start it as early as possible.

apps/docs/app/globals.css
```diff
@@ -77,6 +77,7 @@
   --sidebar-accent-foreground: oklch(0.205 0 0);
   --sidebar-border: oklch(0.922 0 0);
   --sidebar-ring: oklch(0.708 0 0);
+  --mint-500: #00cab1;
```
amadeus: 👀

packages/diffs/src/highlighter/shared_highlighter.ts
```diff
@@ -128,3 +128,15 @@ registerCustomTheme('pierre-light', () => {
...
+  return import(
+    '../themes/pierre-dark-vibrant.json'
+  ) as unknown as Promise<ThemeRegistrationResolved>;
+});
+
+registerCustomTheme('pierre-light-vibrant', () => {
+  return import(
+    '../themes/pierre-light-vibrant.json'
+  ) as unknown as Promise<ThemeRegistrationResolved>;
+});
```
amadeus: are we bundling 2 new themes now?

amadeus (commented):
I think overall changes look good to me. Some questions on VSCode spelling but i might just be dumb.

Is it intentional that we are adding 2 more themes to the diffs project? It will make the bundle bigger, so wondering if it's worth the tradeoff?

Also I think I'm still kinda on my bullshit from my earlier comment -- I'm a bit worried the whole docs stuff for theming may not be very discoverable here? Maybe I am overthinking it tho.

apps/docs/app/docs/Theming/content.mdx
mdo: this is the casing and style they use in their repo, so I've left as that. technically VS Code brand guidelines say don't shorten it, use Visual Studio Code. But alas, meh.

apps/docs/app/theme/ThemeDemo.tsx
mdo: Truly.

apps/docs/app/globals.css
mdo: Ah good eyes I can remove this because I'm no longer using it.

packages/diffs/src/highlighter/shared_highlighter.ts
mdo: I had a toggle for vibrant/regular in the live demo, decided to remove it though because I think I want to revisit the bright colors and make sure they're as good as can be.

PR #274, opened 2026-01-14
Add support for custom css variable themes

amadeus:
A quick pass to expose functionality from `createCssVariablesTheme`

The basic usage is:

```ts
registerCustomCSSVariableTheme('my-css-theme', {
  foreground: '#d4d4d4',
  background: '#1e1e1e',
  'token-comment': '#6a9955',
  'token-string': '#ce9178',
  'token-keyword': '#569cd6',
  'token-function': '#dcdcaa',
  'token-variable': '#9cdcfe',
  'token-constant': '#4fc1ff',
  'token-punctuation': '#d4d4d4',
});
```

### DIFF
--- a/packages/diffs/src/utils/areOptionsEqual.ts
+++ b/packages/diffs/src/utils/areOptionsEqual.ts
@@ -1,7 +1,6 @@
-import type { FileDiffOptions } from 'src/components/FileDiff';
-import { DEFAULT_THEMES } from 'src/constants';
-import type { FileOptions } from 'src/react';
-
+import type { FileDiffOptions } from '../components/FileDiff';
+import { DEFAULT_THEMES } from '../constants';
+import type { FileOptions } from '../react';
 import { areObjectsEqual } from './areObjectsEqual';
 import { areThemesEqual } from './areThemesEqual';
 


packages/diffs/src/utils/areOptionsEqual.ts
```diff
@@ -1,7 +1,6 @@
-import type { FileDiffOptions } from 'src/components/FileDiff';
-import { DEFAULT_THEMES } from 'src/constants';
-import type { FileOptions } from 'src/react';
-
+import type { FileDiffOptions } from '../components/FileDiff';
+import { DEFAULT_THEMES } from '../constants';
+import type { FileOptions } from '../react';
```
amadeus, line 3: This was just cleaning up some absolute import derps that I noticed. it's generally good to fix these when they happen because i've seen issues in various bundled scenarios where it breaks things

packages/diffs/src/index.ts
```diff
@@ -1,3 +1,8 @@
+import {
+  codeToHtml,
+  createCssVariablesTheme as createCSSVariablesTheme,
```
amadeus, line 3: The biggest of nits, but if we export this for others, we should match the conventions of the rest of our file when referencing CSS

PR #277, opened 2026-01-17
[file-tree] mvp - development mode only

SlexAxton:
adds the mvp for the file-tree component. the page `/file-tree` is available when in development.

### DIFF
--- a/apps/docs/app/file-tree/ClientPage.tsx
+++ b/apps/docs/app/file-tree/ClientPage.tsx
@@ -37,16 +37,17 @@ export function ClientPage({
   }, [collapseFolders]);
 
   return (
-    <>
-      <div className="w-2/3">
-        <h2>React SSR File Tree</h2>
-        <FileTreeReact
-          options={fileTreeOptions}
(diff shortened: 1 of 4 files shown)


packages/diffs/test/__snapshots__/parsePatchFiles.test.ts.snap
```diff
@@ -71159,104 +71260,3 @@ exports[`parsePatchFiles should warn on malformed patch with bare newline in hun
...
-        "prevName": undefined,
-        "splitLineCount": 9,
-        "type": "change",
-        "unifiedLineCount": 10,
-      },
-    ],
-    "patchMetadata": undefined,
-  },
-]
-`;
```
amadeus, line 71262: I wonder why your PR changed the order in this snapshot? not a big deal, just weird 🤔

SlexAxton: I bet it’s a lint or race condition

PR #278, opened 2026-01-20
Add Support for ANSI

amadeus:
There were a couple pieces to supporting this.

* It was a specialized language like `'text'` that was already built it
  * This means we had to skip resolving it like `'text'` but still allow the string to be passed through
* ANSI is only supported as a language override
  * Not sure there's a proper extension for ANSI?
* In order to support the background colors I needed to rework some of the token naming conventions to allow ansi background colors to be properly supported
  * This involved renaming the css variable prefix on token variables so as not to potentially inherit the global bg, which makes a lot of other things quite annoying. I am pretty happy with this solution and it shouldn't really have any downstream effects I don't think (i.e. pretty safe)
  * This PR also adds support for text decorations and bg colors in themes if supported

CleanShot 2026-01-19 at 16 07 58

**NOTE:** The PR is beefy because we had to update a bunch of snapshots


apps/demo/src/mocks/fileAnsi.txt
```diff
@@ -0,0 +1,9 @@
+[0;32mcolored foreground[0m
+[0;42mcolored background[0m
+
+[0;1mbold text[0m
+[0;2mdimmed text[0m
+[0;4munderlined text[0m
+[0;7mreversed text[0m
+[0;9mstrikethrough text[0m
+[0;4;9munderlined + strikethrough text[0m
```
amadeus, line 9: Just including this so it's easy to test ANSI stuff in the future.

PR #293, opened 2026-01-26
Add husky pre-commit hook

necolas:
### Description

Uses [husky](https://typicode.github.io/husky/) and `lint-staged` to automatically format and lint js/ts files on commit. Reformatted files are re-staged. As configured, lint failures prevent commit. Use `git commit --no-verify` to skip

### Motivation & Context

Avoid the need for contributors to manually run the scripts each commit.

### DIFF
--- a/package.json
+++ b/package.json
@@ -39,7 +39,8 @@
     "lint:css": "stylelint \"**/*.css\"",
     "lint:css:fix": "stylelint \"**/*.css\" --fix",
     "format": "prettier --write .",
-    "format:check": "prettier --check ."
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -158,6 +159,7 @@
     "@types/bun": "catalog:",
     "eslint": "catalog:",
     "eslint-plugin-react-hooks": "catalog:",
+    "husky": "^9.1.7",
```
amadeus: This would need to be added to the workspace, and then added as a `catalog:`

PR #298, opened 2026-01-27
Add collapsedContextThreshold 

fat:
🏄

### DIFF
--- a/apps/docs/app/docs/VanillaAPI/constants.ts
+++ b/apps/docs/app/docs/VanillaAPI/constants.ts
@@ -150,8 +150,8 @@ const instance = new FileDiff({
   expansionLineCount: 100,
 
   // Auto-expand collapsed context regions at or below this size
-  // (default: 2)
-  collapsedContextThreshold: 2,
+  // (default: 1)
+  collapsedContextThreshold: 1,
 
   // ─────────────────────────────────────────────────────────────
   // INLINE CHANGE HIGHLIGHTING


packages/diffs/src/components/SimpleVirtualizedFileDiff.ts
```diff
@@ -202,6 +203,9 @@ export class SimpleVirtualizedFileDiff<
       expandedHunks: expandUnchanged
         ? true
         : this.hunksRenderer.getExpandedHunksMap(),
+      collapsedContextThreshold:
+        this.options.collapsedContextThreshold ??
+        DEFAULT_COLLAPSED_CONTEXT_THRESHOLD,
```
amadeus: kind of a nit, but can we pull this value out above from `this.options` on line 190? Since we do that with a couple of other props out of options already?

packages/diffs/src/components/SimpleVirtualizedFileDiff.ts
```diff
@@ -315,6 +319,19 @@ export class SimpleVirtualizedFileDiff<
         renderAll: false,
       };
     }
+    const threshold = Math.max(
```
amadeus: Why do we need Math.max for this? The value check on threshold and rangeSize should be sufficient?

packages/diffs/src/components/SimpleVirtualizedFileDiff.ts
```diff
@@ -315,6 +319,19 @@ export class SimpleVirtualizedFileDiff<
...
+      this.options.collapsedContextThreshold ??
+        DEFAULT_COLLAPSED_CONTEXT_THRESHOLD,
+      0
+    );
+    if (threshold > 0 && rangeSize <= threshold) {
```
amadeus: Also since this line is essentially the same return as expandUnchanged, can we just combine this if statement with the one below?

```
if (expandUnchanged || rangeSize <= collapsedContextThreshold)
```


Might also be a nice opportunity to destruct both expandUnchanged and collapsedContextThreshold out at the same time with default values.

packages/diffs/src/components/SimpleVirtualizedFileDiff.ts
```diff
@@ -442,6 +459,9 @@ export class SimpleVirtualizedFileDiff<
       expandedHunks: expandUnchanged
         ? true
         : this.hunksRenderer.getExpandedHunksMap(),
+      collapsedContextThreshold:
+        this.options.collapsedContextThreshold ??
+        DEFAULT_COLLAPSED_CONTEXT_THRESHOLD,
```
amadeus: Small nit to destruct this higher up with the other `this.options` above

packages/diffs/src/renderers/DiffHunksRenderer.ts
```diff
@@ -350,7 +356,8 @@ export class DiffHunksRenderer<LAnnotation = undefined> {
             ? true
             : expandUnchanged
               ? true
-              : this.expandedHunks
+              : this.expandedHunks,
+          this.getOptionsWithDefaults().collapsedContextThreshold
```
amadeus: This a good spot where i did a dumb. On line 325, i use this.options, I should've used `this.getOptionsWithDefaults()`. then can  you destruct both this value and `expandUnchanged` up there?

packages/diffs/src/utils/iterateOverDiff.ts
```diff
@@ -56,6 +57,7 @@ export function iterateOverDiff({
   startingLine = 0,
   totalLines = Infinity,
   expandedHunks,
+  collapsedContextThreshold,
```
amadeus: Question here -- if 2 is our default, should we default prop this to that constant?

```
collapsedContextThreshold = 2
```

That or imo we should default arg it to 0.  Kinda thinking if we go all in on 2, then everywhere we default arg it to 2 for consistency?

Then the rest of the functions internally here should just require a number, and this should satisfy them. Then the call sites to use it are cleaner.

packages/diffs/src/utils/iterateOverDiff.ts
```diff
@@ -503,7 +507,8 @@ function getExpandedRegion(
   isPartial: boolean,
   rangeSize: number,
   expandedHunks: Map<number, HunkExpansionRegion> | true | undefined,
-  hunkIndex: number
+  hunkIndex: number,
+  collapsedContextThreshold: number | undefined
```
amadeus: This becomes just `collapsedContextThreshold: number`

packages/diffs/src/utils/iterateOverDiff.ts
```diff
@@ -514,6 +519,15 @@ function getExpandedRegion(
...
+  const threshold = Math.max(collapsedContextThreshold ?? 0, 0);
+  if (threshold > 0 && rangeSize <= threshold) {
+    return {
+      fromStart: rangeSize,
+      fromEnd: 0,
+      rangeSize,
+      collapsedLines: 0,
+    };
+  }
   if (expandedHunks === true) {
```
amadeus: Similar comment from before, i don't think we need a max here, and if it's a `: number` then we can just do the:

```
if (expandedHunks === true || rangeSize <= collapsedContextThreshold)
```

packages/diffs/test/DiffHunksRendererVirtualization.test.ts
```diff
@@ -326,7 +326,7 @@ describe('DiffHunksRenderer - Virtualization', () => {
       // First rendered line index -- it's hard coded because it's a bit hard
       // to figure out, so mostly a good reference if tests change these
       // assumptions
-      expect(unifiedIndices[0]).toBe(186);
+      expect(unifiedIndices[0]).toBe(184);
```
amadeus, line 329: weird that this number went down... 🤔 

all the rest went up, which is what you would expect if they are rendering a few more lines...

packages/diffs/src/utils/renderDiffWithHighlighter.ts
```diff
@@ -41,6 +41,7 @@ export function renderDiffWithHighlighter(
     startingLine,
     totalLines,
     expandedHunks,
+    collapsedContextThreshold,
```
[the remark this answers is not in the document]
amadeus: Probably akin to one of my other comments, but obvi use the constant

packages/diffs/src/utils/iterateOverDiff.ts
amadeus: Ahh yeah -- codex code review brought up a good point that we should probably default this arg in `renderDiffWithHighlighter` as well

packages/diffs/src/style.css
```diff
@@ -1060,7 +1052,6 @@
     }
   }
 
-  [data-line-type='context-expanded'],
   [data-line-type='change-addition'],
   [data-line-type='change-deletion'] {
```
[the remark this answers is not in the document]
fat: yah we changed this :P

packages/diffs/test/DiffHunksRendererVirtualization.test.ts
fat: Image

```
The number going down makes sense. Previously, collapsedContextThreshold was not being defaulted in the rendering path (it was
  undefined), so small collapsed regions (1-2 lines) stayed collapsed and were rendered as a single hunk separator. Now with the
  default threshold of 2, those small regions are expanded into actual lines.

  When a 1-2 line collapsed region expands, it takes up more rendered lines (1-2 lines instead of 1 separator). So when iterating to
   find line index 100 (the startingLine), the iteration reaches 100 rendered lines sooner — meaning the first rendered line
  corresponds to an earlier source line index. That's why 186 dropped to 184: the expanded small regions before line 100 cause the
  window to start at an earlier point in the file.

  The other values going up follows the same logic — more total source lines are being rendered within the same window because
  previously-collapsed lines are now visible.
```

apps/docs/app/docs/ReactAPI/constants.ts
```diff
@@ -77,6 +77,10 @@ interface DiffOptions {
   // Lines revealed per click when expanding collapsed regions
   expansionLineCount: 100,
 
+  // Auto-expand collapsed context regions at or below this size
+  // (default: 1)
```
fat, line 81: I changed it to 1 i think that's better than 2

PR #316, opened 2026-02-06
Tree docs

mdo:
### `@pierre/trees` package

- **Locked paths**: New `lockedPaths` option prevents specified files/directories from being dragged. Locked items display a lock icon (new SVG sprite symbol) aligned to the right of the row.
- **`themeToTreeStyles` utility**: New exported function that converts a Shiki/VS Code theme object into `--ft-theme-*` CSS custom properties, enabling one-line theming from any editor theme.
- **Theme-aware CSS variable fallback chain**: All CSS custom properties now support a three-tier fallback (`--ft-*` explicit → `--ft-theme-*` from theme → hardcoded default), including new tokens for git status colors, muted foreground, input/search background, and selection/focus outlines.
- **`initialSearch` state config**: New `initialSearch` option on `FileTreeStateConfig` (and the React `<FileTree>` component) pre-populates the search field on mount, with a `useLayoutEffect` to apply expand/collapse filtering before the first paint.
- **Search blur behavior**: Search input `onBlur` now defers closing via `setTimeout` and skips closing when focus moves to another file-tree search input, preventing layout thrash when clicking between multiple trees.
- **Search initial-apply logic**: `expand-matches` mode now expands all folders on initial apply (instead of using an empty baseline), so the first render with `initialSearch` shows all matching items visible.
- **Flattened directory name robustness**: `FlattenedDirectoryName` now falls back gracefully when tree item instances aren't available, using `idToPath` lookups and a `fallbackName` prop.
- **Icon `viewBox` override**: `Icon` component now supports per-icon `viewBox` overrides (used by the lock icon whose native viewBox differs from its rendered size).

### `apps/docs` — Trees documentation site

- **New `/trees` landing page**: Product page with hero section and interactive `TreeApp` demo showcasing drag-and-drop, search, git status, theming, accessibility, and flattening.
- **Full documentation section** (`/trees/docs`): New pages for Overview, Installation, Core Types, React API, Vanilla API, File Tree Options, Styling, SSR, and Utilities — each with live code examples.
- **Shiki theme picker demo**: Interactive theme selector that applies VS Code themes to a live file tree using the new `themeToTreeStyles` utility.
- **Homepage updates**: Updated hero and navigation to surface the Trees product alongside Diffs; shared `product-config.ts` for product metadata.
- **Header/Footer updates**: Navigation and layout updated to support multi-product docs structure.

### Viewing locally

- Run `bun ws docs dev`
- Load up `localhost:3001/trees` and `localhost:3001/trees/docs`

### DIFF
--- a/apps/docs/app/layout.tsx
+++ b/apps/docs/app/layout.tsx
@@ -1,5 +1,5 @@
 // sort-imports-ignore
-import type { Metadata } from 'next';
+import type { Metadata, Viewport } from 'next';
 import { ThemeProvider } from 'next-themes';
 import {
   Fira_Code,
@@ -51,6 +51,11 @@ const geistMono = Geist_Mono({
   variable: '--font-geist-mono',
   subsets: ['latin'],
 });
+
+export const viewport: Viewport = {
+  width: 'device-width',
+  initialScale: 1,
+};
 
 export const metadata: Metadata = {
   title: 'Diffs, from Pierre',
--- a/packages/trees/src/style.css
+++ b/packages/trees/src/style.css
@@ -146,6 +146,7 @@
 
   /* LIST ITEM */
   [data-type='item'] {
+    color: inherit;
     font-family: var(--ft-internal-font-family);
     font-size: var(--ft-internal-font-size);
     outline: none;


apps/docs/app/trees-dev/demo-data.ts
```diff
@@ -8,9 +8,9 @@ import type {
 const sampleFileList: string[] = [
   'README.md',
   'package.json',
-  'Build/index.mjs',
-  'Build/scripts.js',
-  'Build/assets/images/social/logo.png',
+  'build/index.mjs',
+  'build/scripts.js',
+  'build/assets/images/social/logo.png',
```
SlexAxton: This was intentional to check alignment for capital letters. I'd suggest we have a different set of data for the docs though.

apps/docs/app/trees/demo-data.ts
```diff
@@ -0,0 +1,38 @@
...
+  '.gitignore',
+];
+
+export const sharedDemoFileTreeOptions: FileTreeOptions = {
+  flattenEmptyDirectories: true,
+  initialFiles: sampleFileList,
+};
+
+export const sharedDemoStateConfig: FileTreeStateConfig = {
+  initialExpandedItems: ['Build/assets/images/social'],
```
SlexAxton, line 34: if you keep the lowercase `build` above, then this would need to be updated.

packages/trees/src/components/Root.tsx
```diff
@@ -59,26 +65,47 @@ const getFilesSignature = (files: string[]): string =>
...
+  }, [flattens, tree, idToPath]);
+
+  if (segments.length === 0) {
+    return (
+      <span data-item-flattened-subitems>
+        {fallbackName.replace(/\//g, ' / ')}
+      </span>
+    );
+  }
+
```
SlexAxton, line 100: do you know the purpose of this change?

packages/trees/src/components/Root.tsx
```diff
@@ -662,6 +706,15 @@ export function Root({
...
   }, [selectedSnapshot, callbacksRef, tree, idToPath]);
 
+  // When tree mounts with initial search in state, run setSearch once so expand/collapse filter is applied.
+  // useLayoutEffect ensures this runs before paint so the first frame shows the correct expansion.
+  useLayoutEffect(() => {
+    const search = tree.getState().search;
+    if (search != null && search.length > 0) {
+      tree.setSearch(search);
+    }
+  }, [tree]);
```
SlexAxton, line 716: nice

packages/trees/src/components/Root.tsx
mdo: I believe Claude added this when I was getting the search demos on the homepage—where flattened directories wouldn't render correctly for them or something.

packages/trees/src/features/fileTreeSearchFeature.ts
```diff
@@ -280,7 +298,21 @@ export const fileTreeSearchFeature: FeatureImplementation = {
...
-      onBlur: () => tree.closeSearch(),
+      onBlur: () => {
+        // Defer so we can see where focus moved. Don't close when focus moved
+        // to another file-tree search input (avoids layout thrash when clicking
+        // between search fields across multiple trees).
+        setTimeout(() => {
+          const active = document.activeElement;
+          if (
+            active != null &&
+            active.getAttribute?.('data-file-tree-search-input') != null
```
SlexAxton: hmm, i think this is too specific to docs. i wonder if there's some version of this that's correct more generally like if the focus moved anywhere outside of the existing tree. but we'd want to make sure that was the desired behavior for things like an editor (moving from search to editing).

packages/trees/src/react/utils/useFileTreeInstance.ts
```diff
@@ -153,6 +162,7 @@ export function useFileTreeInstance({
             // setExpandedItems/setSelectedItems imperatively.
             initialExpandedItems: sp.initialExpandedItems ?? sp.expandedItems,
             initialSelectedItems: sp.initialSelectedItems ?? sp.selectedItems,
+            initialSearch: sp.initialSearch,
```
SlexAxton: i think if we have `inititalSearch` we'll also just want a controlled `search`, which might be the wrong property name for something like that? maybe `searchQuery` or something, and `initialSearchQuery`. I wouldn't consider this one blocking, but we will likely need to move it.

packages/trees/src/components/Root.tsx
SlexAxton: ok, i think it likely would have had to been bad input data that would require this change? (missing segments) - but ultimately i don't mind keeping around code that helps the file tree not blow up if something is weird, so we can keep it.

packages/trees/src/features/fileTreeSearchFeature.ts
SlexAxton: this feels like the only thing i think is "blocking" - but also i would never actually block on an unreleased product.

packages/trees/src/features/fileTreeSearchFeature.ts
```diff
@@ -299,16 +299,19 @@ export const fileTreeSearchFeature: FeatureImplementation = {
         tree.setSearch(target?.value ?? '');
       },
       onBlur: () => {
-        // Defer so we can see where focus moved. Don't close when focus moved
-        // to another file-tree search input (avoids layout thrash when clicking
-        // between search fields across multiple trees).
+        // Defer so we can check where focus landed. If focus stayed inside this
+        // tree (e.g. clicking a tree item, or another input within the shadow
+        // root), keep search open.
```
SlexAxton: i feel like this might have reversed the behavior that you wanted?

PR #318, opened 2026-02-06
Replace local icons with `@pierre/icons`

mdo:
- Removes all local SVGs and React components
- Replaces with new `@pierre/icons` package
- Updates sprite builder to use source svgs found in npm package


apps/demo/package.json
```diff
@@ -18,6 +18,7 @@
   },
   "dependencies": {
     "@pierre/diffs": "workspace:*",
+    "@pierre/icons": "catalog:",
```
amadeus: I don't think this project was using the icons before, so do we even need this here? Not that it should really matter?

Issue #336, opened 2026-02-14
Bug: Mounting and unmounting file diffs in quick succession causes shiki highlighting to stop working


ShpetimA:
### Prerequisites

- [x] I have [searched](https://github.com/pierrecomputer/pierre/issues?q=is%3Aissue) for duplicate or closed issues
- [x] I have read the [contributing guidelines](https://github.com/pierrecomputer/pierre/blob/main/.github/CONTRIBUTING.md)

### Describe the issue

When swapping between diffs in really quick succession the worker pool gets overloaded and shiki highlighting stops working.

### Reproduction

I couldn't upload a video of the issue reproduced here on gh so i uploaded to youtube feel free to check it out 
https://youtu.be/Z5NRmjsQx0A

### What browser(s) are you seeing the problem on?

Safari

### What version of @pierre/diffs are you using?

1.1.0-beta.13

amadeus:
Fixed in https://github.com/pierrecomputer/pierre/releases/tag/v1.1.0-beta.16

PR #337, opened 2026-02-14
Fix: Mounting and unmounting file diffs in quick succession causes shiki highlighting to stop working

ShpetimA:
### Description

Fixed a bookkeeping race condition in WorkerPoolManager where cleaned up in-flight tasks could leave a worker stuck as permanently busy. When a late response arrives for a task already removed from pendingTasks, we now clear managedWorker.request_id if the response id matches that worker, so stale responses are still ignored but worker capacity is correctly released.
### Motivation & Context

Workers were stuck in busy state and not cleaned up properly when mounting and unmounting files very quickly

### Type of changes

      being opened
      that this PR is being opened
- [ ] Documentation update

### Checklist

- [x] I have read the

AI was used to debug the issue and find the worker pool managment problem

### Related issues

#336

### DIFF
--- a/packages/diffs/src/worker/WorkerPoolManager.ts
+++ b/packages/diffs/src/worker/WorkerPoolManager.ts
@@ -754,7 +754,7 @@ export class WorkerPoolManager {
     task?: AllWorkerTasks
   ) {
     managedWorker.request_id = undefined;
-    if (task !== undefined) {
+    if (task != null) {
       if ('instance' in task) {
         this.instanceRequestMap.delete(task.instance);
       }


packages/diffs/src/worker/WorkerPoolManager.ts
```diff
@@ -723,6 +723,11 @@ export class WorkerPoolManager {
 
     if (task != null) {
       this.clearWorkerTask(task, managedWorker);
```
amadeus: Good catch! However I think there's a better way to fix this.

* Lets rename this function to `cleanWorkerAndTask`.
* Swap the argument positions
* Make `task` undefineable in the function definition and remove this `task != null` gating
  * And obvi fix up the internal function to work with the optionality of `task`

ShpetimA: Yeah i agree done

packages/diffs/src/worker/WorkerPoolManager.ts
```diff
@@ -751,12 +749,17 @@ export class WorkerPoolManager {
...
 
-  private clearWorkerTask(task: AllWorkerTasks, managedWorker: ManagedWorker) {
+  private cleanWorkerAndTask(
+    managedWorker: ManagedWorker,
+    task?: AllWorkerTasks
+  ) {
     managedWorker.request_id = undefined;
-    if ('instance' in task) {
-      this.instanceRequestMap.delete(task.instance);
+    if (task !== undefined) {
```
amadeus:
```suggestion
    if (task != null) {
```

smol nit, we use `!= null` or `== null` to handle defined-ness

ShpetimA: done :)

PR #348, opened 2026-02-21
Collapsed Fixes

amadeus:
Some fixes for the collapsed file stuff.

* Reworked it to not depend on a specialized renderRange (for the most part).
* The virtualizer was broken with the collapsing logic, so i had to fix that up

### DIFF
--- a/packages/diffs/src/components/FileDiff.ts
+++ b/packages/diffs/src/components/FileDiff.ts
@@ -693,7 +693,6 @@ export class FileDiff<LAnnotation = undefined> {
     if (collapsed) {
       this.removeRenderedCode();
       this.clearAuxiliaryNodes();
-      this.renderRange = undefined;
 
       try {
         const hunksResult = this.hunksRenderer.renderDiff(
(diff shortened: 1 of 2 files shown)


apps/demo/src/main.ts
```diff
@@ -647,6 +681,23 @@ cleanButton?.addEventListener('click', () => {
...
+  input.type = 'checkbox';
+  input.checked = checked;
+  input.addEventListener('change', () => {
+    onChange(input.checked);
+  });
+  label.dataset.collapser = '';
+  label.appendChild(input);
+  label.append(' Collapse');
+  return label;
+}
```
amadeus, line 699: Just demo B.S. to test collapsing

packages/diffs/src/components/FileDiff.ts
```diff
@@ -764,6 +792,46 @@ export class FileDiff<LAnnotation = undefined> {
...
+    this.bufferBefore = undefined;
+    this.bufferAfter?.remove();
+    this.bufferAfter = undefined;
+
+    this.codeUnified?.remove();
+    this.codeUnified = undefined;
+    this.codeDeletions?.remove();
+    this.codeDeletions = undefined;
+    this.codeAdditions?.remove();
+    this.codeAdditions = undefined;
```
amadeus, line 810: Honestly this \.remove logic is probably not needed? but maybe it's better from a GC perspective? The AI wanted to do it, so i allowed it

PR #349, opened 2026-02-24
Add expandAll/collapseAll to diffs

atamarzban:
## Summary
- Adds programmatic `expandAll()` and `collapseAll()` methods to `FileDiff`, `VirtualizedFileDiff`, and the React `useFileDiffInstance` hook for controlling collapsed "N unmodified lines" sections
- Adds a text toggle button ("Expand all" / "Collapse all") in the file header next to the filename — visible by default, opt-out via `disableExpandAllButton: true`
- Button is automatically hidden when `expandUnchanged` or `disableFileHeader` is set

## API

**Vanilla JS:**
```ts
const diff = new FileDiff(options);
diff.expandAll();   // expand all collapsed sections
diff.collapseAll(); // collapse everything back
```

**React:**
```tsx
const { ref, expandAll, collapseAll } = useFileDiffInstance(options);
```

**Disable the built-in button:**
```ts
const diff = new FileDiff({ disableExpandAllButton: true, ...rest });
```

### DIFF
--- a/packages/diffs/test/__snapshots__/DiffHunksRender.test.ts.snap
+++ b/packages/diffs/test/__snapshots__/DiffHunksRender.test.ts.snap
@@ -2395,6 +2395,14 @@ exports[`DiffHunksRenderer proper buffers should be prepended to additions colum
             "tagName": "div",
             "type": "element",
           },
+          {
+            "children": [],
+            "properties": {
+              "name": "expand-all-button",
+            },
+            "tagName": "slot",
+            "type": "element",
+          },
         ],
         "properties": {
           "data-header-content": "",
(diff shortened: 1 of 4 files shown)


atamarzban:
@amadeus Would appreciate a review! 🙏

amadeus:
Question on this -- shouldn't you be able to accomplish this functionality already with toggling the `expandUnchanged` prop? Or is there something about that API that doesn't work that I'm not seeing.

It's not the cleanest, but right now in the VanillaJS api you could accomplish it with:

```ts
const diff = new FileDiff(options);
diff.setOptions({ ...diff.options, expandUnchanged: true })
diff.setOptions({ ...diff.options, expandUnchanged: false })

// and in react, you would pass in the prop change to options:

<FileDiff
  // ...
  options={{ expandUnchanged: true }}
/>
<FileDiff
  // ...
  options={{ expandUnchanged: false }}
/>
```

PR #352, opened 2026-02-24
Attempt to prevent scrolling in docs

mdo:


SlexAxton (approved):
Seems okay for now. im curious if there's someone who wants this scroll to occur if they're rendering the page with existing search queries.

PR #367, opened 2026-03-03
Fat/merge conflict diff

fat:
Merge conflict built on top of diff primitive…

VSCODE (note the syntax highlighting failure)
CleanShot 2026-03-02 at 20 34 18@2x

Pierre (note the correct highlighting)
CleanShot 2026-03-02 at 20 34 36@2x

### DIFF
--- a/packages/diffs/src/utils/setWrapperNodeProps.ts
+++ b/packages/diffs/src/utils/setWrapperNodeProps.ts
@@ -17,10 +17,12 @@ export function setPreNodeProperties(
 ): HTMLPreElement {
   // NOTE: We always apply custom properties first so the important properties
   // cannot be overridden.
-  for (const key in customProperties) {
-    const value = customProperties[key];
-    if (value != null) {
-      pre.setAttribute(key, `${value}`);
+  if (customProperties != null) {
+    for (const key in customProperties) {
+      const value = customProperties[key];
+      if (value != null) {
+        pre.setAttribute(key, `${value}`);
+      }
     }
   }
   if (type === 'diff') {


packages/diffs/src/components/FileDiff.ts
```diff
@@ -227,10 +217,32 @@ export class FileDiff<LAnnotation = undefined> {
...
+    return {
+      ...options,
+      hunkSeparators:
+        typeof options.hunkSeparators === 'function'
+          ? 'custom'
+          : options.hunkSeparators,
+    };
+  }
+
+  protected createHunksRenderer(
```
amadeus, line 237: Why do we need this? (leaving a note for myself for later)

packages/diffs/src/components/UnresolvedFile.ts
```diff
@@ -0,0 +1,105 @@
...
+    const fileDiff = this.getOrCreateUnresolvedFileDiff(file);
+    return super.render({
+      ...rest,
+      fileDiff,
+    });
+  }
+
+  private getOrCreateUnresolvedFileDiff(file: FileContents): FileDiffMetadata {
+    const cache = this.unresolvedFileDiffCache;
+    if (cache != null && cache.file === file) {
```
amadeus: Probably should use `areFilesEqual` for this check

packages/diffs/src/utils/getMergeConflictLineTypes.ts
```diff
@@ -0,0 +1,114 @@
...
+              ? completedFrame.baseMarkerLineIndex + 1
+              : undefined,
+        });
+      }
+      continue;
+    }
+
+    lineTypes[index] = frame.stage;
+  }
+
```
amadeus, line 119: Should we do any sort of validation here, if there's no end marker on the last item should we throw an error?

packages/diffs/src/utils/getMergeConflictLineTypes.ts
```diff
@@ -0,0 +1,114 @@
...
+}
+
+export function getMergeConflictActionLineNumber(
+  conflict: MergeConflictRegion
+): number {
+  return Math.max(1, conflict.startLineNumber - 1);
+}
+
+function parseMergeConflicts(lines: string[]): MergeConflictParseResult {
+  const lineTypes = new Array<MergeConflictLineType>(lines.length).fill('none');
```
amadeus: random thought, to reduce an iteration -- what if we don't do the fill here, and then fill in the loop below?

packages/diffs/src/utils/parseMergeConflictDiffFromFile.ts
```diff
@@ -0,0 +1,226 @@
...
+  const patchLines: string[] = [];
+  const currentLineNumbersByOriginalIndex: (number | undefined)[] = new Array(
+    lines.length
+  );
+  const incomingLineNumbersByOriginalIndex: (number | undefined)[] = new Array(
+    lines.length
+  );
+  let currentLineNumber = 0;
+  let incomingLineNumber = 0;
+  for (let index = 0; index < lines.length; index++) {
```
amadeus, line 92: This would be a 3rd iteration of all the lines, is there a clever way we could merge this iteration with the one in `getMergeConflictLineTypes`?

packages/diffs/src/utils/getMergeConflictLineTypes.ts
amadeus: Actually, looking further, just leave it undefined, because you do the old `?? 'none'` in the other stuff.

packages/diffs/src/utils/parseMergeConflictDiffFromFile.ts
```diff
@@ -0,0 +1,226 @@
...
+  return annotations;
+}
+
+export function parseMergeConflictDiffFromFile(
+  file: FileContents
+): ParseMergeConflictDiffFromFileResult {
+  const lines = splitFileContents(file.contents);
+  const lineTypes = getMergeConflictLineTypes(lines);
+  const currentLines: string[] = [];
+  const incomingLines: string[] = [];
```
amadeus: Another set of iterations that we should get rid of. If you make this a string, and just += the string, then you'll make sure that `createResolvedConflictFile` doesn't have to iterate _again_ over all the lines.

Of course you'd need some sort of counter to manage the line counts since you depend on that later on.

packages/diffs/src/utils/parseMergeConflictDiffFromFile.ts
```diff
@@ -0,0 +1,226 @@
...
+    incomingLines
+  );
+  const patch = createMergeConflictPatch({
+    name: file.name,
+    patchLines,
+    currentLineCount: currentLines.length,
+    incomingLineCount: incomingLines.length,
+  });
+
+  const fileDiff = processFile(patch, {
```
amadeus, line 195: Theoretically we can run this patch through `trimPatchContext` if you wanted to not show the entire file, and minimize the resolutions to the relevant conflicts with a specified number of context lines. It defaults to 10, but we could make it configurable (i forced it to 2 to better show it)

Image

It does appear to present a bug when resolving, however... (the whole file becomes collapsed), but that would probably be fixable fairly easily to re-build the diff properly...)

Image

(basically you have to re-build the diff to have context-only hunks, which is not too hard)

packages/diffs/src/utils/parseMergeConflictDiffFromFile.ts
```diff
@@ -0,0 +1,226 @@
...
+    throwOnError: true,
+  });
+
+  if (fileDiff == null) {
+    throw new Error(
+      'parseMergeConflictDiffFromFile: failed to build merge conflict diff metadata'
+    );
+  }
+
+  const actions = getMergeConflictRegions(lines).map((conflict) => {
```
amadeus: Similar comment from earlier. If we could compute actions while iterating over all the lines, it will really help the performance of this in large files.

packages/diffs/src/utils/getMergeConflictActionsUnsafeCSS.ts
```diff
@@ -0,0 +1,15 @@
+const MERGE_CONFLICT_ACTIONS_UNSAFE_CSS = `
+[data-line-annotation],
+[data-gutter-buffer='annotation'] {
+  --diffs-line-bg: var(--diffs-bg);
+}
+`;
```
amadeus: lmao ok we definitely should not do this... 

we should maybe add some sort of data attribute to the `pre` element and then just put this into the main css file.

This is definitely some real AI slop

packages/diffs/src/renderers/DiffHunksRenderer.ts
```diff
@@ -612,17 +673,23 @@ export class DiffHunksRenderer<LAnnotation = undefined> {
...
-        type,
-        createGutterItem(lineType, lineNumber, lineIndex)
-      );
-    }
+      lineIndex: string,
+      sourceType: 'context' | 'context-expanded' | 'change' = 'context',
+      metadata: unknown = undefined
+    ) => {
+      const item = createGutterItem(lineType, lineNumber, lineIndex);
+      this.decorateGutterItem({
```
amadeus: I don't love this, we create a gutter item, then go in and modify it... why can't we just update `createGutterItem` to take stuff as necessary, with some optionality and just do the right thing?

packages/diffs/src/renderers/UnresolvedFileHunksRenderer.ts
```diff
@@ -0,0 +1,164 @@
...
+    lineType,
+    additionLineRaw,
+    deletionLineRaw,
+  }: UnifiedLineDecorationProps): LineDecoration {
+    const mergeConflictType =
+      type === 'change'
+        ? lineType === 'change-deletion'
+          ? 'current'
+          : 'incoming'
+        : (getMergeConflictMarkerType(additionLineRaw) ??
```
amadeus: Shouldn't we have some pre-computed data somewhere that can tell us this info instead of needing to check the line content again? Like maybe ideally we can just pass in the `additionLineIndex` or `deletionLineIndex` to get this?

packages/diffs/src/renderers/UnresolvedFileHunksRenderer.ts
```diff
@@ -0,0 +1,164 @@
...
+
+type MergeConflictMarkerType =
+  | 'marker-start'
+  | 'marker-base'
+  | 'marker-separator'
+  | 'marker-end'
+  | 'current'
+  | 'incoming';
+
+interface MergeConflictLineDecorationMetadata {
```
amadeus: Why do we need a whole interface for a singular type?

packages/diffs/src/renderers/UnresolvedFileHunksRenderer.ts
```diff
@@ -0,0 +1,164 @@
...
+      type === 'change'
+        ? lineType === 'change-deletion'
+          ? 'current'
+          : 'incoming'
+        : (getMergeConflictMarkerType(additionLineRaw) ??
+          getMergeConflictMarkerType(deletionLineRaw));
+    return {
+      gutterLineType: type === 'change' ? 'context' : lineType,
+      metadata: {
+        mergeConflictType,
```
amadeus: Shouldn't we be able to merge this into the top level object?

packages/diffs/src/renderers/UnresolvedFileHunksRenderer.ts
```diff
@@ -0,0 +1,164 @@
...
+    const mergeConflictType =
+      type === 'change'
+        ? side === 'deletions'
+          ? 'current'
+          : 'incoming'
+        : getMergeConflictMarkerType(lineRaw);
+    return {
+      gutterLineType: type === 'change' ? 'context' : type,
+      metadata: {
+        mergeConflictType,
```
amadeus: Similar note here, we probably should just have both of these props be top level of the return object

packages/diffs/src/renderers/UnresolvedFileHunksRenderer.ts
```diff
@@ -0,0 +1,164 @@
...
+      } satisfies MergeConflictLineDecorationMetadata,
+    };
+  }
+
+  protected override decorateGutterItem({
```
amadeus: Ideally I think we extend the function that actually generate the gutter element to add these properties, instead of creation and modification imo

packages/diffs/src/renderers/UnresolvedFileHunksRenderer.ts
```diff
@@ -0,0 +1,164 @@
...
+    metadata,
+  }: GutterDecorationProps): void {
+    const mergeConflictType = getMergeConflictType(metadata);
+    if (mergeConflictType == null) {
+      return;
+    }
+    gutterItem.properties['data-merge-conflict'] = mergeConflictType;
+  }
+
+  protected override decorateContentLine({
```
amadeus: Similar not here, don't love that we create and modify these content lines, Ideally we just one shot them in the function that does everything.

packages/diffs/src/components/FileDiff.ts
fat: so that we can subclass and override 👼

packages/diffs/src/components/UnresolvedFile.ts
fat: smart - fixed!

packages/diffs/src/utils/getMergeConflictLineTypes.ts
fat: Ordering here is already deterministic so don't think we need validation.

Also i dont think we should throw here… keeps the parser resilient in the case where someone has a bad merge etc

packages/diffs/src/utils/getMergeConflictLineTypes.ts
fat: I moved none to the loop below. can't leave it undefined i think bc it's only used in some consumers

packages/diffs/src/utils/getMergeConflictLineTypes.ts
amadeus: well i mean -- isn't it possible someone has a conflict with no end marker?

fat: yah i just mean we shouldn't panic in that case

packages/diffs/src/utils/parseMergeConflictDiffFromFile.ts
fat: pushed some optimizations here…

packages/diffs/src/utils/parseMergeConflictDiffFromFile.ts
fat: pushed some optimizations here…

packages/diffs/src/utils/parseMergeConflictDiffFromFile.ts
fat: fixed

packages/diffs/src/utils/getMergeConflictActionsUnsafeCSS.ts
fat: lol good catch

fat: CleanShot 2026-03-03 at 18 01 47@2x

fat: changed this

packages/diffs/src/renderers/DiffHunksRenderer.ts
fat: changed

packages/diffs/src/renderers/UnresolvedFileHunksRenderer.ts
fat: yup - flattened

packages/diffs/src/renderers/UnresolvedFileHunksRenderer.ts
fat: yup fixed

PR #396, opened 2026-03-13
add dangerous css option

fat:


SlexAxton:
also needs a rebase, but good with me (though i'd strongly discourage anyone use this soon, everything is probably going to change at least once, once it's all wired together and stable).

PR #397, opened 2026-03-14
match on path

fat:
match across entire path… e.g. `docs/foo.tsx`

https://github.com/user-attachments/assets/8a80b5f3-afec-476f-8b47-201c9d14550d

### DIFF
--- a/apps/docs/app/trees/tree-examples/SearchSection.tsx
+++ b/apps/docs/app/trees/tree-examples/SearchSection.tsx
@@ -53,7 +53,8 @@ export function SearchSection() {
         title="Search and filter by name"
         description={
           <>
-            Filter the tree by typing in the search field. Search across file paths and names. Trees includes three{' '}
+            Filter the tree by typing in the search field. Search across file
+            paths and names. Trees includes three{' '}
             <Link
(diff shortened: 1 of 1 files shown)


packages/trees/src/features/fileTreeSearchFeature.ts
```diff
@@ -40,9 +41,56 @@ type FileTreeSearchConfig = {
...
+};
+
+const normalizeTreePath = (path: string): string =>
+  normalizeSearchText(
+    path.startsWith(FLATTENED_PREFIX)
+      ? path.slice(FLATTENED_PREFIX.length)
+      : path
+  );
+
+const isFuzzySubsequenceMatch = (search: string, target: string): boolean => {
```
SlexAxton, line 80: I'll let this merge since search will already fall over on massive file lists, but this probably meaningfully reduces the performance a bit further. I think we'll need crazier things for large datasets either way though, so this is probably better for norm-core cases.

SlexAxton:
will need a rebase

packages/trees/src/features/fileTreeSearchFeature.ts
fat: yah i was thinking same!  thanks for review

PR #409, opened 2026-03-18
add support for trailing slash empty dirs

fat:


packages/trees/src/utils/fileListToTree.ts
```diff
@@ -153,7 +165,7 @@ export function fileListToTree(
           ...(flattenedChildren != null && { flattened: flattenedChildren }),
         },
       };
-    } else if (tree[path] == null) {
+    } else {
```
fat, line 168: this loop is iterating folderChildren, so every path in that loop is definitely a directory

PR #415, opened 2026-03-23
[trees] pull in truncation lib for more control over truncation

SlexAxton:
loosely configured for now.

image

### DIFF
--- a/packages/trees/src/components/OverflowText.tsx
+++ b/packages/trees/src/components/OverflowText.tsx
@@ -1,24 +1,156 @@
-import type { ReactNode } from 'react';
+/** @jsxImportSource preact */
 
-import {
-  splitByIndex,
-  splitCenter,
-  splitExtension,
-  splitFirst,
-  splitLast,
-  splitLeafPath,
-} from '../../lib/splits';
-import type {
-  CustomSplitFn,
+import type { ComponentChildren, CSSProperties } from 'preact';
+
+type PropsWithChildren<T = {}> = T & {
+  children?: ComponentChildren;
+};
+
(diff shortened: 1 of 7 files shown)


lucifer-yf:
@SlexAxton 
image
image


When flattenEmptyDirectories: true
is enabled and the tree is rendered
in a narrow container (e.g. a
sidebar panel pushed to the right
edge), the per-segment truncation
produces a very poor visual result
— every segment gets clipped
individually, leaving a string of
disconnected ellipses like a... /
b... / c... / dd... / .... This
makes the path nearly unreadable
and wastes space with repeated ...
markers.

VS Code handles this well by
truncating the entire flattened
path as a single string at the end
(aaa / bbb / ccc /...), which keeps
as much of the path visible as
possible and is far easier to scan
at a glance. Supporting
end-truncation for the whole path
would bring @pierre/trees in line
with this established UX pattern
and significantly improve usability
in space-constrained layouts.

SlexAxton:
@lucier-yf im pulled off on other things for another week or so, but I have a WIP to simplify truncation here:

https://github.com/pierrecomputer/pierre/pull/939

I do think there's a middle ground where the *last* directory is still visible rather than seeing the *least* significant* middle directories. But I mostly agree the current thing is not ideal for very deep folders.

lucifer-yf:
@SlexAxton Now I see that many people have raised this requirement. But has it been implemented in the end?
https://github.com/pierrecomputer/pierre/pull/997
https://github.com/pierrecomputer/pierre/pull/1068
https://github.com/pierrecomputer/pierre/issues/941

PR #446, opened 2026-03-27
Ability to trim hunks while resolving

amadeus:
This was mostly AI slopped, it's unclear if i want to keep it in its current state, but it is a proof of concept...


amadeus:
Going to close this, it hasn't really been something that has been requested a whole lot, and it's a decent amount of work to get right.

Issue #420, opened 2026-03-25
Feature: Expose token level hover actions, similar to what we have now for lines


ShpetimA:
### Prerequisites

- [x] I have [searched](https://github.com/pierrecomputer/pierre/issues?q=is%3Aissue) for duplicate or closed feature requests
- [x] I have read the [contributing guidelines](https://github.com/pierrecomputer/pierre/blob/main/.github/CONTRIBUTING.md)

### Proposal

I would like to have token based hover actions on top of the line level ones that we have atm.

For example something like: 
onTokenEnter
onTokenLeave
onTokenClick

I have a fork of the feature here https://github.com/ShpetimA/pierre/tree/pierre-token , I was curious if you guys would add something like this in the future or I could even create a pr for this.

https://github.com/user-attachments/assets/b105122e-08ed-45f7-a3ba-deaaeda3686b

###  Lsp integration with annotation (have):
The reason why I want this feature is because atm I'm working on a local code reviewing tool and I'm starting to add lsp support on top, atm with the current api I can only add annotations which are good for showing diagnostic information : 

<img width="1184" height="911" alt="Image" src="https://github.com/user-attachments/assets/6859968e-b77f-4d10-a85d-be1408d75a5d" />

### Additional lsp actions (missing):
But as additional features I also want to add lsp information on hover like showing **type definitions**, **gtd** **gtr** etc. To achieve this I need these new onToken callbacks or something to that extent.

PR #449, opened 2026-03-28
feature: Add token level actions and css styling selector options

ShpetimA:
### Description

This PR adds token-level interaction support to the diffs package. File and diff views can now detect hover and click events on individual syntax-highlighted tokens instead of only whole lines, which also enables features like css selectors and styling on token level elements.

#### Changes

- Added token metadata to rendered syntax spans using data-token, data-char-start, and data-char-end.
- Introduced token interaction types and callbacks for onTokenEnter, onTokenLeave, and onTokenClick.
- Updated InteractionManager to resolve token targets, track hovered tokens independently from hovered lines, and expose getHoveredToken().
- Applied a data-token-hovered attribute to the active token so token-level hover state can be styled or inspected.
- Trimmed leading and trailing horizontal whitespace out of interactive token spans so only the actual token content becomes hoverable/clickable.
- Updated snapshots across file rendering, diff rendering, virtualization, patch rendering, and annotation tests to reflect the new token span structure.

The reason for separating the whitespaces inside createTransformerWithState was because Shiki without transforming includes white spaces which would look weird if you would try to apply css styling on token level for ex : 

#### Trimmed Tokens:

https://github.com/user-attachments/assets/82210970-d5fe-4b8e-b808-f23217472f2e

#### Non Trimmed Tokens:

https://github.com/user-attachments/assets/8ed6eebb-db80-49cd-9639-0f629c236468

Ran also benchmarking for the shiki transformer and its not adding any overhead:

image

### Motivation & Context

As I requested in the #420 this feature would allow things like lsp integration on hover, click interactions on the diff viewer.

### Type of changes

      being opened
      that this PR is being opened
- [ ] Documentation update

### Checklist

- [x] I have read the

AI was used during the creation of this pr to better understand the codebase and also do boilerplate changes like writing some of the types that are just repeating etc. Every line of code has been checked by me no vibe coding :)

### Related issues

#420

### DIFF
--- a/apps/demo/src/main.ts
+++ b/apps/demo/src/main.ts
@@ -354,7 +354,7 @@ function renderDiff(parsedPatches: ParsedPatch[], manager?: WorkerPoolManager) {
         //     'enter',
         //     props.tokenText,
         //     props.lineNumber,
-        //     props.charStart
+        //     props.lineCharStart
         //   );
         //   props.tokenElement.style.backgroundColor = 'light-dark(black, white)';
         //   props.tokenElement.style.color = 'light-dark(white, black)';
@@ -365,7 +365,7 @@ function renderDiff(parsedPatches: ParsedPatch[], manager?: WorkerPoolManager) {
         //     'leave',
         //     props.tokenText,
         //     props.lineNumber,
-        //     props.charStart
+        //     props.lineCharStart
         //   );
         //   props.tokenElement.style.backgroundColor = '';
         //   props.tokenElement.style.color = '';
@@ -757,7 +757,7 @@ if (renderFileButton != null) {
       //     'enter',
       //     props.tokenText,
       //     props.lineNumber,
-      //     props.charStart
+      //     props.lineCharStart
       //   );
       //   props.tokenElement.style.backgroundColor = 'light-dark(black, white)';
       //   props.tokenElement.style.color = 'light-dark(white, black)';
@@ -768,7 +768,7 @@ if (renderFileButton != null) {
       //     'leave',
       //     props.tokenText,
       //     props.lineNumber,
-      //     props.charStart
+      //     props.lineCharStart
       //   );
       //   props.tokenElement.style.backgroundColor = '';
       //   props.tokenElement.style.color = '';
@@ -825,7 +825,7 @@ if (renderFileConflictButton != null) {
         //     'enter',
         //     props.tokenText,
         //     props.lineNumber,
-        //     props.charStart
+        //     props.lineCharStart
         //   );
         //   props.tokenElement.style.backgroundColor = 'light-dark(black, white)';
         //   props.tokenElement.style.color = 'light-dark(white, black)';
@@ -836,7 +836,7 @@ if (renderFileConflictButton != null) {
         //     'leave',
         //     props.tokenText,
         //     props.lineNumber,
-        //     props.charStart
+        //     props.lineCharStart
         //   );
         //   props.tokenElement.style.backgroundColor = '';
         //   props.tokenElement.style.color = '';
--- a/packages/diffs/src/managers/InteractionManager.ts
+++ b/packages/diffs/src/managers/InteractionManager.ts
@@ -18,8 +18,8 @@ import { createGutterUtilityElement } from '../utils/createGutterUtilityElement'
 
 interface TokenCache {
   tokenElement: HTMLElement;
-  charStart: number;
-  charEnd: number;
+  lineCharStart: number;
+  lineCharEnd: number;
   tokenText: string;
 }
 
@@ -112,8 +112,8 @@ interface ResolvedTokenTarget<TMode extends InteractionManagerMode> {
   splitLineIndex: number | undefined;
   tokenElement: HTMLElement;
   tokenText: string;
-  charStart: number;
-  charEnd: number;
+  lineCharStart: number;
+  lineCharEnd: number;
 }
 
 export interface MergeConflictActionTarget {
@@ -1309,8 +1309,8 @@ export class InteractionManager<TMode extends InteractionManagerMode> {
   }
 
   private toTokenEventBaseProps({
-    charEnd,
-    charStart,
+    lineCharEnd,
+    lineCharStart,
     lineNumber,
     side,
     tokenElement,
@@ -1319,8 +1319,8 @@ export class InteractionManager<TMode extends InteractionManagerMode> {
     if (this.mode === 'file') {
       return {
         type: 'token',
-        charEnd,
-        charStart,
+        lineCharEnd,
+        lineCharStart,
         lineNumber,
         tokenElement,
         tokenText,
@@ -1329,8 +1329,8 @@ export class InteractionManager<TMode extends InteractionManagerMode> {
 
     return {
       type: 'token',
-      charEnd,
-      charStart,
+      lineCharEnd,
+      lineCharStart,
       lineNumber,
       side,
       tokenElement,
@@ -1414,15 +1414,20 @@ export class InteractionManager<TMode extends InteractionManagerMode> {
         const startAttr = element.getAttribute('data-char');
 
         if (startAttr != null) {
-          const charStart = Number.parseInt(startAttr, 10);
-          if (!Number.isNaN(charStart)) {
+          const lineCharStart = Number.parseInt(startAttr, 10);
+          if (!Number.isNaN(lineCharStart)) {
             const tokenText = element.textContent ?? '';
-            const charEnd = charStart + tokenText.length;
+            const lineCharEnd = lineCharStart + tokenText.length;
             if (
               tokenText.trim() !== '' ||
               this.options.enableTokenInteractionsOnWhitespace === true
             ) {
-              tokenInfo = { tokenElement, charStart, charEnd, tokenText };
+              tokenInfo = {
+                tokenElement,
+                lineCharStart,
+                lineCharEnd,
+                tokenText,
+              };
             }
             continue;
           }
--- a/packages/diffs/src/types.ts
+++ b/packages/diffs/src/types.ts
@@ -534,8 +534,8 @@ export interface DiffLineEventBaseProps extends Omit<
 export interface TokenEventBase {
   type: 'token';
   lineNumber: number;
-  charStart: number;
-  charEnd: number;
+  lineCharStart: number;
+  lineCharEnd: number;
   tokenText: string;
   tokenElement: HTMLElement;
 }


packages/diffs/src/react/utils/renderDiffChildren.tsx
```diff
@@ -31,6 +34,7 @@ interface RenderDiffChildrenProps<LAnnotation, T> {
   ): ReactNode;
   lineAnnotations: DiffBasePropsReact<LAnnotation>['lineAnnotations'];
   getHoveredLine(): GetHoveredLineResult<'diff'> | undefined;
+  getHoveredToken(): GetHoveredTokenResult<'diff'> | undefined;
```
amadeus: This method is never used, so no need to pass it through

packages/diffs/src/react/utils/renderFileChildren.tsx
```diff
@@ -21,6 +24,7 @@ interface RenderFileChildrenProps<LAnnotation> {
   renderGutterUtility: FileProps<LAnnotation>['renderGutterUtility'];
   renderHoverUtility: FileProps<LAnnotation>['renderHoverUtility'];
   getHoveredLine(): GetHoveredLineResult<'file'> | undefined;
+  getHoveredToken(): GetHoveredTokenResult<'file'> | undefined;
```
amadeus: Similar note, this method is never used, so no need to pass it through.

packages/diffs/src/react/utils/useFileDiffInstance.ts
```diff
@@ -126,13 +128,19 @@ export function useFileDiffInstance<LAnnotation>({
...
+  }, []);
+
   const getHoveredLine = useCallback(():
     | GetHoveredLineResult<'diff'>
     | undefined => {
     return instanceRef.current?.getHoveredLine();
   }, []);
 
-  return { ref, getHoveredLine };
+  return { ref, getHoveredLine, getHoveredToken };
```
amadeus: Which, by proxy from the earlier notes, there should be no need to export this which means no changes required in these files (this file and `useFileInstance` and `useUnresolvedFileInstance`).

packages/diffs/src/react/File.tsx
```diff
@@ -48,6 +48,7 @@ export function File<LAnnotation = undefined>({
     renderHoverUtility,
     lineAnnotations,
     getHoveredLine,
+    getHoveredToken,
```
amadeus: These files also won't need any change (File, FileDiff, MultiFileDiff, PatchDiff and UnresolvedFileDiff)

packages/diffs/src/utils/createTransformerWithState.ts
```diff
@@ -39,6 +40,12 @@ export function createTransformerWithState(
         }
         return pre;
       },
+      span(hast, _line, _col, _lineElement, token) {
+        if (token?.offset != null && token.content != null) {
+          return processToken(hast, token);
+        }
+        return hast;
+      },
```
amadeus: Regardless of performance impact (and I do think it's likely there's one on extremely large files), the impact to dom/html/text size is very real. We should only apply this hook if the onToken hooks are setup. `useTokenTransform` should be the boolean we feed to this function and we should only pass the span transformer through if that boolean is true.

For context, a quick test on a code file of around 700 lines, we went from an innerHTML size of: `320,108` chars, to `550,671` chars

Fixing this will also mean we don't need to update any of the test snapshots either.

packages/diffs/src/utils/processToken.ts
```diff
@@ -0,0 +1,80 @@
...
+  if (leadingWhitespace.length > 0) {
+    node.children.push(createTextNodeElement(leadingWhitespace));
+  }
+  node.children.push(
+    createHastElement({
+      tagName: 'span',
+      properties: {
+        'data-token': text,
+        'data-char-start': start,
+        'data-char-end': end,
```
amadeus: So I got a couple concerns about stuff here, but will break it into a few different comments.

Do we need `data-token`? It feels like we are essentially duplicating the content of `textContent` which I'm not a fan of. (It's an important reason why I want these transforms to be opt in only, since it could essentially be more than doubling the size of our dom output for code)

amadeus: Another question I have -- do we actually need `data-char-end`? Can we not derive it from `data-char-start + textContent.length?` It could be something that could potentially helps reduced DOM size from these transforms.

amadeus: Additionally, something else I've just discovered that I'm not really sure how to solve -- our line diff type checker will split up tokens in a way that seems to make the text value of `data-token` and/or `textContent` of the span invalid (for the token itself). See this video showing how stuff can be broken up:

https://github.com/user-attachments/assets/61e0c4b6-00a8-4764-9b63-e2b0ffa052df

So my question here, for the tools that integrate these APIs, is the full value of the token important, or is it more important to simply detect character position within the document, to let the LSP do it's thing?

If it's just the character position from a token (even if it's split up, I'm actually wondering if there's a more lightweight way we could do this with simply some basic metrics where each line gets the character start of that line, the pointer/mouse movement events we can simply calculate what character we are hovering and then figure out character from each line start 🤔

amadeus: I was also looking through the shiki apis, and noticed there's an option you can pass to `codeToHast`: `mergeWhitespaces: 'never'`. If you set this prop when also doing the span transforms, you'll never need to do the whitespace processing

packages/diffs/src/managers/InteractionManager.ts
```diff
@@ -289,18 +348,39 @@ export class InteractionManager<TMode extends InteractionManagerMode> {
...
+          lineNumber: this.hoveredToken.lineNumber,
+          side: this.hoveredToken.annotationSide,
+          token: this.hoveredToken.token,
+        } as GetHoveredTokenResult<TMode>;
+      }
+      if (this.mode === 'file' && this.hoveredToken.type === 'token') {
+        return {
+          lineNumber: this.hoveredToken.lineNumber,
+          token: this.hoveredToken.token,
+        } as GetHoveredTokenResult<TMode>;
```
amadeus: What's your thinking on the need a nested object structure here? 

What if we made it so TokenInfo included lineNumber?

amadeus (commented):
Alright, sorry for the delays on this. 

First off, thanks a bunch for putting in work on this!

I have a bunch of notes here so I hope you'll bear with me. Some core worries for me are that always applying these transforms will balloon dom size, so it's very important that we only apply the transform if token events are setup.

This will mean you'd need to extend `RenderDiffOptions` and `RenderFileOptions` to include an option for this. It'll also mean you'll need to ensure it's properly hooked up for the WorkerPool (it should be treated the same way the other render options are, on or off for all Workers unless you set new `setRenderOptions`.

There's some other concerns around actual token value which can get split up which I left notes for in my comment that I think would be useful to get context on how these APIs are used.

packages/diffs/src/utils/processToken.ts
ShpetimA: You are right that we don't need data-token I had it removed initially but thought it would be good as a css selector aswell. But didn't think about ballooning the dom size. Prob the end aswell could be removed so that should reduce dom size by quite a bit. Ill check the transform value in shiki.

ShpetimA: > Additionally, something else I've just discovered that I'm not really sure how to solve -- our line diff type checker will split up tokens in a way that seems to make the text value of `data-token` and/or `textContent` of the span invalid (for the token itself). See this video showing how stuff can be broken up:
> 
>  CleanShot.2026-03-29.at.00.13.26.mp4 
> So my question here, for the tools that integrate these APIs, is the full value of the token important, or is it more important to simply detect character position within the document, to let the LSP do it's thing?
> 
> If it's just the character position from a token (even if it's split up, I'm actually wondering if there's a more lightweight way we could do this with simply some basic metrics where each line gets the character start of that line, the pointer/mouse movement events we can simply calculate what character we are hovering and then figure out character from each line start 🤔

For this one I'll look into it a bit more, first ill apply the quick wins we can get from reducing dom sizes. Since it seems like a bigger api change.

packages/diffs/src/react/utils/useUnresolvedFileInstance.ts
```diff
@@ -171,7 +171,14 @@ export function useUnresolvedFileInstance<LAnnotation>({
...
 
-  return { ref, getHoveredLine, fileDiff, actions, markerRows, getInstance };
+  return {
+    ref,
+    getHoveredLine,
+    fileDiff,
+    actions,
+    markerRows,
+    getInstance,
+  };
```
amadeus: Mostly a nit, but i'd prefer we don't add noise to git blame where it's not necessary, so we should turn this into 1 line

packages/diffs/src/utils/createTransformerWithState.ts
```diff
@@ -20,6 +20,9 @@ export function createTransformerWithState(
   const state: SharedRenderState = { lineInfo: [] };
   const transformers: ShikiTransformer[] = [
     {
+      preprocess(_code, options) {
+        options.mergeWhitespaces = 'never';
+      },
```
amadeus: Oh clever to do this with preprocess.

We'll still only want to apply it if we are doing `onToken` stuff, so i think something like:

```tsx
export function createTransformerWithState(
  useTokenTransformer: boolean,
  useCSSClasses = false
): CreateTransformerWithStateReturn {
  const state: SharedRenderState = { lineInfo: [] };
  const transformers: ShikiTransformer[] = [
    {
      line(node) {
        // Remove the default class
        delete node.properties.class;
        return node;
      },
      pre(pre) {
        const code = findCodeElement(pre);
        const children: ElementContent[] = [];
        if (code != null) {
          let index = 1;
          for (const node of code.children) {
            if (node.type !== 'element') continue;
            children.push(processLine(node, index, state));
            index++;
          }
          code.children = children;
        }
        return pre;
      },
      ...(useTokenTransformer
        ? {
            preprocess(_code, options) {
              options.mergeWhitespaces = 'never';
            },
            span(hast, _line, _col, _lineElement, token) {
              if (token?.offset != null && token.content != null) {
                hast.properties['data-char-start'] = token.offset;
                return hast;
              }
              return hast;
            },
          }
        : null),
    },
  ];
  if (useCSSClasses) {
    transformers.push(tokenStyleNormalizer, toClass);
  }
  return { state, transformers, toClass };
}
```
to wire up the logic conditionally is the right path

packages/diffs/src/managers/InteractionManager.ts
```diff
@@ -883,6 +1016,22 @@ export class InteractionManager<TMode extends InteractionManagerMode> {
...
+    this.hoveredToken.tokenElement.removeAttribute('data-token-hovered');
+    this.hoveredToken = undefined;
+  }
+
+  private setHoveredToken(hoveredToken: TokenEventProps<TMode>) {
+    if (this.hoveredToken != null) {
+      this.clearHoveredToken();
+    }
+    this.hoveredToken = hoveredToken;
+    this.hoveredToken.tokenElement.setAttribute('data-token-hovered', '');
```
amadeus: I think for now we should avoid doing any dom manipulation. If users want to set that up they could with the existing enter/leave events.

packages/diffs/src/utils/renderFileWithHighlighter.ts
```diff
@@ -44,7 +48,10 @@ export function renderFileWithHighlighter(
     totalLines = Infinity;
   }
   const isWindowedHighlight = startingLine > 0 || totalLines < Infinity;
-  const { state, transformers } = createTransformerWithState();
+  const { state, transformers } = createTransformerWithState(
+    false,
+    useTokenTransformer
```
amadeus: Kind of a nit, but lets put `useTokenTransformer` as the first boolean argument for this function because `useCSSClasses` is sort of deprecated/ghost functionality I'm not sure we'll bring back or if we do, it will be a project that we'll need to re-work anyways.

packages/diffs/src/utils/createTransformerWithState.ts
```diff
@@ -32,13 +40,43 @@ export function createTransformerWithState(
...
       },
+      ...(useTokenTransformer
+        ? {
+            tokens(lines) {
+              for (const line of lines) {
+                let col = 0;
+                for (const token of line) {
+                  const tokenWithOriginalRange =
+                    token as TokenWithOriginalRange;
+                  tokenWithOriginalRange[ORIGINAL_TOKEN_COL_START] ??= col;
```
amadeus: I think there's a bug with the current implementation -- where every line with `data-token-col-start` starts at 0. Is this intended? or should start be based on global file char level?

Reason I ask is because TokenInfo only has start/end but no line number, so therefore the data is incomplete.

If the position is based on the character globally, i feel like we can probably use some built in values already from the tokens and may not need to track new values every time?

Sorry if I'm being dumb right now, just trying to think through the whole picture -- what should the API look like for people to consume

packages/diffs/src/utils/createTransformerWithState.ts
```diff
@@ -49,6 +87,134 @@ export function createTransformerWithState(
...
+  container.children = wrappedChildren;
+  return containerTokenState;
+}
+
+function getTokenColStart(node: HASTElement): number | undefined {
+  const value = node.properties['data-token-col-start'];
+  if (typeof value === 'number') {
+    return value;
+  }
+  if (typeof value === 'string') {
```
amadeus: Assuming we control all of this, i wonder if we should `throw` an error if value is a string, just to make sure we don't regress something later on?

packages/diffs/src/utils/createTransformerWithState.ts
ShpetimA: For this I went throught the lsp specification and there for the api it was linenumber plus token offset based kn line so thats why i swapped it out was intentional. Also on the token evemts pretty sure we have line info aswell need to recheck there

ShpetimA:
> Also as a general thing, i think the vibe coded thing to just create wrapper tokens seems to generally work pretty well, and it didn't appear to really cost that much in practice too, so that was dope.

sometimes the robots do cool stuff 🫡

packages/diffs/src/utils/createTransformerWithState.ts
amadeus: Oki yeah that makes sense. Then maybe that change that I made was not a good one, thanks for the context there

amadeus:
> I don't mind either way you can also do commits to clean up stuff. https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/ this is the lsp specification I was trying to follow if you read here you will see token position is line based.

This is very helpful, thanks! I'll rework the patch a bit and post another one, definitely helpful to have this as context! 🙏

packages/diffs/src/managers/InteractionManager.ts
ShpetimA: Removed this whole function aswell as the type It was leftover api from getHoveredToken I had as a function initially but we removed it. The main type that the user will see is
```js 
export interface OnTokenClickProps extends TokenEventBaseProps {
  event: PointerEvent;
}

export interface TokenEventBaseProps extends Omit<LineEventBaseProps, 'type'> {
  type: 'token';
  token: TokenInfo;
  tokenElement: HTMLElement;
}
// This is all the info we expose on token actions, similar to what pierre exposes for line level already with the added tokenEl and tokenInformation.
interface ResolvedTokenTarget<TMode extends InteractionManagerMode> {
  kind: 'token';
  lineType: LineTypes;
  lineElement: HTMLElement;
  lineNumber: number;
  numberColumn: boolean;
  numberElement: HTMLElement;
  side: TMode extends 'diff' ? AnnotationSide : undefined;
  splitLineIndex: number | undefined;
  tokenElement: HTMLElement;
  token: TokenInfo;
}
```

packages/diffs/src/managers/InteractionManager.ts
```diff
@@ -883,6 +975,21 @@ export class InteractionManager<TMode extends InteractionManagerMode> {
...
+    }
+    this.hoveredToken.tokenElement.removeAttribute('data-token-hovered');
+    this.hoveredToken = undefined;
+  }
+
+  private setHoveredToken(hoveredToken: OnTokenEventProps<TMode>) {
+    if (this.hoveredToken != null) {
+      this.clearHoveredToken();
+    }
+    this.hoveredToken = hoveredToken;
```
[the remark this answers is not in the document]
amadeus: This was an intentional decision

packages/diffs/src/utils/createTransformerWithState.ts
```diff
@@ -3,18 +3,25 @@ import {
...
 }
 
+type TokenWithLineChar = ThemedToken & {
+  __lineChar?: number;
+};
+
 export function createTransformerWithState(
+  useTokenTransformer = false,
   useCSSClasses = false
 ): CreateTransformerWithStateReturn {
```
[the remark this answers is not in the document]
amadeus: While this function is technically a public API, it was mostly just used internally and shouldn't be a major issue

apps/demo/src/main.ts
```diff
@@ -42,6 +42,8 @@ import {
   renderDiffAnnotation,
 } from './utils/renderAnnotation';
 
+// FAKE_DIFF_LINE_ANNOTATIONS.length = 0;
+// FAKE_LINE_ANNOTATIONS.length = 0;
```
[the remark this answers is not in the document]
amadeus: Not applicable, this is my personal testing playground with specific stuff commented out for easy access

packages/diffs/src/managers/InteractionManager.ts
```diff
@@ -165,6 +209,7 @@ interface HandlePointerEventProps {
 
 export class InteractionManager<TMode extends InteractionManagerMode> {
   private hoveredLine: EventBaseProps<TMode> | undefined;
+  private hoveredToken: OnTokenEventProps<TMode> | undefined;
   private pre: HTMLPreElement | undefined;
```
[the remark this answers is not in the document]
amadeus: PR description is somewhat out of date at this point i believe.

ShpetimA:
Updated the token info logic to include the new offset info feel free to check it out whenever you have time

packages/diffs/src/utils/renderFileWithHighlighter.ts
```diff
@@ -106,6 +108,14 @@ export function renderFileWithHighlighter(
   return { code, themeStyles, baseThemeType };
 }
 
+function getCharOffsetBase(lines: string[], startingLine: number): number {
+  let charOffsetBase = 0;
+  for (let index = 0; index < startingLine; index++) {
+    charOffsetBase += lines[index]?.length ?? 0;
```
amadeus: Oh man, not sure if i like this. 

Also because it exposes that any diff that `isPartial` will have a char position that's not relatable to the file (and when i eventually add partial diff hydration, we'll see these values changing on hydration)

So with that said, I think the line variant might be the best way since it works everywhere and it'll be the stable

ShpetimA: If you see this as breaking its fine lets just revert the commit and go with the line based approach maybe we rename the values still to lineCharStart etc if in the future we add the offset logic. I was kind of iffy on this part aswell when i found out about this api while doing the change

amadeus: Yeah lets revert for now. Leaving the variables to lineCharStart is probably fine tho, but no strong feelings there.

amadeus: Not your fault tho, my bad for not fully thinking it through with partial/windowing stuff

ShpetimA: Not an issue we tried it atleast, I reverted will make a commit with the lineStart naming since its more explicit

ShpetimA: Pushed a commit with the type names change, my brain is a little fried been coding all day and its late for me, double check for me if you can if we are all good

PR #496, opened 2026-04-11
Hunk Separators: add hidden expand all button

amadeus:
A quick pass at adding a hidden Expand All button to hunk separators for more flexibility when using custom css.

Also made the `unmodified lines` clickable as well to expand in both directions.

- [x] Need to add an underline css style to unmodified lines to help explain clickability


amadeus:
@codex

PR #513, opened 2026-04-17
Updated header slot

mdo:
- Adds icon buttons for new file and folder actions
- Customizes the project name with another icon
- Updates some dropdown styles, improves context menu example
- Modifies trees-dev card styles


SlexAxton (approved):
I think this is good for now, but it's possible we'll want to provide this as part of the trees lib, like inject a slot OR get our default header. doesn't block this going in tho.

mdo:
> I think this is good for now, but it's possible we'll want to provide this as part of the trees lib, like inject a slot OR get our default header. doesn't block this going in tho.

Happy to follow up to make that happen, adding to the list.

PR #516, opened 2026-04-17
Fix scroll appearing on split view with annotations

jonatanklosko:
When using split-view and annotations, the annotation content width is overestimated by 1px leading to unnecessary scroll.

#### Before

(Note the horizontal scroll at the bottom)

image

#### After

image

### DIFF
--- a/packages/diffs/src/managers/ResizeManager.ts
+++ b/packages/diffs/src/managers/ResizeManager.ts
@@ -175,7 +175,7 @@ export class ResizeManager {
 
   private handleResizeObserver = (entries: ResizeObserverEntry[]) => {
     for (const entry of entries) {
-      const { target, borderBoxSize } = entry;
+      const { target, borderBoxSize, contentBoxSize } = entry;
       if (!(target instanceof HTMLElement)) {
         console.error(
           'FileDiff.handleResizeObserver: Invalid element for ResizeObserver',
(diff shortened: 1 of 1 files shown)


packages/diffs/src/managers/ResizeManager.ts
```diff
@@ -211,14 +210,14 @@ export class ResizeManager {
...
-        column.childHeight = specs.blockSize;
+        column.childHeight = borderBoxSize[0].blockSize;
         const newHeight = Math.max(
           item.column1.childHeight,
           item.column2.childHeight
         );
         this.applyNewHeight(item, newHeight);
       } else if (item.type === 'code') {
-        const update: CodeColumnUpdate = [target, specs.inlineSize];
+        const update: CodeColumnUpdate = [target, contentBoxSize[0].inlineSize];
```
jonatanklosko, line 220: This is only relevant in split view, where each `data-code` has 1px border:

https://github.com/pierrecomputer/pierre/blob/a5df5aacf02307c9b641c3c0967587f88bac4eec/packages/diffs/src/style.css#L472-L483

That border would be included in the measured size here. Later we compute the annotation content width as `inlineSize - item.numberWidth`:

https://github.com/pierrecomputer/pierre/blob/a5df5aacf02307c9b641c3c0967587f88bac4eec/packages/diffs/src/managers/ResizeManager.ts#L238-L247

We want to exclude the border from `inlineSize`, which we effectively do by using contentBoxSize here (instead of subtracting something hardcoded).

amadeus:
good catch, nice clean fix, thanks!

PR #524, opened 2026-04-18
Fat/statuses

fat:
Add a bunch of other git statuses.

image

### Slop description

This PR expands git status support in @pierre/trees beyond added, modified, and deleted to also support untracked, renamed, copied, and ignored. It centralizes status presentation across the legacy and path-store renderers, adds styling tokens/selectors for the new states, and updates docs/demo data to show the fuller status set.

It also changes ignored behavior so explicit ignored directories can propagate ignored styling to descendants without pre-expanding entire subtrees, which keeps the implementation safe for very large datasets. Ignored rows no longer render a badge glyph, and ignored icons, including folder chevrons, now render at reduced opacity. Coverage was added/updated for legacy tree and path-store git-status behavior, including directory handling and inherited ignored styling.


packages/trees/src/components/Root.tsx
```diff
@@ -801,11 +801,18 @@ export function Root({
...
-          const itemGitStatus = gitStatusMap?.statusById.get(itemId);
+          const ancestors = getAncestors(itemId);
+          const itemGitStatus =
+            gitStatusMap?.statusById.get(itemId) ??
+            (ancestors.some(
+              (ancestorId) =>
+                gitStatusMap?.ignoredDirectoryIds.has(ancestorId) === true
+            )
+              ? 'ignored'
+              : undefined);
```
SlexAxton: You dont have to go through the trouble of removing this code, but this is old trees, so it didn't need updates.

packages/trees/src/components/TreeItem.tsx
```diff
@@ -5,7 +5,8 @@ import { useMemo } from 'preact/hooks';
 
 import type { ItemInstance, TreeInstance } from '../core/types/core';
```
SlexAxton: also old trees

packages/trees/src/features/git-status/feature.ts
```diff
@@ -15,6 +15,8 @@ type GitStatusCache = {
   gitStatusSignature: string;
```
SlexAxton: old trees

packages/trees/src/path-store/file-tree.ts
```diff
@@ -420,6 +420,7 @@ export class PathStoreFileTree
       composition: this.#composition,
       controller: this.#controller,
       directoriesWithGitChanges: this.#gitStatusState?.directoriesWithChanges,
+      ignoredGitDirectories: this.#gitStatusState?.ignoredDirectoryPaths,
```
SlexAxton: we can do a bigger prop naming pass if you want, but i probably wouldnt do

`directories with git changes`

and then immediately do

`ignored git directories` (backwards from the existing)

packages/trees/src/model/gitStatus.ts
```diff
@@ -4,6 +4,7 @@ import { normalizeInputPath } from '../utils/normalizeInputPath';
 
 export interface PathStoreTreesGitStatusState {
   readonly directoriesWithChanges: ReadonlySet<string>;
+  readonly ignoredDirectoryPaths: ReadonlySet<string>;
```
SlexAxton, line 7: same weird naming here, also the name switches again to add Paths

packages/trees/src/render/FileTreeView.tsx
```diff
@@ -288,41 +293,52 @@ function getDragEdgeScrollDelta(clientY: number, scrollRect: DOMRect): number {
...
 
+function getInheritedIgnoredGitStatus(
+  ancestorPaths: readonly string[],
+  ignoredDirectoryPaths: ReadonlySet<string> | undefined
+): GitStatus | null {
+  if (ignoredDirectoryPaths == null || ignoredDirectoryPaths.size === 0) {
+    return null;
+  }
+
+  for (let index = ancestorPaths.length - 1; index >= 0; index -= 1) {
```
SlexAxton, line 334: probably will slow down benchmarks by like 10-20% if i had to guess, but we can test it.

packages/trees/src/utils/gitStatusPresentation.ts
```diff
@@ -0,0 +1,25 @@
+import type { GitStatus } from '../types';
+
+// These labels keep the status slot width stable while still distinguishing
```
SlexAxton: im curious how we're measuring width from this accurately without pretext

packages/trees/src/features/git-status/feature.ts
```diff
@@ -101,6 +105,9 @@ const buildGitStatusCache = (
     for (const id of idCandidates) {
       if (id != null) {
         statusById.set(id, entry.status);
+        if (entry.status === 'ignored' && entry.path.endsWith('/')) {
+          ignoredDirectoryIds.add(id);
```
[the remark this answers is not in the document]
SlexAxton: real

packages/trees/src/model/gitStatus.ts
```diff
@@ -55,6 +57,9 @@ export function resolvePathStoreGitStatusState(
       normalizedPath.isDirectory
     );
     statusByPath.set(canonicalPath, entry.status);
+    if (entry.status === 'ignored' && normalizedPath.isDirectory) {
+      ignoredDirectoryPaths.add(canonicalPath);
```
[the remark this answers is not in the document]
SlexAxton: also real

packages/trees/src/utils/gitStatusPresentation.ts
fat: i think it's assuming monospace incorrectly

SlexAxton: probably need to think of a different plan, but also idk if we need this functionality

packages/trees/src/render/FileTreeView.tsx
fat:
```
  What I did:

  - getInheritedIgnoredGitStatus(...) now takes a Map<string, boolean> cache and memoizes whether a given ancestor path is inside an ignored subtree. See trees/src/path-store/
    view.tsx:324.
  - PathStoreTreesView now keeps that cache in a useMemo(...) keyed by ignoredGitDirectories, so it persists across renders while the ignored-directory set is unchanged. See trees/src/
    path-store/view.tsx:1086.

  Why this is better:

  - Before: each row could do a backward O(depth) scan every render.
  - Now: once a shared ancestor chain has been resolved, siblings and subsequent renders hit the cache.
  - In practice, large visible groups under the same subtree stop re-paying the full ancestor walk.

  So the hot path becomes closer to:

  - first encounter of a chain: O(depth)
  - later rows sharing that chain: near O(1)

  That’s a better fit for virtualization benchmarks, especially when many visible rows live under the same few ancestor paths.
```

packages/trees/src/model/gitStatus.ts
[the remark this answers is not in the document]
fat: fixed

packages/trees/src/utils/gitStatusPresentation.ts
fat:
```
We’re not measuring it accurately from the text at all.

  The real width stability comes from the fixed status slot width in trees/src/style.css:902:

  [data-item-section='status'] {
    flex: 0 0 auto;
    width: 12px;
    text-align: center;
    align-items: center;
  }

  So the comment in trees/src/utils/gitStatusPresentation.ts:3 is a bit misleading. The single-character labels help because they fit comfortably inside that 12px slot, but they are not
  what makes the width stable. The width is stable because the slot is hard-coded.t
```

i'm just going to remove this comment

PR #526, opened 2026-04-18
Fix infinite loop on plain-text diffs

dimfeld:
### Description

Rendering a diff with language "text" or "ansi" would cause an infinite loop, because those languages are skipped in `getSharedHighlighter` and so never become attached.

This change updates `areLanguagesAttached` to always return true for those languages, to match the corresponding special treatment of those languages in other places.

### Motivation & Context

Infinite loop when rendering "text" diffs. I specifically saw this on a `.gitignore` diff in my project using the Vanilla JS API. This diff had no attached language but the autodetection chose "text" as the language.

Excerpt of the stack trace:

```
  rerender (FileDiff.js:287)
  (anonymous) (FileDiff.js:76)
  onHighlightSuccess (DiffHunksRenderer.js:275)
  (anonymous) (DiffHunksRenderer.js:224)
  Promise.then
  renderDiff (DiffHunksRenderer.js:222)
  render (FileDiff.js:353)
  rerender (FileDiff.js:287)
  (anonymous) (FileDiff.js:76)
  onHighlightSuccess (DiffHunksRenderer.js:275)
  (anonymous) (DiffHunksRenderer.js:224)
  Promise.then
  renderDiff (DiffHunksRenderer.js:222)
  render (FileDiff.js:353)
  rerender (FileDiff.js:287)
```

### Type of changes

      being opened
      that this PR is being opened
- [ ] Documentation update

### Checklist

- [x] I have read the
- [N/A] I have updated the documentation accordingly (if applicable)
- [x] All new and existing tests pass (`bun run diffs:test`)  -- Used `cd packages/diffs && bun run test` since diffs:test command doesn't seem to exist.

Fed the above stack trace into Codex, which diagnosed the error and generated the fix and tests.

I manually reviewed all changes and verified that it fixes the problem in my application.

### DIFF
--- a/packages/diffs/src/highlighter/languages/areLanguagesAttached.ts
+++ b/packages/diffs/src/highlighter/languages/areLanguagesAttached.ts
@@ -5,6 +5,9 @@ export function areLanguagesAttached(
   languages: SupportedLanguages | SupportedLanguages[]
 ): boolean {
   for (const language of Array.isArray(languages) ? languages : [languages]) {
+    if (language === 'text' || language === 'ansi') {
+      continue;
+    }
     if (!AttachedLanguages.has(language)) {
       return false;
     }
--- a/packages/diffs/test/areLanguagesAttached.test.ts
+++ b/packages/diffs/test/areLanguagesAttached.test.ts
@@ -1,0 +1,22 @@
+import { describe, expect, test } from 'bun:test';
+
+import { areLanguagesAttached } from '../src/highlighter/languages/areLanguagesAttached';
+import { AttachedLanguages } from '../src/highlighter/languages/constants';
+
+describe('areLanguagesAttached', () => {
+  test('treats builtin text-like languages as already attached', () => {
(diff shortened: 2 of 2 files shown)


wladimiiir:
I am experiencing the same issue: https://github.com/pierrecomputer/pierre/issues/534 that is being fixed by this PR.

Alternative fix might be by adding the 'text' and 'ansi' to the list of `AttachedLanguages` inside `packages/diffs/src/highlighter/shared_highlighter.ts`.

```
  // 'text' and 'ansi' are loaded during highlighter creation but skipped in
  // the language loop below. Track them as attached so areLanguagesAttached()
  // returns the correct value.
  AttachedLanguages.add('text');
  AttachedLanguages.add('ansi');
  ```

amadeus:
Great catch, and thanks for the fix, will merge when everything passes! 🙏 

Also sorry, totally missed this PR the other day when you opened it.

PR #527, opened 2026-04-18
New homepage with path store and new React wrapper

mdo:


packages/trees/package.json
```diff
@@ -16,6 +16,14 @@
...
       "import": "./dist/index.js"
     },
+    "./path-store": {
+      "types": "./dist/path-store/index.d.ts",
+      "import": "./dist/path-store/index.js"
+    },
+    "./path-store/react": {
+      "types": "./dist/path-store/react/index.d.ts",
+      "import": "./dist/path-store/react/index.js"
+    },
```
SlexAxton, line 26: i think these are dead now

PR #535, opened 2026-04-19
Trees Custom Scrollbars

amadeus:
This is a quick initial pass at adding a custom scrollbar implementation similar to diffs, although making it a bit more configurable with a css override. Additionally there was a quick pass at inline padding for the scroll view that also helps apply to the search input.

CleanShot 2026-04-19 at 15 38 04

Standard css scrollbar properties must be only selectively enabled, or else they opt the browser into a specific render path that forces default scrollbars and ignores custom styles. In practice this is basically just Mozilla specific gating, as the non-standard APIs will _just work_ in even ancient versions of Chrome and Safari.

This PR also includes the ability to `measure` the actual gutter size to ensure inline padding accomodates this.

### DIFF
--- a/packages/trees/src/components/web-components.ts
+++ b/packages/trees/src/components/web-components.ts
@@ -1,6 +1,8 @@
 import { FILE_TREE_STYLE_ATTRIBUTE, FILE_TREE_TAG_NAME } from '../constants';
 import rawStyles from '../style.css';
 import { wrapCoreCSS } from '../utils/cssWrappers';
+import { ensureMeasuredScrollbarGutter } from '../utils/scrollbarGutter';
+
 let sheet: CSSStyleSheet | undefined;
 
 export function ensureFileTreeStyles(shadowRoot: ShadowRoot): void {
@@ -43,6 +45,17 @@ export function ensureFileTreeStyles(shadowRoot: ShadowRoot): void {
   }
 }
 
+// Keeps every tree host on the same setup path so runtime mounting, hydration,
+// and autonomous custom-element upgrades share one shadow-root preparation step.
+export function prepareFileTreeShadowRoot(
+  host: HTMLElement,
+  shadowRoot: ShadowRoot
+): void {
+  adoptDeclarativeShadowDom(host, shadowRoot);
+  ensureFileTreeStyles(shadowRoot);
+  ensureMeasuredScrollbarGutter(host, shadowRoot);
+}
+
 export function adoptDeclarativeShadowDom(
   host: HTMLElement,
   shadowRoot: ShadowRoot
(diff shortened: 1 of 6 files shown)


packages/trees/src/components/web-components.ts
```diff
@@ -43,6 +44,17 @@ export function ensureFileTreeStyles(shadowRoot: ShadowRoot): void {
...
+// Keeps every tree host on the same setup path so runtime mounting, hydration,
+// and autonomous custom-element upgrades share one shadow-root preparation step.
+export function prepareFileTreeShadowRoot(
+  host: HTMLElement,
+  shadowRoot: ShadowRoot
+): void {
+  adoptDeclarativeShadowDom(host, shadowRoot);
+  ensureFileTreeStyles(shadowRoot);
+  ensureMeasuredScrollbarGutter(host, shadowRoot);
+}
```
amadeus, line 57: We used to just perform 

```tsx
  adoptDeclarativeShadowDom(host, shadowRoot);
  ensureFileTreeStyles(shadowRoot);
```

in a bunch of places to prepare the shadow root, since we added a third method to the ring, they are all converged into a function.

amadeus:
@codex

packages/trees/src/style.css
```diff
@@ -512,6 +514,9 @@
         var(--trees-focus-ring-width)
     );
 
+    --trees-scrollbar-gutter: var(--trees-scrollbar-gutter-override, 6px);
+    --trees-padding-inline: var(--trees-padding-inline-override, 16px);
```
amadeus, line 518: Aesthetically as a default, I'm a bit worried that 16px is too much fwiw (12px feels better)

packages/trees/src/utils/scrollbarGutter.ts
```diff
@@ -0,0 +1,44 @@
...
+  wrapper.setAttribute(FILE_TREE_SCROLLBAR_MEASURE_ATTRIBUTE, 'true');
+  const child = document.createElement('div');
+  child.style.position = 'relative';
+  child.style.height = '200%';
+  wrapper.appendChild(child);
+  shadowRoot.appendChild(wrapper);
+
+  const measuredGutter = Math.max(wrapper.offsetWidth - wrapper.clientWidth, 0);
+  wrapper.remove();
+  measuredGutterCache.set(shadowRoot, measuredGutter);
```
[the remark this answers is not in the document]
amadeus: Implemented a partial fix for this: with an `isConnected` check.

To do a proper check would be more work that I'm not sure is worth it at the moment (would involve creating a resize observer which really didn't feel worth it right now at least, until we get more signal from people on if they experience this issue or not)

SlexAxton (approved):

packages/trees/src/style.css
SlexAxton: cc @mdo for your consideration at some point

SlexAxton: also might need to get pulled into the density stuff

PR #579, opened 2026-04-22
[trees] Optimize sticky scroll stuff maybe

SlexAxton:
Avoid full visible-row materialization in sticky scroll layout by computing sticky candidates and mounted window rows from compact visible-row context. Adds PathStore row-context metadata, updates FileTreeController/FileTreeView layout flow, and adds sticky-scroll validation benchmarks plus focused context tests.

Experiments: #3, #4, #11, #12, #13, #14, #17
Metric: sticky_full_layout_p50_ms 413.06ms → ~0.09ms; sticky_cold_view_update_window_p50_ms 592.86ms → ~0.13-0.18ms; 14-dispatch model/controller scroll sequence ~0.74ms total. Explicit full visible-row materialization remains hundreds of ms, so this avoids the full-tree path rather than making it disappear.


SlexAxton:
also noticing this floating context menu here, after focusing and blurring

image

PR #583, opened 2026-04-22
[trees] Avoid full projection rebuild on expand-collapse

SlexAxton:
Adds a targeted PathStore visible-index lookup backed by node lookup, ancestor visibility checks, subtree counts, and child chunk sums. FileTreeController uses that lookup for focus resolution instead of forcing visibleIndexByPath construction, and ordinary non-search expand/collapse events rebuild only the limited projection needed for the viewport.

Experiments: #2, #3
Metric: controller_toggle_p50_ms 639.68ms → 0.134667ms (-99.98%)

### DIFF
--- a/packages/path-store/src/child-index.ts
+++ b/packages/path-store/src/child-index.ts
@@ -229,6 +229,47 @@ export function selectChildIndexByVisibleIndex(
   );
 }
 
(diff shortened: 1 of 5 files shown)


SlexAxton:
@codex review

PR #594, opened 2026-04-22
WS TTY Fix Round 2

amadeus:
Soo, my previous fix at #570 only worked initially if you hadn't loaded any routes.

Initially the AI extended the timeout to wait for child processes, but then I told it to go back to first principles and figure out how to better solve this. 

Anyways, it came up with this, it does appear to work even after a route loads and I hope it's much more stable than before.


SlexAxton (approved):
🤷🏻‍♂️

PR #595, opened 2026-04-22
mb fix search bugs with custom sorts (or something)

amadeus:
Potential fix for search when using whatever atrocity my llm created to have a custom sort in the /gh route.

### DIFF
--- a/packages/path-store/src/builder.ts
+++ b/packages/path-store/src/builder.ts
@@ -169,6 +169,8 @@ export function prepareInput(
     {
       paths: preparedPaths.map((entry) => entry.path),
       preparedPaths,
+      preparedPathsValidateOrder:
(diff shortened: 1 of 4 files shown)


packages/path-store/src/builder.ts
```diff
@@ -224,6 +226,13 @@ export function getPreparedInputEntries(
   return preparedPaths;
 }
 
+export function getPreparedInputPreparedPathsValidateOrder(
+  preparedInput: import('./public-types').PathStorePreparedInput
```
SlexAxton, line 230: why

amadeus: haha, sorry i never even bothered to look at the code, anyways, your feedback in the thread appeared to fix the issue

PR #600, opened 2026-04-23
Split the docs

mdo:
- `apps/docs/` is now `apps/docs-diffs/` and `apps/docs-trees`
- Shared bits are now in `packages/docs-shared`

Still working through the long tail details.


mdo:
Follow-up for the dev memory blowouts I've been seeing—`docs-shared` now builds to `dist/` via tsdown. This let me start my dev server without it blowing up the memory to 100GB.

SlexAxton:
ehhhh fine, could just be two directories in one app

PR #601, opened 2026-04-23
Editor

ije:
## Architecture

The editor is a plugable system for the `File` component. It provides:

- Text Editing
- Selection management
- History (undo, redo)
- Search
- Quick Edit
- Automatic Indentation

The editor adds the `contenteditable` property to the code line element, that supports native web selection and editing features on mobile phone.

```js
const contentEl = fileContainer.shadowRoot.querySelector("[data-content]")
contentEl.contentEditable = true
contentEl.addEventListener("beforeInput", handleInput)
```

Selections are created via the native web selection API, multiple cursors has been supported.

```js
document.addEventListener('selectionchange', () => {
  const selectionRaw = document.getSelection();
  const composedRanges = selectionRaw?.getComposedRanges({
    shadowRoots: [fileContainerShadowRoot],
  });
  const selection = convertSelection(composedRanges);
  if (selection) {
    // Text in the editor has been selected.
    // You can now get the `start.line`, `start.column`, `end.line` and `end.column` from the selection
  }
})
```

## API

Vanilla JS:

```js
import { VirtualizedFile } from "@pierre/diffs"
import { Editor } from "@pierre/diffs/editor"

// render a file
const fileInstance = new VirtualizedFile({ ... });
fileInstance.render({ ... });

// make the file editable
const editor = new Editor({
  onChange: (file, lineAnnotations) => console.log("change", file.name, file.contents)
})
const dispose  = editor.edit(fileInstance)
dispose() // or call `editor.cleanUp()`
```

lazy import (recommended):

```js
import { VirtualizedFile } from "@pierre/diffs"

// render a file
const fileInstance = new VirtualizedFile({ ... });
fileInstance.render({ ... });

const edit = async (file: File) => {
  const { Editor } = await import("@pierre/diffs/editor")
  const editor = new Editor({
    onChange: (file, lineAnnotations) => console.log("change", file.name, file.contents)
  })
  return editor.edit(file)
}

// click to edit (lazy importing the editor.js)
button.addEventListener("click", () => void edit(fileInstance))
```

React:

```tsx
import { Editor } from "@pierre/diffs"
import {
  type FileContents,
  File,
  EditorProvider,
} from '@pierre/diffs/react';

const file: FileContents = {
  name: 'example.ts',
  contents: `function greet(name: string) {
  console.log(\`Hello, \${name}!\`);
}

export { greet };`,
};

const editor = new Editor({
  onChange: (file, lineAnnotations) => console.log("change", file.name, file.contents)
})

export function EditCodeFile() {
  const [editable, setEditable] = useState(true)
  return (
    <EditorProvider editor={editor}>
      <File
        file={file}
        options={{
          theme: { dark: 'pierre-dark', light: 'pierre-light' },
        }}
        editable={editable}
      />
    </EditorProvider>
  );
}
```

## TODOS
- [x] Basic editing
- [x] Selection & cursor 
- [x] Multiple cursors (Pressing ⌘)
- [x] History(undo/redo)
- [x] Search
- [x] Quick Edit 
- [x] Automatic indentation
- [x] Shortcuts
  - 'selectAll' `⌘ + A`
  - 'copy' `⌘ + C`
  - 'cut' `⌘ + X`
  - 'paste' `⌘ + V/Y`
  - 'findNextMatch' `⌘ + D`
  - 'indent' `Tab`
  - 'outdent' `Shift + Tab`
  - 'moveCursorToDocStart' `⌘ + ↑`
  - 'moveCursorToDocEnd' `⌘ + ↓`
  - 'expandSelectionDocStart' `⌘ + Shift + ↑`
  - 'expandSelectionDocEnd' `⌘ + Shift + ↓`
  - 'undo' `⌘ + Z`
  - 'redo' `⌘ + Shift + Z`
- [x] Browser compatibility
  - [x] Chrome
  - [x] Firefox
  - [x] Safari
  - [x] Android Chrome
  - [x] iOS Safari (has input bug)
- [x] ~~LSP~~ _next version_


packages/diffs/src/types.ts
```diff
@@ -707,7 +709,7 @@ export interface ForceFilePlainTextOptions {
   startingLine?: number;
   totalLines?: number;
   // Pre-split lines for caching in windowing scenarios
-  lines?: string[];
+  lineOffsets?: number[];
```
amadeus: Why this change?

packages/diffs/src/react/FileDiff.tsx
```diff
@@ -14,6 +14,7 @@ export interface FileDiffProps<
 > extends DiffBasePropsReact<LAnnotation> {
   fileDiff: FileDiffMetadata;
   disableWorkerPool?: boolean;
+  contentEditable?: boolean;
```
amadeus, line 17: Similar q from File.tsx

packages/diffs/src/components/FileDiff.ts
```diff
@@ -957,6 +987,105 @@ export class FileDiff<LAnnotation = undefined> {
...
+    const fileDiff = this.fileDiff;
+    if (fileDiff != null && !fileDiff.isPartial) {
+      const { name, lang, cacheKey } = fileDiff;
+      const file = {
+        name,
+        lang,
+        cacheKey,
+      } as FileContents;
+      Object.defineProperty(file, 'contents', {
+        get: () => fileDiff.deletionLines.join(''),
```
amadeus, line 1062: I wonder if this is going to cause differences with partial diffs where they render differently in edit mode, since if the file is partial we actually render each hunk separately, and i assume in the context of the editor, we render it all as 1 file?

packages/diffs/src/components/VirtualizedFile.ts
```diff
@@ -287,8 +289,7 @@ export class VirtualizedFile<
     }
 
     const { disableFileHeader = false, collapsed = false } = this.options;
-    const lines = this.getOrCreateLineCache(this.file);
-    const lastLineIndex = getLastVisibleLineIndex(lines);
+    const lastLineIndex = this.getOrCreateLineOffSets(this.file).at(-1) ?? -1;
```
amadeus: Pretty sure this codex bug/find is valid

packages/diffs/src/components/VirtualizedFile.ts
```diff
@@ -477,18 +477,15 @@ export class VirtualizedFile<
...
-        lines,
-        callback: ({ lineIndex }) => {
-          this.addLayoutCheckpoint(lineIndex, this.height);
-          this.height += this.getLineHeight(lineIndex, false);
-        },
-      });
+      for (let lineIndex = 0; lineIndex < lineCount; lineIndex++) {
+        this.addLayoutCheckpoint(lineIndex, this.height);
+        this.height += this.getLineHeight(lineIndex, false);
+      }
```
amadeus, line 485: what was the need for this change?

packages/diffs/src/components/VirtualizedFile.ts
```diff
@@ -909,18 +937,3 @@ export class VirtualizedFile<
...
-    lastLine === '' ||
-    lastLine === '\n' ||
-    lastLine === '\r\n' ||
-    lastLine === '\r'
-  ) {
-    return lines.length - 2;
-  }
-
-  return lines.length - 1;
-}
```
amadeus, line 926: Pretty sure it's not safe to remove this, as it was accommodating for some important end of file behavior

packages/diffs/src/types.ts
```diff
@@ -740,6 +742,7 @@ export interface RenderedFileASTCache {
   options: RenderFileOptions;
   result: ThemedFileResult | undefined;
   renderRange: RenderRange | undefined;
+  isDirty?: boolean;
```
amadeus, line 745: Can you explain a bit about the architecture for `isDirty`? What does it serve, how is it used?

packages/diffs/src/components/VirtualizedFile.ts
```diff
@@ -813,51 +842,50 @@ export class VirtualizedFile<
...
-            }
-            overflowCounter--;
-          }
-        }
+    const startingLineIndex = checkpoint?.lineIndex ?? 0;
+    for (
+      let lineIndex = startingLineIndex;
+      lineIndex < lineCount;
+      lineIndex++
+    ) {
```
amadeus, line 850: How come this iteration change?

packages/diffs/src/components/VirtualizedFile.ts
```diff
@@ -1,6 +1,8 @@
 import { DEFAULT_VIRTUAL_FILE_METRICS } from '../constants';
```
amadeus, line 1: So there's a bunch of changes to VirtualizedFile, but no changes to VirtualizedFileDiff. Is that just work that still needs to happen? or was there something that needed to change about VirtualizedFile that wasn't applicable to VirtualizedFileDiff?

packages/diffs/src/components/VirtualizedFile.ts
```diff
@@ -574,6 +600,10 @@ export class VirtualizedFile<
       this.isSetup = true;
     } else {
       this.top ??= this.getVirtualizedTop();
+      if (didFileChange) {
+        this.getSimpleVirtualizer()?.markDOMDirty();
+        this.resetLayoutCache(true);
```
amadeus, line 605: Not sure this is the right place for this in context of CodeView.

This whole if statement might need to be gated on simple virtualizer?

packages/diffs/src/editor/css.ts
```diff
@@ -0,0 +1,218 @@
+const DEBUG_SELECTION = false;
+
+export const editorCSS: string = /* CSS */ `
```
amadeus, line 3: Is this CSS we can stuff in our diffs style.css?

packages/diffs/src/utils/computeFileOffsets.ts
```diff
@@ -0,0 +1,23 @@
+const LINE_FEED = 10; // \n
+const CARRIAGE_RETURN = 13; // \r
+
+/**
+ * Computes line start offsets for a string.
+ */
+export function computeLineOffsets(contents: string): number[] {
+  const offsets: number[] = [0];
+  for (let i = 0; i < contents.length; i++) {
```
amadeus, line 9: A bit worried about the performance of this, iterating over every character vs before when we were just doing splits on lines

packages/diffs/src/utils/iterateOverFile.ts
```diff
@@ -1,82 +0,0 @@
...
-        lineNumber: lineIndex + 1,
-        content: lines[lineIndex],
-        isLastLine,
-      }) === true ||
-      isLastLine
-    ) {
-      break;
-    }
-  }
-}
```
amadeus, line 82: Probably relates back to some earlier questions about why this was removed?

packages/diffs/src/utils/iterateOverFile.ts
```diff
@@ -1,82 +0,0 @@
...
-    if (
-      lastLine === '' ||
-      lastLine === '\n' ||
-      lastLine === '\r\n' ||
-      lastLine === '\r'
-    ) {
-      return Math.max(0, lines.length - 2);
-    }
-    return lines.length - 1;
-  })();
```
amadeus, line 65: Mostly because there's a lot of deliberate stuff that went into this here

packages/diffs/src/utils/cleanLastNewline.ts
```diff
@@ -1,3 +1,10 @@
 export function cleanLastNewline(contents: string): string {
-  return contents.replace(/\n$|\r\n$/, '');
+  let end = contents.length;
+  if (contents.charCodeAt(end - 1) === /* \n */ 10) {
+    end--;
+    if (contents.charCodeAt(end - 1) === /* \r */ 13) {
+      end--;
+    }
+  }
+  return contents.slice(0, end);
```
amadeus, line 9: How come this needed to change? Was it a performance thing?

packages/diffs/src/react/types.ts
```diff
@@ -46,4 +46,9 @@ export interface FileProps<LAnnotation> {
   style?: CSSProperties;
   prerenderedHTML?: string;
   disableWorkerPool?: boolean;
+  contentEditable?: boolean;
+  onChange?: (
+    file: FileContents,
+    lineAnnotations?: LineAnnotation<LAnnotation>[]
+  ) => void;
```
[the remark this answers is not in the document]
amadeus: I think this is a valid issue?

packages/diffs/src/react/EditorContext.tsx
```diff
@@ -0,0 +1,27 @@
...
+      editor.cleanUp();
+    };
+  }, [editor]);
+  return (
+    <EditorContext.Provider value={editor}>{children}</EditorContext.Provider>
+  );
+}
+
+export function useEditor<LAnnotation>(): Editor<LAnnotation> | undefined {
+  return useContext(EditorContext) as Editor<LAnnotation> | undefined;
```
amadeus, line 26: Is there a way we can fix the types here so we don't need to use `as`?

packages/diffs/src/react/utils/useFileInstance.ts
```diff
@@ -107,12 +115,21 @@ export function useFileInstance<LAnnotation>({
...
     });
+    if (contentEditable && editor !== undefined) {
+      newOptions = {
+        ...newOptions,
+        useTokenTransformer: true,
+        enableGutterUtility: false,
+        enableLineSelection: false,
+        lineHoverHighlight: 'disabled',
+      };
+    }
```
amadeus: Can we fix `mergeFileOptions` above to include this work, and not separate into a separate spread?

amadeus: Also, i think we should probably throw an error if the editor is undefined and contentEditable was set to true?

packages/diffs/src/react/utils/useFileDiffInstance.ts
```diff
@@ -128,6 +142,17 @@ export function useFileDiffInstance<LAnnotation>({
...
+  useIsometricEffect(() => {
+    if (
+      contentEditable &&
+      editor !== undefined &&
+      instanceRef.current != null
+    ) {
+      return editor.edit(instanceRef.current);
+    }
+    return undefined;
+  }, [contentEditable, editor]);
```
amadeus, line 149: Similar note, we should throw if someone is trying to edit with no editor instance?

packages/diffs/src/react/utils/useFileDiffInstance.ts
```diff
@@ -110,12 +114,22 @@ export function useFileDiffInstance<LAnnotation>({
...
+    if (contentEditable && editor !== undefined) {
+      newOptions = {
+        ...newOptions,
+        useTokenTransformer: true,
+        enableGutterUtility: false,
+        enableLineSelection: false,
+        expandUnchanged: true,
+        lineHoverHighlight: 'disabled',
+      };
+    }
```
amadeus: Similar note to file -- can we push this logic into `mergeFileDiffOptions`?

amadeus (commented):
So I've mostly just looked over the changes to existing diffs library. I do have a couple concerns about some potential behavior changes, but there also might be some prior context I am not aware of on how things have changed.

Also chatted a bit with the team, what I will do here shortly is create a new specialized branch for the beta/alpha of editor, and then you can target that branch with this PR, and we can cut beta releases from that

amadeus:
Created the new target branch for this stuff:

https://github.com/pierrecomputer/pierre/pull/749

`beta-1.3`

packages/diffs/src/components/VirtualizedFile.ts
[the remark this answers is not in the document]
amadeus: I believe this is a legit bug, which i have other comments about with regards to the need for the change

packages/diffs/src/react/types.ts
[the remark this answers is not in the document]
ije: removed

packages/diffs/src/components/VirtualizedFile.ts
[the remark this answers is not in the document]
ije: fixed `lastLineIndex`. i removed `getLastVisibleLineIndex` function, because now we need to render empty lines at the end of the file.

packages/diffs/src/types.ts
ije: yeah i think this is slower to initialize and read, but can reduce memory usage. this aligns to the text buffer. however i can change it back it `string[]`.

packages/diffs/src/react/FileDiff.tsx
ije: to make the file component editable, for example:
```tsx
function F() {
  const [editable, setEditable]= (false)
  
  return <EditorProvider editor={editor}>
    <File
      file={file}
      options={{
        theme: { dark: 'pierre-dark', light: 'pierre-light' },
      }}
      contentEditable={editable}
    />
  </EditorProvider>
}
```

packages/diffs/src/components/FileDiff.ts
ije: currently the editor can not edit partial diffs

Screenshot 2026-05-29 at 23 58 40

packages/diffs/src/components/VirtualizedFile.ts
ije: fixed.

packages/diffs/src/components/VirtualizedFile.ts
ije: when editing a file instance, add/delete lines will cause the scroll height change. for file diff instance we just call `rerender`

packages/diffs/src/components/VirtualizedFile.ts
ije: ~~related to `lines` to `lineOffsets`~~ we don't use `iterateOverFile` since the editor could change the line count of the file

packages/diffs/src/components/VirtualizedFile.ts
ije: fixed

packages/diffs/src/components/VirtualizedFile.ts
ije: ~~again, related to `lines` to `lineOffsets` change. maybe i should not touch this part lol~~ we don't use `iterateOverFile` since the editor could change the line count of the file

packages/diffs/src/components/VirtualizedFile.ts
ije: i think this is safe to remove. now we need to render all lines to align with the editor.

packages/diffs/src/editor/css.ts
ije: i prefer to load it standalone, to remain diffs style.css smaller, and can be lazy import.

Image

packages/diffs/src/react/utils/useFileDiffInstance.ts
ije: done

packages/diffs/src/react/utils/useFileDiffInstance.ts
ije: makes sense! updated.

packages/diffs/src/react/utils/useFileInstance.ts
ije: done.

packages/diffs/src/react/EditorContext.tsx
ije: because useContext(EditorContext) returns ` Editor<unknown>`

packages/diffs/src/utils/cleanLastNewline.ts
ije: yep, should has better performance than regexp

packages/diffs/src/utils/computeFileOffsets.ts
ije: yes, lineOffsets is slower than `lines: []string` on init/read. this is come from the pieceTable. for large file, store `lineOffsets` can reduce memory. I can revert the change on `File` component, but the pieceTable still needs this funtion.

packages/diffs/src/types.ts
ije: when editing a file, the editor will update the `code` of the render cache. mark it as `dirty` to purge the render cache in workerpool when call `cleanUp`. this avoid `file.render({...})` gets polluted code lines.

packages/diffs/src/react/FileDiff.tsx
amadeus: Sorry this was poor messaging on my part. This was just relating back to this note: https://github.com/pierrecomputer/pierre/pull/601#discussion_r3321486620

that there was an onChange event, and it looks like you fixed this already

packages/diffs/src/utils/iterateOverFile.ts
ije: this can be replaced by for loop safely, we now don't use it since the editor can change the line count of the file, but the `lineCache`(i reverted) remain to the original file lines.

packages/diffs/src/utils/computeFileOffsets.ts
ije: reverted to `lines`

packages/diffs/src/components/VirtualizedFile.ts
ije: i also added a testing for the change: https://github.com/pierrecomputer/pierre/pull/601/changes#diff-e2202f70437792fe3d5ba0bd16eeda7f0ce4704e4484f06df95087545302df8cR155-R168

ije:
@amadeus `lineOffsets` has been change back to `lines`, but the `iterateOverFile.ts` file is not reverted, it can be replaced with for loop safely. we do not use it now since the editor could change the line count of the file. 

can you please take a look again thanks 💚

packages/diffs/src/components/FileDiff.ts
amadeus: Gotcha, makes sense

packages/diffs/src/components/VirtualizedFile.ts
amadeus: Gotcha, yeah, this is an interesting problem, because i think most editors `hide` that last line but i also understand it's a bit more complicated here. This is probably fine for our beta, and i can do some experiments to see what sorta side-effects it'll have.

packages/diffs/src/components/VirtualizedFile.ts
amadeus: I don't know if i fully understand, couldn't diffs also change scroll height too?

packages/diffs/src/types.ts
amadeus: Perhaps just leave a quick comment on it there for future context?

packages/diffs/src/editor/css.ts
amadeus: hmm, maybe we can do it in a separate stylesheet then? I think it's fine for now, but with more complex CSS stuff I'd love to have it in a more dedicated css file for general edits. (our build system turns these into strings anyways, plus we can make it a shared stylesheet which i think is useful. But not a blocker for this.

packages/diffs/src/utils/computeFileOffsets.ts
amadeus: yeah, not necessarily opposed to this stuff, just trying to be diligent about tradeoffs. It's probably fine to get something merged for beta, and I/we can conduct some tests with stuff to really understand impacts.

packages/diffs/src/react/EditorContext.tsx
amadeus: Yeah i get that, but i'm wondering if we can fix the context itself to take a type we can pass through is all

PR #602, opened 2026-04-23
Fix Misc hydration issues

amadeus:
* Fixed an issue with headers being disabled
* Hardened up the concept of when we should attempt to render a pre tag


amadeus:
@codex

amadeus:
@codex

amadeus:
@codex

PR #603, opened 2026-04-23
Hydration Fix v2

amadeus:
And anotha one...

Basically the heuristic for first render was pretty broken (i even had a comment about it not being safe).

Anyways, with the last round of hydration fixes, we created a new scenario for broken renders on Virtualized instances. This should hopefully fix that.


amadeus:
@codex

PR #604, opened 2026-04-24
Reorganize docs into `(trees)` and `(diffs)`

mdo:
Replacement for #600 without splitting the docs app into two entirely separate apps. More to do still.


mdo:
@codex review

SlexAxton (approved):
Assuming both pages load how you expect i like this setup for now.

PR #605, opened 2026-04-24
[diffs] Add diffs demo profile workload

SlexAxton:
Add the demo/profile harness used to measure large diff rendering through the normal diffs package path. This branch establishes the clean autoresearch baseline without changing library rendering behavior.

Experiments: #3
Metric: post_paint_ready_ms baseline 91.3ms


amadeus:
Is this something we still want to add?

SlexAxton:
I can always generate it again, so up to you on whether it’s useful for you.

PR #606, opened 2026-04-24
[diffs] Optimize diffs virtualized initial render

SlexAxton:
Reduce large diff post-paint time by avoiding unnecessary synchronous layout reads, lazy-creating expensive shadow/style/sprite resources, tightening ResizeManager work, and reducing initial virtualized rendering work while keeping the safer 800px overscan for the active render window.

Experiments: #4, #5, #9, #11, #12, #20, #23, #27, #31, #32, #33, #39, #42, #53, #54, #61, #62, #64, #65, #77, #81, #90, #91, #93, #104, #111, #113
Metric: post_paint_ready_ms 91.3ms → 24.6ms (-73.1%)


amadeus:
There's been a lot of changes in my codeviewer branch around how dom stuff is orchestrated, so while these might be beneficial, it's unclear how well they'd map over to the new work there, we should probs coordinate on running stuff on that branch.

that said, i need to stabilize a few more things with that branch first and then we can look into optimizing.

SlexAxton:
nw

PR #608, opened 2026-04-24
minify diff css a bit

amadeus:
a quick lil thing to minify the css a bit that we bundle. Should theoretically be a safe change


amadeus:
@codex

PR #609, opened 2026-04-27
fix(@pierre/trees): mark `preact` and `preact-render-to-string` as optional peer dependencies

dimaMachina:
### Description

I don’t see the point of installing these dependencies for React-only projects

### Type of changes

      being opened
      that this PR is being opened
- [ ] Documentation update

### Checklist

- [ ] I have read the
- [ ] My code follows the code style of the project (`bun run lint`)
- [ ] My code is formatted properly (`bun run format`)
- [ ] All new and existing tests pass (`bun run diffs:test`)

### DIFF
--- a/packages/trees/package.json
+++ b/packages/trees/package.json
@@ -71,8 +71,6 @@
   },
   "dependencies": {
     "@pierre/path-store": "workspace:*",
-    "preact": "catalog:",
-    "preact-render-to-string": "catalog:"
   },
   "devDependencies": {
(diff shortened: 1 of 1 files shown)


SlexAxton:
preact is used internally regardless of whether you use the vanilla api it the react wrapper. We initially wrote the library to be fully without a rendering library, but our custom system grew just as large as preact is, so it made less sense to keep.

it seems possible that we could swap for react internally when using react, but that’s not currently something I’ve been able to easily untangle without creating two totally distinct implementations.

PR #613, opened 2026-04-27
[trees] recreate buggy gh flow with fixture

SlexAxton:
Trying to recreate a bug


amadeus:
Don't think we need this anymore now that we figured out the issue was related to strict mode and this change did not reproduce it.

PR #616, opened 2026-04-28
[trees] Strict Mode Fix

amadeus:
When a dev environment has react strict mode enabled, the lifecycle for the FileTree component gets messed up. This new variant ensures a stable instance, and a cleanup that's always deferred by a tick.

Also addes a regression test


SlexAxton (approved):
bless

PR #626, opened 2026-05-01
Ensure diffshub Vercel project builds

mdo:
Needed to address the LLMs thing most of all, TBD if I keep the CI build task for this, or if we build the rest of the pages too for verification since this would otherwise miss some things.


amadeus:
dang, look who has the technology

PR #630, opened 2026-05-01
Hook up DiffsHub view options

mdo:
Attempts to hook up toggles for line numbers and background colors.

### DIFF
--- a/apps/docs/app/(diffshub)/(view)/_components/utils.ts
+++ b/apps/docs/app/(diffshub)/(view)/_components/utils.ts
@@ -135,13 +135,8 @@ function createPatchOrderSort(paths: readonly string[]): CodeViewFileTreeSort {
 // Finalizes the stable tree input from a fresh fetch. Callers are expected to
 // populate paths, pathToItemId, and gitStatus in the same pass that builds
 // the viewer items so the tree data structure does not require its own walk
-// over items.
-//
-// Modified files are filtered out so they render as the visual default — no
-// blue tint, no "M" badge. Only added, deleted, and renamed files retain their
-// own status indicators. Filtering here also prevents the folder dot from
-// appearing on directories whose only changed children are modified files;
-// dots on folders with added/renamed/deleted children are suppressed via CSS.
+// over items. Modified-status entries should be excluded from gitStatus before
+// calling here so the tree renders them as the visual default (no tint or badge).
 export function createCodeViewFileTreeSource(
   paths: readonly string[],
   pathToItemId: ReadonlyMap<string, string>,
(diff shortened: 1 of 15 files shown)


packages/diffs/src/components/VirtualizedFileDiff.ts
```diff
@@ -120,19 +120,27 @@ export class VirtualizedFileDiff<
     return this.metrics.lineHeight * multiplier;
   }
 
-  // Override setOptions to clear height cache when diffStyle changes
+  // Override setOptions to clear height cache when layout-affecting options
+  // change, and to force a re-render when visual-only options change (those
+  // don't affect layout, but the DOM attribute must still be updated).
+  // disableLineNumbers is treated as layout-affecting because hiding the number
+  // columns widens the code column, which can shift wrap points in overflow:wrap
+  // mode and invalidate measured row heights.
```
amadeus: I think it might be nicer to not have this comment here, and perhaps litter comments inside around why we are doing these things.

Also, we'll need to add the disable line numbers check to VirtualizedFile as well

packages/diffs/src/components/VirtualizedFileDiff.ts
```diff
@@ -144,6 +152,14 @@ export class VirtualizedFileDiff<
       }
       this.renderRange = undefined;
     }
+    // Visual-only and layout-affecting options both need forceRenderOverride so
+    // FileDiff.render's early-return guard doesn't skip applyPreNodeAttributes.
```
amadeus, line 153: This is a good example of the inline type comments that would be nice to to add around layout changes requiring a clear of cache and stuff

amadeus (commented):
The branch is lookin good! 2 smol notes, and some logic that will need to be parallelized into VirtualizedFile

apps/docs/app/(diffshub)/(view)/_components/utils.ts
```diff
@@ -136,14 +136,20 @@ function createPatchOrderSort(paths: readonly string[]): CodeViewFileTreeSort {
...
+// dots on folders with added/renamed/deleted children are suppressed via CSS.
 export function createCodeViewFileTreeSource(
   paths: readonly string[],
   pathToItemId: ReadonlyMap<string, string>,
   gitStatus: readonly GitStatusEntry[]
 ): CodeViewFileTreeSource {
   const sort = createPatchOrderSort(paths);
   return {
-    gitStatus,
+    gitStatus: gitStatus.filter((e) => e.status !== 'modified'),
```
amadeus: I don't love this happening here -- it could potentailly be a very big operation if there's a lot of files... could we maybe just not push to this array if it's status === 'modified'?

amadeus (commented):
Looking good! The only thing would be the performance optimization for gitStatus.

Also this doesn't add support for indicator style yet it seems, is that something you were planning on doing later? also totally fine to do after too

amadeus (approved):
🙏

Issue #638, opened 2026-05-05
Bug:  ResizeObserver loop completed with undelivered notifications


lexdotdev:
### Prerequisites

- [x] I have [searched](https://github.com/pierrecomputer/pierre/issues?q=is%3Aissue) for duplicate or closed issues
- [x] I have read the [contributing guidelines](https://github.com/pierrecomputer/pierre/blob/main/.github/CONTRIBUTING.md)

### Describe the issue

[ResizeManager.handleResizeObserver](https://github.com/pierrecomputer/pierre/blob/e160b7e093e908e6cef11f82f4c567aea5c68868/packages/diffs/src/managers/ResizeManager.ts#L176) writes layout-affecting CSS variables (`--diffs-column-content-width`, `--diffs-column-number-width`, `--diffs-column-width`, `--diffs-annotation-min-height`) synchronously inside its `ResizeObserver` callback. Those writes can re-resize the observed `<code>` and gutter elements before the callback returns, so the browser fires:

```
ResizeObserver loop completed with undelivered notifications.
```

Yeah, I understand that its a harmless functionally, the missed entries eventually are delivered on the next fram by browser.  but dev-tool overlays (Bun HMR, webpack-dev-server, Vite) surface it as a runtime error, and Sentry/error-monitoring noise treats it as a real crash. 

> So the real issue is that consumers cannot easily fix it from outside the library

### Reproduction

Mount `<File>` (or `FileDiff`) once with line annotations and resize the container. The error fires on every layout shift

```ts
import { File } from '@pierre/diffs/react';
import { useEffect, useState } from 'react';
import { createRoot } from 'react-dom/client';

const sample = `package main
import "fmt"
func main() {
  for i := 0; i < 100; i++ {
    fmt.Println("hello", i)
  }
}
`;

window.addEventListener('error', (e) => {
  if (e.message.includes('ResizeObserver loop')) {
    console.log('REPRO HIT:', e.message);
  }
});

function Demo() {
  // Toggle width on a timer so the observed <code> element is resized after mount
  const [wide, setWide] = useState(false);
  useEffect(() => {
    const id = setInterval(() => setWide((w) => !w), 200);
    return () => clearInterval(id);
  }, []);

  return (
    <div style={{ width: wide ? 800 : 500, border: '1px solid #ccc' }}>
      <File
        file={{ name: 'main.go', contents: sample }}
        options={{
          theme: {
            light: 'github-light-default',
            dark: 'github-dark-default',
          },
[...]

PR #639, opened 2026-05-05
Defer ResizeManager column writes into rAF

lexdotdev:
Address #638


amadeus:
Hey sorry for the delays in getting back to you on this. I've been pretty heads down on the new CodeView component, which does rework a lot of how the ResizeManager code works, so I would want to look at fixing this in the new variants of that. Once CodeView is merged into main, I can dig back into this and see about fixing it.

amadeus:
Should be fixed in latest versions, if there's still a bug, please re-open a new PR against the new updates to resize manager!

PR #644, opened 2026-05-06
View transitions mdo edition

mdo:


mdo:
Brought this back properly rebased, but I think we skip it—our pages load too fast and we get some weirdness. Unsure how we'd animate in the tree and diffs alongside this. Going to punt for now.

PR #645, opened 2026-05-07
Revamp comments

mdo:
- New sidebar comment list
- New form and comment design


apps/docs/app/(diffshub)/(view)/_components/annotation-shared.tsx
```diff
@@ -0,0 +1,77 @@
...
+  'joe',
+  'kris',
+  'mdo',
+  'nicolas',
+  'pia',
+  'toshi',
+  'zac',
+] as const;
+
+function buildPersona(name: string): { name: string; avatarSrc: string } {
```
amadeus: Might be dope set `name: AVATAR_NAMES[number]` i think? or something like that. That would require some downstream stuff to then perfectly match all of that, but would be a nice type security thing

amadeus (commented):
Lookin sharp! Mostly just a types question that doesn't really matter but is just me on my bullshit, can ignore if you want

amadeus (approved):
lfg

PR #647, opened 2026-05-07
Add C language icons to Trees

mdo:


amadeus:
Should we show this same icon for c++?

PR #649, opened 2026-05-07
[diffs] Map common Rails/Ruby filenames to ruby

nickmerwin:
Common Rails/Ruby files currently fall through to plain text because their
filenames/extensions aren't in `EXTENSION_TO_FILE_FORMAT`. Adds entries so they
pick up the bundled Ruby grammar:

- Rails template extensions: `.jbuilder`, `.builder`, `.rabl`, `.arb`
- Other Ruby extensions: `.ru` (Rack config), `.podspec` (CocoaPods)
- Toolchain filenames: `Gemfile`, `Rakefile`, `Guardfile`, `Capfile`,
  `Berksfile`, `Brewfile`, `Vagrantfile`, `Thorfile`, `Appraisals`, `Dangerfile`

Matches the existing pattern of full-filename entries already in the table
(`Dockerfile`, `CMakeLists.txt`, `CODEOWNERS`, `COMMIT_EDITMSG`, `Makefile`).

`.jbuilder` is the Rails default for JSON views — see https://github.com/rails/jbuilder.


amadeus:
Seems reasonable to me! Thanks for the contribution!

PR #652, opened 2026-05-07
feat(trees): add Vue adapter

onmax:
### Description

Adds first-class Vue 3 support for `@pierre/trees` through the new `@pierre/trees/vue` entrypoint. The adapter includes a Vue component, composables for selection and search state, SSR handoff through `preloadedData`, and a Vue browser fixture.

The Trees docs now include Vue getting started and API pages alongside React and vanilla, plus updates to the integration and SSR guides.

### Motivation & Context

Vue users can now adopt Trees without using the React wrapper or dropping down to the vanilla API. The adapter keeps the existing imperative `FileTree` model as the shared contract, so core renderer behavior stays aligned across integrations.

Planning to add support in Reka UI and Nuxt UI upstream as well :)

### Type of changes

      being opened
      that this PR is being opened
- [x] Documentation update

Codex helped implement the Vue adapter, docs, tests, and review fixes under my direction. Also used codex to create a project inside `/tmp` and verify everythign works in a brand-new project using `@Browser use` and tweak things.

### Related issues

None.

PS: I know this have a high cost for maintenance, but the other route I can think of is moving this code to [Reka UI](https://reka-ui.com/) and support there instead. I am also planning to add a PR to add diffs support. Let me know what works for you. Thanks!

### DIFF
--- a/apps/docs/app/(trees)/_docs/DocsPage.tsx
+++ b/apps/docs/app/(trees)/_docs/DocsPage.tsx
@@ -8,6 +8,7 @@ import * as chooseYourIntegrationConstants from '../docs/Guides/ChooseYourIntegr
 import * as customizeIconsConstants from '../docs/Guides/CustomizeIcons/constants';
 import * as getStartedWithReactConstants from '../docs/Guides/GetStartedWithReact/constants';
 import * as getStartedWithVanillaConstants from '../docs/Guides/GetStartedWithVanilla/constants';
+import * as getStartedWithVueConstants from '../docs/Guides/GetStartedWithVue/constants';
 import * as handleLargeTreesEfficientlyConstants from '../docs/Guides/HandleLargeTreesEfficiently/constants';
 import * as navigateSelectionFocusAndSearchConstants from '../docs/Guides/NavigateSelectionFocusAndSearch/constants';
 import * as renameDragAndTriggerItemActionsConstants from '../docs/Guides/RenameDragAndTriggerItemActions/constants';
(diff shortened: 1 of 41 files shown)


SlexAxton:
I'm curious if you have strong opinions about whether this should go directly in the library? On the one hand, AI can probably write decent vue code, and the wrappers are *decently* thin, but on the other hand, I know quite little about vue myself, so it's hard for me to evaluate whether I'm making good updates there.

Would this be better served by a 'community adapters' section in the docs and a project that you or someone else like you maintains?

onmax:
I totally understand the concern.

I can move this into a separate community-maintained package and maintain it myself, so the Vue-specific maintenance burden does not fall on you

Would that direction work for you?

My current thought is to keep it as a single package with a thin documentation layer, exporting the Vue adapter pieces from one place. Before I go that route, I would appreciate any concerns or recommendations about this direction, or anything else that would make this easier to reference from the Pierre docs as a community adapter.

Happy to adapt the approach based on what would fit best on your side.

SlexAxton:
that works for me. Once it's up I can link it in the official docs as well. We appreciate it!

SlexAxton:
re: advice: I'd just stick as close as you can to a port of the react wrapper component as is reasonable in your framework. The key things to double check are that SSR and ssr + slot hydration work as you expect. (e.g. you can render a context menu that hydrates, or a header that is interactive).

There's also a gotcha with hydration and our use of declarative shadow dom. Essentially declarative shadow dom is such that the dom that is eventually generated is *different* than the html that's shipped down the wire (without any js changing it, just by nature of DSD). So the hydration boundary needs to hydrate based on the eventual shadow dom situation, rather than the same markup it sent down the wire from the server. Which usually means you'll have to ignore hydration warnings.

happy to answer any other questions or be generally helpful.

SlexAxton:
Closing in favor of the external ownership. However, please let me know when it's ready to share and we can add it to the docs.

PR #661, opened 2026-05-09
More DiffsHub UI polish

mdo:
- Responsive URL design for homepage
- Clear button for URL on desktop on view page (shows on hover of form)
- Add Alex and Dwayn avatars, fix a few other avatars
- Remove v6-v7 linux URL for now on homepage
- Make submit lighter on mobile for better balance


amadeus:
I feel like it would still be cool to keep the linux link for memes, just probably bury it a bit somehow?

apps/docs/app/(diffshub)/_home/Home.tsx
```diff
@@ -16,10 +16,10 @@ function Divider() {
 }
 
 const EXAMPLE_URLS = [
-  'https://github.com/nodejs/node/pull/59805',
-  'https://github.com/torvalds/linux/compare/v6.0...v7.0',
-  'https://github.com/ghostty-org/ghostty/pull/12291',
-  'https://github.com/pierrecomputer/pierre/commit/0800fb',
+  'nodejs/node/pull/59805',
+  // 'torvalds/linux/compare/v6.0...v7.0',
```
amadeus: can probably yeet this comment since it's added below

packages/diffs/src/managers/InteractionManager.ts
```diff
@@ -54,6 +54,10 @@ export interface SelectionWriteOptions {
   notify?: boolean;
 }
 
+export interface SelectionWriteOptions {
+  notify?: boolean;
+}
+
```
amadeus: This also shouldn't be here

READYOU
amadeus, line 1: This should probably also not be happening?

amadeus (commented):
Overall polish changes look good to me. There's def some rebase corruption with an entire api end point that was added back in, README was weirdly renamed to READYOU and definitely an old interface that was added back in on InteractionManager

amadeus (approved):
👯

PR #666, opened 2026-05-11
[diffs] Remove old /gh route

amadeus:
Now it should just redirect to diffshub.com, since that site is in a pretty good spot


amadeus:
@codex

amadeus:
@codex

PR #668, opened 2026-05-12
Filter by Git Status & Responsive fixes

mdo:


amadeus (approved):

amadeus:
@mdo do we still want to get this merged?

mdo:
> @mdo do we still want to get this merged?

@amadeus I just rebased might, need to clean up more. After that I think I still need to play with a little of the UI and logic here to feel out if the flow is right. Last we talked, we maybe discussed reversing the behavior of the menu items there now. Curious to see what @necolas thinks if he has time to look though too.

necolas:
The git status popover doesn't match the active shiki theme.
My initial thought upon seeing the tick mark over the status symbol was that it was bugged

image

wonder if it might be better to put the check state symbol on the right like it is on the themes popover

image

mdo:
Ya that's a bug from the rebase. It'd be on the right. I can also look into how to match it's theme, too.

necolas (approved):
These icons dont have hover/focus states for light themes

Image

Happy for this to be merged though

PR #673, opened 2026-05-13
[diffshub] Even Moar Polish

amadeus:
A bunch of small tweaks/improvements:
* Add custom scrollbars that match files for the comments section
* Adds an example + button icon thing to the empty state for the comments to better describe how to add them
* Use `whitespace-pre-wrap` on comments in code to allow them to be multi-line
* There was a small bug with macOS on a trackpad where once a drag was started, you could add some horizontal drag to it shifting the code left or right while swiping (fixed with only overflow-auto-y)
* Fix bottom rounded corners on last sticky item to have the hover state better match with the dropshadow/layering

### DIFF
--- a/apps/docs/app/(diffshub)/(view)/_components/CodeViewCommentsList.tsx
+++ b/apps/docs/app/(diffshub)/(view)/_components/CodeViewCommentsList.tsx
@@ -1,7 +1,7 @@
 'use client';
 
 import type { AnnotationSide } from '@pierre/diffs';
-import { IconConvoFill } from '@pierre/icons';
+import { IconConvoFill, IconPlus } from '@pierre/icons';
 import { memo } from 'react';
 
 import { CommentAuthorAvatar } from './annotation-shared';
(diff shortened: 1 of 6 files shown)


apps/docs/app/(diffshub)/(view)/_components/CodeViewCommentsList.tsx
```diff
@@ -29,23 +28,17 @@ function getCommentLineClassName(side: AnnotationSide): string {
...
-          className
-        )}
-      >
+      <div className="text-muted-foreground flex h-full min-h-0 flex-col items-center justify-center gap-2 px-7 text-center text-sm">
         <IconConvoFill size={24} className="mb-2" />
         <div className="flex flex-col">
           <strong className="font-medium">No comments yet</strong>
           <p>
-            Hover over a line and click the blue button to add fake code
+            Hover over a line and click the blue + button to add fake code
```
mdo: Typo?

mdo (commented):
One quick comment from couch review on phone.

apps/docs/app/(diffshub)/(view)/_components/CodeViewCommentsList.tsx
amadeus: The + a typo? or that the line ends with `code`? Because the next line says comment (so code comment, but your comment definitely confused me at first).

That said tho, i kinda want to apply some styles to make it look like the + button

Issue #679, opened 2026-05-15
Feature: diffs: respect prefers-reduced-motion + prewarm scrollTo target


clemg:
### Prerequisites

- [x] I have [searched](https://github.com/pierrecomputer/pierre/issues?q=is%3Aissue) for duplicate or closed feature requests
- [x] I have read the [contributing guidelines](https://github.com/pierrecomputer/pierre/blob/main/.github/CONTRIBUTING.md)

### Proposal

Hi all!

Saw [this tweet](https://x.com/fat/status/2055062500996202579) and I wanted to try the demo. I found out that `prefers-reduced-motion` was missing

Also played a bit with very large diffs on the beta branch for diffshub.com and found out that having instant TPs far enough lags a bit the syntax highlighter. Prewarming it (essentially starting the highlight asap when clicking) reduces, from my testing, the FOUC time by about 33% (~75ms) on my machine

At best it highlights earlier, at worst it does as good as today

### Motivation and context

I noticed that very large files, even more when far away, might take a few 100s ms to get highlighted, which flashes. It's very visible with reduced motion is on (: settings > accessibility > prefers reduced motion)

amadeus:
I believe this is now completed.

PR #680, opened 2026-05-15
Diffs: respect prefers-reduced-motion and prewarm syntax highlights

clemg:
### Description

Implements #679 

Two little changes across two areas:

diffs:
- `resolveEffectiveScrollBehavior` uses `'instant'` when `prefers-reduced-motion` is enabled
- prewarm in `scrollTo` to trigger the syntax highlighting as soon as we can, reducing the length of the FOUC. It's scoped to only warm the file(s) or line(s) that will be visible, ignoring previous and following stuff

### Motivation & Context

See #679 
Saw [the demo on Twitter](https://x.com/fat/status/2055062500996202579), played with it, and saw that this was missing (I use it frequently). Then the syntax highlighting taking too much time bothered me and figured I might as well propose a small fix

### Type of changes

- [x] Refactoring (non-breaking change)
      being opened
      that this PR is being opened
- [ ] Documentation update

(discussed in #679, happy to split into 2 PRs if you want, but I feel like that small and related enough for just 1)

### Checklist

- [x] I have read the

I used CC to review my changes and to test many times with large diffs to confirm I wasn't crazy thinking prewarm was actually useful
Code, issue and PR are human-made

### Related issues

Discussion: #679

### DIFF
--- a/packages/diffs/src/components/CodeView.ts
+++ b/packages/diffs/src/components/CodeView.ts
@@ -765,13 +765,13 @@ export class CodeView<LAnnotation = undefined> {
     }
   }
 
-  private prewarmScrollTarget(target: PendingScrollTarget): void {
+  private primeScrollTarget(target: PendingScrollTarget): void {
     if (target.type === 'position') return;
 
     const item = this.idToItem.get(target.id);
     if (item == null) return;
 
-    item.instance.prewarmWorkerHighlight();
+    item.instance.primeHighlightCache();
   }
 
   private resolveEffectiveScrollBehavior(
@@ -806,7 +806,7 @@ export class CodeView<LAnnotation = undefined> {
       return;
     }
 
-    this.prewarmScrollTarget(pendingTarget);
+    this.primeScrollTarget(pendingTarget);
 
     const behavior = this.resolveEffectiveScrollBehavior(
       pendingTarget,
--- a/packages/diffs/src/components/CodeView.ts
+++ b/packages/diffs/src/components/CodeView.ts
@@ -771,11 +771,7 @@ export class CodeView<LAnnotation = undefined> {
     const item = this.idToItem.get(target.id);
     if (item == null) return;
 
-    if (item.type === 'file') {
-      item.instance.prewarmWorkerHighlight(item.item.file);
-    } else {
-      item.instance.prewarmWorkerHighlight(item.item.fileDiff);
-    }
+    item.instance.prewarmWorkerHighlight();
   }
 
   private resolveEffectiveScrollBehavior(
--- a/packages/diffs/src/components/CodeView.ts
+++ b/packages/diffs/src/components/CodeView.ts
@@ -32,6 +32,7 @@ import type {
 import { areObjectsEqual } from '../utils/areObjectsEqual';
 import { areSelectionsEqual } from '../utils/areSelectionsEqual';
 import { createWindowFromScrollPosition } from '../utils/createWindowFromScrollPosition';
+import { prefersReducedMotion } from '../utils/prefersReducedMotion';
 import { roundToDevicePixel } from '../utils/roundToDevicePixel';
 import type { WorkerPoolManager } from '../worker';
 import type { FileOptions } from './File';
@@ -764,10 +765,27 @@ export class CodeView<LAnnotation = undefined> {
     }
   }
 
+  private prewarmScrollTarget(target: PendingScrollTarget): void {
+    if (target.type === 'position') return;
+
+    const item = this.idToItem.get(target.id);
+    if (item == null) return;
+
+    if (item.type === 'file') {
+      item.instance.prewarmWorkerHighlight(item.item.file);
+    } else {
+      item.instance.prewarmWorkerHighlight(item.item.fileDiff);
+    }
+  }
+
   private resolveEffectiveScrollBehavior(
     target: CodeViewScrollTarget,
     destination: number
   ): Exclude<CodeViewScrollBehavior, 'smooth-auto'> {
+    if (prefersReducedMotion()) {
+      return 'instant';
+    }
+
     if (target.behavior !== 'smooth-auto') {
       return target.behavior ?? 'instant';
     }
@@ -791,6 +809,8 @@ export class CodeView<LAnnotation = undefined> {
     if (destination == null) {
       return;
     }
+
+    this.prewarmScrollTarget(pendingTarget);
 
     const behavior = this.resolveEffectiveScrollBehavior(
       pendingTarget,
--- a/packages/diffs/src/components/File.ts
+++ b/packages/diffs/src/components/File.ts
@@ -295,7 +295,9 @@ export class File<LAnnotation = undefined> {
     this.placeHolder = undefined;
     this.unsafeCSSStyle = undefined;
 
-    if (!recycle) {
+    if (recycle) {
+      this.fileRenderer.recycle();
+    } else {
       this.fileRenderer.cleanUp();
       this.workerManager = undefined;
       this.file = undefined;
--- a/packages/diffs/src/components/File.ts
+++ b/packages/diffs/src/components/File.ts
@@ -4,6 +4,7 @@ import { toHtml } from 'hast-util-to-html';
 import {
   CUSTOM_HEADER_SLOT_ID,
   DEFAULT_THEMES,
+  DEFAULT_TOKENIZE_MAX_LENGTH,
   DIFFS_TAG_NAME,
   EMPTY_RENDER_RANGE,
   HEADER_METADATA_SLOT_ID,
@@ -48,6 +49,7 @@ import {
 import { getLineAnnotationName } from '../utils/getLineAnnotationName';
 import { getOrCreateCodeNode } from '../utils/getOrCreateCodeNode';
 import { upsertHostThemeStyle } from '../utils/hostTheme';
+import { isFilePlainText } from '../utils/isFilePlainText';
 import { prerenderHTMLIfNecessary } from '../utils/prerenderHTMLIfNecessary';
 import { getMeasuredScrollbarGutter } from '../utils/scrollbarGutter';
 import { setPreNodeProperties } from '../utils/setWrapperNodeProps';
@@ -637,8 +639,18 @@ export class File<LAnnotation = undefined> {
   }
 
   public prewarmWorkerHighlight(): void {
-    if (this.file == null) return;
-    this.fileRenderer.prewarmWorkerHighlight(this.file);
+    const { file, workerManager } = this;
+    if (file == null || workerManager == null || isFilePlainText(file)) {
+      return;
+    }
+    const lines = this.fileRenderer.getOrCreateLineCache(file);
+    if (
+      lines.length >
+      (this.options.tokenizeMaxLength ?? DEFAULT_TOKENIZE_MAX_LENGTH)
+    ) {
+      return;
+    }
+    workerManager.prewarmFileAST(file);
   }
 
   private cleanChildNodes() {
--- a/packages/diffs/src/components/File.ts
+++ b/packages/diffs/src/components/File.ts
@@ -634,6 +634,11 @@ export class File<LAnnotation = undefined> {
     return true;
   }
 
+  public prewarmWorkerHighlight(file: FileContents): void {
+    if (this.workerManager?.isWorkingPool() !== true) return;
+    this.workerManager.highlightFileAST(this.fileRenderer, file);
+  }
+
   private cleanChildNodes() {
     this.resizeManager.cleanUp();
     this.interactionManager.cleanUp();
--- a/packages/diffs/src/components/File.ts
+++ b/packages/diffs/src/components/File.ts
@@ -638,7 +638,7 @@ export class File<LAnnotation = undefined> {
     return true;
   }
 
-  public prewarmWorkerHighlight(): void {
+  public primeHighlightCache(): void {
     const { file, workerManager } = this;
     if (file == null || workerManager == null || isFilePlainText(file)) {
       return;
@@ -650,7 +650,7 @@ export class File<LAnnotation = undefined> {
     ) {
       return;
     }
-    workerManager.prewarmFileAST(file);
+    workerManager.primeFileHighlightCache(file);
   }
 
   private cleanChildNodes() {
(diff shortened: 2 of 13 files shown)


packages/diffs/src/components/File.ts
```diff
@@ -634,6 +634,11 @@ export class File<LAnnotation = undefined> {
     return true;
   }
 
+  public prewarmHighlight(file: FileContents): void {
```
amadeus: I think we should name this `prewarmWorkerHighlight` so it's explicit about what it's doing

packages/diffs/src/utils/prefersReducedMotion.ts
```diff
@@ -0,0 +1,9 @@
+export function prefersReducedMotion(): boolean {
```
amadeus, line 1: We should make this a library level export so the diffshub implementation can use it.

apps/docs/app/(diffshub)/_home/ScrollDownButton.tsx
```diff
@@ -8,11 +8,14 @@ export function ScrollDownButton() {
       type="button"
       aria-label="Scroll down"
       className="absolute bottom-[calc(1rem+env(safe-area-inset-bottom))] left-1/2 -translate-x-1/2 cursor-pointer md:hidden"
-      onClick={() =>
+      onClick={() => {
+        const reduceMotion = window.matchMedia(
+          '(prefers-reduced-motion: reduce)'
+        ).matches;
```
amadeus: Should use our utility function that we export.

packages/diffs/src/components/CodeView.ts
```diff
@@ -747,10 +748,29 @@ export class CodeView<LAnnotation = undefined> {
     }
   }
 
+  private prewarmScrollTarget(destination: number): void {
```
amadeus: This shouldn't take a destination target number, but it imo it should take the scrollTo item, and it should resolve a target item or items based on that which will usually be a far more efficient than a linear scan through every item.

Actually more that I think about it, maybe lets ONLY pre-warm if it's a target item, line or range, since we can efficiently resolve those.

amadeus (commented):
Hey thanks for the PR! I think these are generally good things we should do, however I think there's some architectural changes we should make to this implementation (thus the whole note about chatting with us first before opening a PR).

Might also be good to break this into 2 changes (one for CodeView, and one for DiffsHub).

apps/docs/app/(diffshub)/_home/ScrollDownButton.tsx
clemg: Will be in a second PR then!

packages/diffs/src/components/CodeView.ts
clemg: Yep, originally that's what I did but it wouldn't include positions
I guess if we want to avoid O(n) for positions, we would have to binary search it or other trickery

clemg:
Thank you for the fast response! I'm sorry about missing the discussion, I don't know why I felt confident that it could be skipped, will follow the rules for next time
I kept this PR for CodeView and will open a new one for updating the demo

note: what do you think about exposing publicly the prewarm functions, so that we could call them sooner on hovering the file names in the file tree in the demo? Maybe it would include some false positives though (hovering a file name but not clicking it)

amadeus:
Word, thanks for the changes! I just had a realization that the prewarm methods can't and shouldn't be on the FileDiff/File because it's the wrong scope and can create extra work (didn't realize this on my first review). The renderers have idempotency built in. I'm quickly having codex do an amendment to this and will probably push a commit to this branch shortly (easier for me to just directionally vibe this than try to do back and forths on the review process/explanations for it)

packages/diffs/src/renderers/DiffHunksRenderer.ts
```diff
@@ -82,7 +83,7 @@ interface PushLineWithAnnotation {
 
 interface GetRenderOptionsReturn {
   options: RenderDiffOptions;
-  forceRender: boolean;
+  forceHighlight: boolean;
```
amadeus, line 86: Renamed this because because forceRender didn't make sense in context of the render function, it actually represented a forceHighlight (mostly atoning for my past mistakes here)

packages/diffs/src/renderers/DiffHunksRenderer.ts
```diff
@@ -570,6 +572,42 @@ export class DiffHunksRenderer<LAnnotation = undefined> {
...
+  public prewarmWorkerHighlight(diff: FileDiffMetadata): void {
+    if (this.workerManager?.isWorkingPool() !== true) {
+      return;
+    }
+
+    const { options } = this.getRenderOptions(diff);
+    if (
+      isDiffPlainText(diff) ||
+      isDiffMassive(diff, this.getTokenizeMaxLength()) ||
+      this.hasHighlightedRenderCache(diff, options)
```
amadeus: Ensure we don't kick off a highlight task if we already have a highlighted renderCache

packages/diffs/src/renderers/DiffHunksRenderer.ts
```diff
@@ -570,6 +572,44 @@ export class DiffHunksRenderer<LAnnotation = undefined> {
...
+    ) {
+      return;
+    }
+
+    const cache = this.getMatchingWorkerResultCache(diff, options);
+    if (cache != null) {
+      if (this.renderCache == null) {
+        this.renderCache = {
```
amadeus: If we have a valid cache and no renderCache, it simply means we just need to populate a renderCache for next render cycle.

amadeus:
Lol i know i just pushed 2 commits, but i realized all this is wrong and it needs to go in another direction to be more implementation sound. so more commits incoming 😬

clemg:
What's wrong with the way you went?

amadeus:
Essentially the version i wrote here utilized renderCache, however that's a poor system for pre-caching ast's because renderCache is also something that gets nulled out when recycled during virtualization.

Also using renderCache ended up creating a bunch of new internal scenarios about the state of renderCache, making it more complicated to understand and reason about (it was no longer being tied to render calls). It also meant the component would start reacting to onHighlightSuccess events which may or may not be applicable if the component is no longer mounted. It could also create scenarios where ASTs get leaked and never GCd if the pre-warmed elements never render again.

So, the better system will be:

We already have a highlighted ast caching system with the WorkerPool, which is an LRU cache of responses from the WorkerPool that have a cacheKey. The renderers are already built to hydrate with this cache anyways, so the only thing we really need to tweak with WorkerPool is a system where if request for highlighting is still in progress, we just have the renderer piggyback onto it. This will also allow me to improve the system so if you mount multiple components with the same file or diff on your page, they could share in that same request which I think would also be a good thing to have.

Anyways, hopefully that makes sense, and sorry for the thrash, I'm one of those people that's pretty dumb right off the bat and I have to tinker with things to really understand them and build a mental model for them.

packages/diffs/src/components/File.ts
```diff
@@ -634,6 +638,21 @@ export class File<LAnnotation = undefined> {
     return true;
   }
 
+  public primeHighlightCache(): void {
```
amadeus, line 641: Felt like `prime` was a cleaner name than `prewarm`

packages/diffs/src/renderers/DiffHunksRenderer.ts
```diff
@@ -225,10 +226,7 @@ export class DiffHunksRenderer<LAnnotation = undefined> {
   }
 
   public cleanUp(): void {
-    this.highlighter = undefined;
-    this.diff = undefined;
-    this.renderCache = undefined;
-    this.workerManager?.cleanUpPendingTasks(this);
+    this.recycle();
```
amadeus, line 229: Just a bit of house keeping here.

packages/diffs/src/worker/WorkerPoolManager.ts
```diff
@@ -83,24 +85,35 @@ interface ThemeSubscriber {
...
+
+  // Tasks that map to highlightKey are essentially singular, so we only
+  // need one map to include queued and active
+  private taskByHighlightKey = new Map<string, RenderTask>();
+
+  // Tasks that have already been sent to a worker and are awaiting a response.
+  private activeTaskById = new Map<WorkerRequestId, AllWorkerTasks>();
+  private activeRequestByInstance = new Map<
+    RenderTaskInstance,
+    WorkerRequestId
```
amadeus, line 111: Basically took the time while making these updates to try and have a more consistent naming structure.

queued: waiting for a worker
active: being sent to the worker or waiting on a response.

amadeus:
@codex review

amadeus:
@codex review

amadeus:
Thanks for bearing with me through this! 🙏

clemg:
No worries, it's very cool seeing you work your way through this! (including dead ends)
I didn't thought that this would be a PR that big, even though you included additional refactoring

I have two small questions:
- we still have public highlighting methods in File/FileDiff, even though the implementation is delegated to the renderer, is that what you meant in your first comment?
- `areDiffTargetsEqual` and `areFileRenderOptionsEqual` are public exports, but no consumer would rightfully use this right?

This has been very instructive 😁

amadeus:
Probably going to be a bit long winded, but this is essentially my thought process as I went through this project. A lot of it works because I've built most of this system so I have a ton of context on how it works (it was my design, both for better and for worse 😅) and was thus able to reason through all these scenarios. It would be hard to expect someone just jumping into the code to be able to work through this stuff very efficiently at first.

> we still have public highlighting methods in File/FileDiff, even though the implementation is delegated to the renderer, is that what you meant in your first comment?

So yeah, the API for this still makes sense to be exposed conveniently on the instance, but actual owner of the work is and should be the WorkerPoolManager. It also allows for more convenient usage of priming the cache without even requiring an instance, but also nice to have simple instance based primers for CodeView directly since it controls the virtualization so explicitly.

So a bit about how i got there:

In your original version, we were creating a new type of relationship/contract between the worker pool and components (components being File and FileDiff) by giving the component the ability to directly trigger highlights. While the code was fairly simple to reason about and kept changes minimal, it now meant we created a new place where highlights could be triggered from (this ultimately creates more complexity/race conditions around rendering in general given we have multiple paths). Some of the main issues with this initial architecture were:

* Renderers had some built in logic to prevent kicking off unnecessary or duplicate highlight jobs but components did not
  * This was a side effect of changing the component/workerpool contract I mentioned above. We now created a scenario that whenever a highlight job was kicked off, it would always run.
  * Therefore with the existing prewarm APIs, we'd always waste cycles and a worker thread to highlight a file/diff even if it's been highlighted before
* Additionally it changed the meaning/intention of renderCache
  * `renderCache` is supposed to imply -- `i've rendered content, and this is the cached data from that last render`.  Now it meant that OR here's a cache to build from on the next render
  * We now had scenarios where we could potentially leak fully highlighted ASTs that would just be held indefinitely on an instance until the component was fully rendered and recycled again.
    * i.e. you click a diff to scroll to it but something interrupts that scroll animation and we never reach the destination. now we have a fully populated renderCache that holds onto the AST on the instance.
    * The LRU could evict that AST at some point, but if we never return to the diff to render it, we'll never trigger the recycle path to null out the existing render cache.
* Also at the end of the highlight process, we would be kicking off various reactionary APIs that generally would mostly assume we were rendered/mounted, and therefore be potentially doing work that isn't necessary or kicking off listener callbacks that might now have confusing implications
  * This was a side effect of WorkerPool being architected in a way where it assumed there was a renderer attached to every highlight job.

Initially in my refactor, I was focused on the first issue. Not doing unnecessary work and keeping the old renderer/worker pool contract intact felt like the right fix. I got that working and committed. But then was I was reviewing the committed code, I started having a lot of alarm bells go off in my head around how renderCache was being used, and how we were doing duplicate work in both `warm` and `render` with the renderCache. This led me to the realization that now we had situations where we could leak ASTs if we pre-warmed but never actually rendered and properly recycled that component.

It was from here where I realized I needed to take a step back and re-think this entire architecture. We have an LRU cache which will manage this memory usage for us, and renderers already know how to hydrate themselves from it, so we should probably be utilizing these existing APIs instead of trying to invent an entirely new path. The side effect of this of course is that we had to potentially re-think how tasks worked within WorkerPoolManager: they now might not have instances attached to them anymore, and they might need to support adding/removing instances through the lifecycle of the task itself.

The required changes would have to encompass:
* Ability to specify a task as pre-warm (or what i renamed to prime later on) which would mean it would run regardless of whether it had instances attached or not
  * We also still need to keep the existing system for aggressive scrolling that would allow us to drop tasks if the component unmounts before it's job starts under normal circumstances
* A new type of key-ing system for highlighting that naturally worked with cacheKey that could enable us to attach instances to inflight jobs if they require the same work and/or support jobs that are prime only
  * Because LRU only works with a cacheKey, it naturally ensured no silly footguns around memory management
* Also led me to add a bit more defensive-ness in the worker pool to not kick off jobs we already have an LRU cache for (even though renderers did this for us already, now that we had these public worker apis as well we needed to include some of this hardening)

Anyways, with this new architecture it meant we could keep a lot of the existing component/renderer architecture in place; same mental models, same flow of data, and at the same time improve the CodeView experience, ultimately do less work, and provide new global APIs for priming cache that can be used safely by anyone.

> areDiffTargetsEqual and areFileRenderOptionsEqual are public exports, but no consumer would rightfully use this right?

I just have a general philosophy of this project to try and export as much as possible, to give people the tools to use as many pieces as they want however they want. It's unlikely most people will ever need those utils, but it can't really hurt to make them available, and could potentially provide value if someone wanted to duplicate some internal component but change certain aspects of how it's used.

clemg:
Hey, wow, what an answer!

It definitely clears some spots that I misunderstood/didn't see that way, and actually makes a lot of sense lol

I have two new questions though:
- when the renderer switches fiels mid-tokenization, I think it stays subscribed to the old file's task and ends up overwriting its renderCache with the stale result, then self-correcting on the next render call. I guess that we could add something like `this.detachInstanceFromActiveTasks(instance);` after detaching from queued tasks, effectively dropping the old file's task as soon as we don't need it anymore?
- is it worth aborting in-flight render-task tracking at the start of `setRenderOptions` to avoid the cache-fill-then-clear churn during theme swaps for instance?

About the last part and the functions being public: yeah I get it, I was just thinking about what I would like as a user implementing this lib, which would probably be something around the lines of "letting the lib devs handling as much as they can to avoid doing something bad myself", but I guess it can't hurt to let this be public!

I hope that my little initial PR that turned into a bigger thing, + you explaining all of this isn't too annoying!
It sure is interesting to me 🤠

Thank you again

PR #683, opened 2026-05-16
Misc Release Prep

amadeus:
Move all hardcoded blobs to R2. Had to also fix an issue with the path loader with duplicate paths due to a deletion and then a symlink in linux.

Currently hosting these assets on R2 under blobs.diffs.wtf


amadeus:
@codex review

PR #686, opened 2026-05-16
[diffs] reduce patch parsing string retention

amadeus:
This was definitely a moment of jesus take the wheel. I let codex go a bit deeper on figuring out how to optimize memory usage and performance when it came to parsing diffs.

I don't have the numbers but i think parsing was improved by at like 50% and memory usage scales immensely as the file size grows.

An example of improved memory usage:

**BEFORE:**
IMG_5353

**AFTER:**
IMG_6803


amadeus:
@codex review

PR #688, opened 2026-05-16
[diffs] ResizeManager: use shared ResizeObserver

amadeus:
As the title says, make all ResizeManager's share a single ResizeObserver.


amadeus:
@codex review

PR #694, opened 2026-05-18
[diffs] fix pluralization of '0 unmodified lines'

necolas:
Does what it says on the tin

### DIFF
--- a/packages/diffs/src/renderers/DiffHunksRenderer.ts
+++ b/packages/diffs/src/renderers/DiffHunksRenderer.ts
@@ -1328,7 +1328,7 @@ export class DiffHunksRenderer<LAnnotation = undefined> {
 }
 
 function getModifiedLinesString(lines: number) {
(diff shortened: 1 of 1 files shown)


amadeus:
Might be worth applying this on the beta instead of stable, since i don't think we'll do a stable release before 1.2 at this point

SlexAxton:
Might not be worth it, but the More Official™ to do this would be:

```ts
const enCardinalRules = new Intl.PluralRules("en-US"); // can cache this once

const pluralKey = enCardinalRules.select(lines);

if (pluralKey === 'one') {
  return `${lines} unmodified line`;
} else if (pluralKey === 'other') {
  return `${lines} unmodified lines`;
} else {
  throw new Error(`Unexpected plural key: ${pluralKey}`);
}
```

amadeus:
@necolas do we still want to get this in?

PR #695, opened 2026-05-18
[diffs] fix ResizeManager double-delete on number element swap

necolas:
The old numberElement was unobserved but never removed from the function-scoped snapshot, so the trailing reaping loop treated it as stale and striped inline styles.


amadeus:
ResizeManager changes significantly in the beta, we should probably look at seeing if this is necessary there?

PR #696, opened 2026-05-18
[diffs] cancel previous source stream in FileStream.setupStream

necolas:
Be sure that a second setupStream cancels the first, rather than simply overwriting it.


amadeus (approved):
lgtm!

haven't touched FileStream in any of the beta, so it's safe to merge this

Issue #676, opened 2026-05-14
Feature: Trees — public API to scroll a path into view programmatically


srid:
## Use case

I'm using `@pierre/trees` (vanilla) from a SolidJS wrapper. The host application drives selection from outside Pierre — e.g., a click on a `path:line` reference in another panel needs to (a) open the file in a viewer, (b) reveal and select the corresponding row in the file tree.

The selection part works fine: passing `initialSelectedPaths` at construction or calling `tree.getItem(path)?.select()` post-mount lands `aria-selected="true"` on the row.

**What's missing**: when the targeted path is far enough down that the virtualizer hasn't rendered it (deep in a large worktree), the row never enters the visible window. The tree shows the user that "nothing is selected" even though `#selectedPaths` contains the path.

## Current behavior (Pierre 1.0.0-beta.3)

`scrollFocusedRowIntoView` is gated on `shouldOwnDomFocus`:

```js
// dist/render/FileTreeView.js:1557
if (shouldRestoreStickyKeyboardScrollTop
    || shouldRestoreStickyFocusedRowViewportOffset
    || shouldRestoreStickyKeyboardViewportOffset
    || shouldRestoreFocusedRowViewportOffset
    || (shouldOwnDomFocus && focusedPathChanged && pendingStickyFocusPath !== focusedPath
        && !shouldPreserveStickyKeyboardFocusViewport
        && scrollFocusedRowIntoView(scrollElement, focusedIndex, itemHeight, resolvedViewportHeight, stickyOverlayHeight)))
  updateViewportRef.current();
```

And `shouldOwnDomFocus = domFocusOwnerRef.current || focusWithinTree` (line 1549). `domFocusOwnerRef.current = true` is set only inside user-initiated handlers (row click line 1733, keyboard nav lines 1216/1222/1286, focus event line 782, etc.). `focusWithinTree` requires `document.activeElement` to already be inside the tree root (line 1548).

The public `FileTree` class surface (`dist/render/FileTree.d.ts`) exposes `focusPath(path)` and `getItem(path)` (which returns a handle with `focus()`/`select()`), but none of these set `domFocusOwnerRef` or otherwise trigger scroll. There's no `scrollToPath` / `revealPath` / `scrollIntoView` method.

## Empirical confirmation

Probe: tree with 60 files alphabetically before `zzz_target.txt`. Navigate via the host application (which calls `tree.getItem("zzz_target.txt")?.select()` and seeds `initialSelectedPaths`). After navigation:

[...]

PR #699, opened 2026-05-18
[trees] Add support for path restoration and scrolling to paths

SlexAxton:
# Summary

 Adds programmatic tree scrolling and improves initial mount behavior for active items.

## Changes

 - Mounts trees with an initial active/selected item scrolled into the visible virtualized slice.
 - Adds `FileTree.scrollToPath(path, options?)`.
     - `offset: "nearest"` (default), `"top"`, or `"center"`.
     - `focus: true` by default; pass false to scroll without changing model focus.
     - Does not move DOM focus.
     - No-ops when the path is not in the current visible projection.
 - Adds shared scroll-target math with unit coverage.
 - Handles sticky ancestor rows when computing scroll offsets.
 - Clears one-shot scroll requests after the view consumes them.
 - Updates docs and README with the new API.
 - Updates the trees-dev React demo:
     - initial active item starts in view
     - buttons for first/middle/last `scrollToPath`
     - checkbox to toggle whether `scrollToPath` updates model focus


SlexAxton:
fixes #676

PR #700, opened 2026-05-18
[diffshub] stream tree paths incrementally to fix O(N²) publish slowdown

necolas:
Streaming a large compare (e.g. torvalds/linux v6.0..v7.0) made the Diff Stats counter tick more slowly the longer the stream ran, and stuttered scroll on every tick. Each streamed batch published a fresh CodeViewFileTreeSource, and CodeViewFileTree responded by calling `model.resetPaths(source.paths)`, which constructs a brand new PathStore over the full accumulated path list. Per publish that is O(N log N) where N is total paths streamed so far, so across ~80 publishes the total work is roughly O(N² log N) on the main thread and each publish blocks input proportionally to N. The trees model already exposes a localized mutation fast path via `model.batch` / `model.add`; diffshub just wasn't using it.

Link each tree-source snapshot to the prior one through a new `previousSource` field. The accumulator now also owns a persistent `rankByPath` map and a stable sort comparator over it, so we stop rebuilding the comparator's rank map on every snapshot. When CodeViewFileTree sees a snapshot whose `previousSource` matches the one it last applied, it batches the tail of new paths into the model with `{ type: 'add', path }` ops instead of resetting. `resetPaths` is now reserved for the initial mount and any non-append change (e.g. a new request).

This drops per-publish tree work from O(N) to O(delta), keeps the counter incrementing at a steady rate as the stream progresses, and removes the recurring long task that was starving scroll during streaming.

### DIFF
--- a/apps/docs/app/(diffshub)/(view)/_components/CodeViewFileTree.tsx
+++ b/apps/docs/app/(diffshub)/(view)/_components/CodeViewFileTree.tsx
@@ -1,7 +1,10 @@
 'use client';
 
 import { useStableCallback } from '@pierre/diffs/react';
-import type { FileTree as FileTreeModel } from '@pierre/trees';
+import type {
+  FileTreeBatchOperation,
+  FileTree as FileTreeModel,
+} from '@pierre/trees';
 import { FileTree, useFileTree } from '@pierre/trees/react';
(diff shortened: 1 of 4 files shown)


amadeus (approved):

amadeus:
Also this fixes the folders getting uncollapsed while streaming in results! 🙏

PR #702, opened 2026-05-19
Update to Pierre Soft themes

mdo:
- Updates to latest `@pierre/theme` which adds light and dark soft themes
- Imports both soft and regular themes now
- ~~Defaults Diffs and Trees examples to use the new soft themes, as well as DiffsHub~~ Defaults DiffsHub to use new soft themes
- Updates some static values to use new colors from soft themes


packages/diffs/src/constants.ts
```diff
@@ -45,8 +45,8 @@ export const HEADER_METADATA_SLOT_ID = 'header-metadata';
 export const CUSTOM_HEADER_SLOT_ID = 'header-custom';
 
 export const DEFAULT_THEMES: ThemesType = {
-  dark: 'pierre-dark',
-  light: 'pierre-light',
+  dark: 'pierre-dark-soft',
+  light: 'pierre-light-soft',
```
amadeus: Should we really change the defaults?

apps/docs/app/(diffs)/_components/WorkerPoolContext.tsx
```diff
@@ -46,7 +46,7 @@ const PoolOptions: WorkerPoolOptions = {
 };
 
 const HighlighterOptions: WorkerInitializationRenderOptions = {
-  theme: { dark: 'pierre-dark', light: 'pierre-light' },
```
amadeus, line 49: we use the worker pool in some of the diffs pages, assuming we don't want to change themes on the diffs.com pages? So we might not be able to change this here.

apps/docs/app/(diffshub)/(view)/_components/CodeViewFileTree.tsx
```diff
@@ -126,7 +139,7 @@ export const CodeViewFileTree = memo(function CodeViewFileTree({
     <FileTree
       className="h-full min-h-0 overflow-auto overscroll-contain md:ml-3"
       model={model}
-      style={DENSITY_OVERRIDE_STYLES}
+      style={{ ...themeStyles, ...DENSITY_OVERRIDE_STYLES }}
```
amadeus: Can we wrap this in a useMemo?

apps/docs/app/(diffshub)/_home/Home.tsx
```diff
@@ -82,35 +83,7 @@ export default function DiffshubHome() {
...
-          >
-            <code className="text-foreground/75">
-              @
-              <code className="decoration-muted-foreground group-hover:decoration-foreground group-transition-all underline decoration-[1px] underline-offset-3">
-                pierre/trees
-              </code>
-            </code>
-          </Link>
-          .
+          using the new CodeView component.
```
amadeus, line 86: Do we want to call out trees? also, we could link CodeView section in the docs: http://diffs.com/docs#codeview (will be broken until release)

apps/docs/app/(diffs)/_components/WorkerPoolContext.tsx
amadeus: now that we talked about it -- i think we need to toggle this based on the site build type

apps/docs/app/(diffshub)/_home/Home.tsx
mdo: Trees and Diffs are called out in the first FAQ. Felt like we had too many links in top paragraph and they took away from CodeView launch/demo.

amadeus (approved):

PR #708, opened 2026-05-19
[diffshub] add header toggle to collapse/expand all files

necolas:
Adds a two-state button next to the split/unified toggle that flips between expanded (default) and collapsed for every loaded file. New files streamed in while the toggle is in collapsed mode arrive collapsed. Clicking a file in the sidebar tree expands it (cumulative) in addition to scrolling.

New icon in header:

image

### DIFF
--- a/apps/docs/app/(diffshub)/(view)/_components/usePatchLoader.ts
+++ b/apps/docs/app/(diffshub)/(view)/_components/usePatchLoader.ts
@@ -54,6 +54,7 @@ const GENERIC_PATCH_LOAD_ERROR_MESSAGE =
   'We couldn’t load that diff. Check the URL and try again.';
 
 interface UsePatchLoaderOptions {
+  collapseMode: 'expanded' | 'collapsed';
   domain?: string;
   onLoadStart(): void;
   path: string;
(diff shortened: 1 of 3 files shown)


apps/docs/app/(diffshub)/(view)/_components/usePatchLoader.ts
```diff
@@ -103,6 +106,59 @@ export function usePatchLoader({
...
+        }
+        item.collapsed = targetCollapsed;
+        item.version = getNextItemVersion(item);
+        if (viewer.updateItem(item)) {
+          didChange = true;
+        }
+      }
+      if (didChange) {
+        viewer.getInstance()?.render(true);
+      }
```
amadeus: We _probably_ don't need this, cause `updateItem` will trigger these updates for us, if it's needed

necolas:
@codex

necolas:
@codex

apps/docs/app/(diffshub)/(view)/_components/usePatchLoader.ts
necolas: fixed

amadeus:
btw, i'm working on a quick side thing that i think can make this way more optimized. your code is fine, just updating some internal things to make it faster

mdo (approved):
Nice, looks good!

Issue #707, opened 2026-05-19
Bug: remap['file-tree-icon-file'] unreachable as fallback with set: 'complete' or set: 'standard'


StefanDBTLabs:
### Prerequisites

- [x] I have [searched](https://github.com/pierrecomputer/pierre/issues?q=is%3Aissue) for duplicate or closed issues
- [x] I have read the [contributing guidelines](https://github.com/pierrecomputer/pierre/blob/main/.github/CONTRIBUTING.md)

### Describe the issue

When using `set: 'complete'` or `set: 'standard'`, the `remap` entry for `'file-tree-icon-file'` has no effect as a fallback for unknown file types.

**Root cause:** `resolveBuiltInFileIconToken` returns `'default'` for any extension it doesn't recognize, rather than `undefined`. Because the resolver always returns a non-null token for these sets, `resolveIcon` returns early with the built-in default icon and never reaches the `iconRemap` check at the bottom of the function.

With `set: 'minimal'` or `set: 'none'`, `resolveBuiltInFileIconToken` returns `undefined`, so the fallback remap is reachable. This means custom fallback icons and colored icons (`colored: true`, only effective with `set: 'complete'`) are mutually exclusive — there's no single configuration that supports both.

**Expected behavior:** `remap['file-tree-icon-file']` should be usable as a fallback for unknown file types regardless of which built-in set is active, including `set: 'complete'`.

**Workaround:** Use `set: 'minimal'`, inject the complete sprite manually via `spriteSheet`, and enumerate all desired extensions explicitly in `byFileExtension`. Icon colors must also be injected manually via `unsafeCSS` since `colored: true` has no effect with `set: 'minimal'`. This can work but is very awkward.

* this issue was written with the help of AI tools.

### Reproduction

```
import { FileTree as FileTreeModel } from '@pierre/trees';
import { FileTree } from '@pierre/trees/react';

const model = new FileTreeModel({
  paths: ['src/index.ts', 'src/unknown.xyz'],
  icons: {
    set: 'complete',
    remap: {
      // Expected: .xyz files use this icon. Actual: never fires.
      'file-tree-icon-file': { name: 'my-custom-icon', viewBox: '0 0 16 16' },
    },
    spriteSheet: `
      <svg xmlns="http://www.w3.org/2000/svg" style="display:none">
        <symbol id="my-custom-icon" viewBox="0 0 16 16">
          <rect width="16" height="16" fill="red" />
        </symbol>
      </svg>
    `,
  },
[...]

PR #710, opened 2026-05-19
fix(trees): I want to use `remap` when the set is `standard` or `complete` but ATM those sets skip over `remap`

StefanDBTLabs:
### Description

`remap` is a great way to quickly override icons. But remap only works when the `set` is `minimal`. Specifically I am trying to remap `file-tree-icon-file`. [The docs seem to suggest this should work](https://trees.software/docs#customize-icons) but it did not.

### Motivation & Context

With `set:'standard'` or `set:'complete'`, `resolveBuiltInFileIconToken` always returns a token — including `'default'` for files with no specific extension match. This causes the `iconRemap` check to be skipped entirely, making `remap['file-tree-icon-file']` unreachable as a generic file icon fallback unless the caller uses `set:'minimal'`.

**Root cause:** In `iconResolver.ts`, the built-in token path returns early before the `iconRemap` check whenever `builtInToken != null`. Since `resolveBuiltInFileIconToken` returns `'default'` as a catch-all for `standard`/`complete`, the remap is never consulted.

When the resolved built-in token is the generic `'default'`, we now check `iconRemap['file-tree-icon-file']` first. Specific tokens (`typescript`, `python`, etc.) are unaffected — they continue to short-circuit before the remap check. Falls back to `file-tree-builtin-default` if no remap entry is defined, preserving existing behaviour.

| Scenario | Before | After |
|---|---|---|
| `set:'complete'` + unknown extension + remap defined | `#file-tree-builtin-default` | remap icon ✓ |
| `set:'complete'` + known extension (`.ts`) + remap defined | `#file-tree-builtin-typescript` | `#file-tree-builtin-typescript` ✓ |
| `set:'complete'` + unknown extension + no remap | `#file-tree-builtin-default` | `#file-tree-builtin-default` ✓ |
| `set:'minimal'` + remap defined | remap icon ✓ | remap icon ✓ |

### Type of changes

      being opened
      that this PR is being opened
- [ ] Documentation update

### Checklist

- [x] I have added tests to cover my changes
- [x] All new and existing tests pass (`bun test`)

I found this issue while using claude and trying to replace the default file icon while using the complete set. After reading the docs and looking at the code with claude's help I found that rename only works in the minimal set. So I set minimal and my remap and then rebuilt my own set by exporting all the svg icons and css styles. While this worked it was very awkward.

I then used claude to help write the bug ticket and finally to write the code for this PR, which I carefully reviewed, and help with parts of the explanation of the technical details. I hope my balance of human work and LLMs is acceptable and if not I apologize.

### Related issues

pierrecomputer/pierre#707

### DIFF
--- a/packages/trees/src/render/iconResolver.ts
+++ b/packages/trees/src/render/iconResolver.ts
@@ -109,6 +109,15 @@ export function createFileTreeIconResolver(icons?: FileTreeIcons): {
         extensionCandidates
       );
       if (builtInToken != null && normalizedIcons.set !== 'none') {
+        // When the resolved token is the generic 'default' fallback, let the
+        // user's remap['file-tree-icon-file'] win — that slot is explicitly
+        // meant to override the generic file placeholder.
+        if (builtInToken === 'default') {
+          const remappedEntry = iconRemap?.[name];
+          if (remappedEntry != null) {
+            return remapEntryToIcon(remappedEntry, name);
+          }
+        }
         return {
           name: getBuiltInFileIconName(builtInToken),
           remappedFrom: name,
--- a/packages/trees/test/file-tree-icon-config.test.ts
+++ b/packages/trees/test/file-tree-icon-config.test.ts
@@ -102,6 +102,53 @@ describe('file-tree icon config', () => {
     }
   });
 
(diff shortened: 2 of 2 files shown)


SlexAxton:
Thank you for your contribution. You are correct that this was the intent. The code looks good to me and thank you for the test. 🙏

PR #711, opened 2026-05-19
[diffshub] comments sidebar polish

necolas:
- Show a numeric count badge on the sidebar comment icon when there are saved comments, sized so the icon-only button widens just enough to fit the count.
- Make each file-path header in the comments list a clickable button that scrolls the viewer to that file (reusing the existing onSelectItem flow), and allow drag-selecting text inside the path and comment rows by guarding onClick against a non-empty window.getSelection().
- Classify each saved comment's anchor line as 'change' or 'context' at save time by walking the hunkContent, and render context-line comments as a muted "Line N" instead of a misleading "Line +N" / "Line -N".


necolas:
@codex

amadeus (approved):
clever!

mdo (approved):
Nice!

PR #714, opened 2026-05-19
[diffshub] split _components into _components/_lib

necolas:
`(view)/_components` mixed real components with plain modules. Move every non-component into a new `(view)/_lib`. Plain modules are pure renames; `utils.ts` splits into five focused files by concern (viewer route, diff item guards, change-type mapping, patch order, comment sidebar).

Sibling imports inside `_lib/` and `_components/` and the handful of external consumers (`Home.tsx`, `DiffUrlForm.tsx`, the catch-all route page, and both tests) are repointed accordingly. No behavior change.


amadeus:
@necolas is this still a valid thing to do?

PR #722, opened 2026-05-20
[trees] add support for incremental git status

amadeus:
One thing that would be nice to have for diffshub, is the ability to partially update git statuses while incrementally adding files.

This was completely AI generated, so not sure if it's even the right way/desired way to do this.

I have another PR that utilizes this that I can merge after if you're happy with this.


SlexAxton (approved):

SlexAxton:
works for me

PR #727, opened 2026-05-20
Body class to scope styles

mdo:
Solving a scoping problem of some silly styles I had added just for DiffsHub, so now we can better scope by a body class. This is really only useful for DiffsHub.com right now.


amadeus (approved):
🙏

PR #735, opened 2026-05-22
Add docs manifests for configuring web app versions of our sites

mdo:
Testing out a little manifest generation to configure the icon, style, etc for web app versions of our sites. Most useful for DiffsHub given it's utility, but nice for the rest, too.

Extra motivation here is that Next.js automatically picks up root `favicon.ico` and automatically injects that everywhere, and that somehow takes priority every time. This is why we had some wonky icons across Diffs, Trees, and DiffsHub.


mdo:
@codex review

PR #737, opened 2026-05-23
[diffs] WorkerPool & Element Pool Hardening

amadeus:
Basically the theme switcher branch for diffshub has exposed a ton of flaws in my WorkerPoolManager and element pooling architecture that are essentially fixed here.

The fixes are grouped into 3 commits that address a few important scenarios.

### Clean `renderCache` on worker pool theme change
When the worker pool has a theme change, we need to blow away our existing render cache to ensure we never render from a stale theme. Additionally this helps the theme change feel instant/responsive because we can render plain text right away and not wait for a response from the worker.

### `WorkPoolManager.setRenderOptions` race conditions
The core change here was that given the async nature of loading/setting themes (both to load the theme and update the workers), we needed better validation under rapid theme changes to ensure that tasks are properly dropped and if things come back out of order they can be properly ignored.

### Fixing poisoned elements
While we were clearing the element pool correctly on theme changes, there was another bug wherein if we were recycling on that tick we were then pooling invalid elements. So this added a sort of version tracker for the element pool that gets incremented and tied to versioned renders to ensure that pooled element contract is always correct with its children

### DIFF
--- a/packages/diffs/src/components/CodeView.ts
+++ b/packages/diffs/src/components/CodeView.ts
@@ -464,8 +464,6 @@ const SCROLL_REBASE_TARGET_BOTTOM =
   SCROLL_REBASE_CONTAINER_HEIGHT - SCROLL_REBASE_TARGET_TOP;
 const SCROLL_REBASE_THRESHOLD =
   SCROLL_REBASE_CONTAINER_HEIGHT - SCROLL_REBASE_TRIGGER_TOP;
-const CODE_VIEW_ELEMENT_POOL_LIMIT = 32;
-
 interface ScrollToAnimation {
   position: number;
   velocity: number;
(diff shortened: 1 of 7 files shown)


packages/diffs/src/components/CodeView.ts
```diff
@@ -907,9 +907,30 @@ export class CodeView<LAnnotation = undefined> {
     item.instance.primeHighlightCache();
   }
 
+  private getElementPoolLimit() {
```
amadeus, line 910: Another smol optimization that went into here -- in fully collapsed diff views, i was noticing that we were blowing our max pool windows constantly, so I reworked the max size to better fit the max collapsed items we might render. Also made it larger for `isContainerManaged` which actually means elements need an extra tick to be cleaned up before they can be re-cycled.

PR #741, opened 2026-05-24
[diffs] Remove necessary set shorthands and defaults

Que-tin:
### Description

This is the first of a couple isolated PRs to improve the stylings of diffs.

It removes necessary shorthands that under the hood set or reset way too many properties that aren't even used and also removes a couple of stylings that I stumbled upon that are user agent defaults across browsers anyways. 

I'll comment each individual change in the PR, check there for more info

### Type of changes

- [x] Refactoring (non-breaking change)
      being opened
      that this PR is being opened
- [ ] Documentation update

### Checklist

- [x] I have read the

Not at all.

### DIFF
--- a/packages/diffs/src/style.css
+++ b/packages/diffs/src/style.css
@@ -222,7 +222,7 @@
     margin: 0;
     padding: 0;
     display: block;
-    outline: none;
+    outline-style: none;
     font-family: var(--diffs-font-family, var(--diffs-font-fallback));
   }
 
@@ -811,7 +811,7 @@
   }
 
   [data-code]::-webkit-scrollbar-track {
-    background: transparent;
+    background-color: transparent;
   }
 
   [data-code]::-webkit-scrollbar-thumb {
@@ -1048,7 +1048,6 @@
   [data-expand-button],
   [data-separator-content] {
     display: flex;
-    flex: 0 0 auto;
     align-items: center;
     background-color: var(--diffs-bg-separator);
   }
@@ -1081,7 +1080,7 @@
   }
 
   [data-separator-content] {
-    flex: 1 1 auto;
+    flex-grow: 1;
     padding: 0 1ch;
     height: 100%;
     color: var(--diffs-fg-number);
@@ -1105,7 +1104,6 @@
     min-width: 0;
     text-overflow: ellipsis;
     white-space: nowrap;
-    flex: 0 1 auto;
   }
 
   @supports (width: 1cqi) {
@@ -1525,7 +1523,7 @@
   [data-merge-conflict-actions-content] {
     display: flex;
     align-items: center;
-    gap: 0.25rem;
+    column-gap: 0.25rem;
     padding-inline: 0.5rem;
     min-height: 1.75rem;
     font-family: var(
@@ -1539,8 +1537,8 @@
 
   [data-merge-conflict-action] {
     appearance: none;
-    border: 0;
-    background: transparent;
+    border-style: none;
+    background-color: transparent;
     color: var(--diffs-fg-number);
     font: inherit;
     font-style: normal;
@@ -1572,10 +1570,9 @@
     position: relative;
     background-color: var(--diffs-bg);
     display: flex;
-    flex-direction: row;
     justify-content: space-between;
     align-items: center;
-    gap: var(--diffs-gap-inline, var(--diffs-gap-fallback));
+    column-gap: var(--diffs-gap-inline, var(--diffs-gap-fallback));
     min-height: calc(
       1lh + (var(--diffs-gap-block, var(--diffs-gap-fallback)) * 3)
     );
@@ -1586,9 +1583,8 @@
 
   [data-header-content] {
     display: flex;
-    flex-direction: row;
     align-items: center;
-    gap: var(--diffs-gap-inline, var(--diffs-gap-fallback));
+    column-gap: var(--diffs-gap-inline, var(--diffs-gap-fallback));
     min-width: 0;
     white-space: nowrap;
   }
@@ -1615,7 +1611,7 @@
   [data-diffs-header='default'] [data-metadata] {
     display: flex;
     align-items: center;
-    gap: 1ch;
+    column-gap: 1ch;
     white-space: nowrap;
   }
 
@@ -1696,7 +1692,7 @@
     display: flex;
     align-items: center;
     justify-content: center;
-    border: none;
+    border-style: none;
     appearance: none;
     width: 1lh;
     height: 1lh;


packages/diffs/src/style.css
```diff
@@ -222,7 +222,7 @@
     margin: 0;
     padding: 0;
     display: block;
-    outline: none;
+    outline-style: none;
```
Que-tin, line 225: Literally the same, without other properties being set under the hood

packages/diffs/src/style.css
```diff
@@ -811,7 +811,7 @@
   }
 
   [data-code]::-webkit-scrollbar-track {
-    background: transparent;
+    background-color: transparent;
```
Que-tin, line 814: Literally the same, without other properties being set under the hood.

Also already commonly used across the file

packages/diffs/src/style.css
```diff
@@ -1048,7 +1048,6 @@
   [data-expand-button],
   [data-separator-content] {
     display: flex;
-    flex: 0 0 auto;
```
Que-tin, line 1051: This was unnecessary.

flex-grow: 0 is the default.
Same goes with flex-basis: auto.

[data-expand-button] was already setting flex-shrink: 0 individually

[data-separator-content] was overriding the whole thing anyway (see next comment)

packages/diffs/src/style.css
```diff
@@ -1081,7 +1080,7 @@
   }
 
   [data-separator-content] {
-    flex: 1 1 auto;
+    flex-grow: 1;
```
Que-tin, line 1083: Same as above

packages/diffs/src/style.css
```diff
@@ -1105,7 +1104,6 @@
     min-width: 0;
     text-overflow: ellipsis;
     white-space: nowrap;
-    flex: 0 1 auto;
```
Que-tin, line 1108: These are basically the UA-defaults

packages/diffs/src/style.css
```diff
@@ -1525,7 +1523,7 @@
   [data-merge-conflict-actions-content] {
     display: flex;
     align-items: center;
-    gap: 0.25rem;
+    column-gap: 0.25rem;
```
Que-tin, line 1526: We have a default flexbox no need to set row-gap as well

packages/diffs/src/style.css
```diff
@@ -1572,10 +1570,9 @@
     position: relative;
     background-color: var(--diffs-bg);
     display: flex;
-    flex-direction: row;
     justify-content: space-between;
     align-items: center;
-    gap: var(--diffs-gap-inline, var(--diffs-gap-fallback));
+    column-gap: var(--diffs-gap-inline, var(--diffs-gap-fallback));
```
Que-tin, line 1575: We have a default flexbox no need to set row-gap as well

packages/diffs/src/style.css
```diff
@@ -1586,9 +1583,8 @@
 
   [data-header-content] {
     display: flex;
-    flex-direction: row;
     align-items: center;
-    gap: var(--diffs-gap-inline, var(--diffs-gap-fallback));
+    column-gap: var(--diffs-gap-inline, var(--diffs-gap-fallback));
```
Que-tin, line 1587: We have a default flexbox no need to set row-gap as well

packages/diffs/src/style.css
```diff
@@ -1615,7 +1611,7 @@
   [data-diffs-header='default'] [data-metadata] {
     display: flex;
     align-items: center;
-    gap: 1ch;
+    column-gap: 1ch;
```
Que-tin, line 1614: We have a default flexbox no need to set row-gap as well

packages/diffs/src/style.css
```diff
@@ -1696,7 +1692,7 @@
     display: flex;
     align-items: center;
     justify-content: center;
-    border: none;
+    border-style: none;
```
Que-tin, line 1695: Same as with outline -> outline-style

packages/diffs/src/style.css
```diff
@@ -1586,9 +1583,8 @@
 
   [data-header-content] {
     display: flex;
-    flex-direction: row;
```
Que-tin, line 1589: UA-Default

packages/diffs/src/style.css
```diff
@@ -1572,10 +1570,9 @@
     position: relative;
     background-color: var(--diffs-bg);
     display: flex;
-    flex-direction: row;
```
Que-tin, line 1575: UA-Default

packages/diffs/src/style.css
```diff
@@ -1539,8 +1537,8 @@
 
   [data-merge-conflict-action] {
     appearance: none;
-    border: 0;
-    background: transparent;
+    border-style: none;
+    background-color: transparent;
```
Que-tin, line 1541: Same as at the start

packages/diffs/src/style.css
amadeus: I know this probably isn't needed, but i am a fan of seeing the direction listed out here to help establish context.

packages/diffs/src/style.css
amadeus: Similar note above, I personally prefer the context here to help establish what's happening with layout.

packages/diffs/src/style.css
amadeus: I personally prefer to see these types of flex values in context because you don't have to remember ua defaults and you can see what is being defined explicitly.

amadeus (commented):
Most of these I'm good with, but I do have a bit of a stylistic preference in being more explicit around flex properties. So ideally we keep those but the rest look good to me. Given this PR was opened a while ago (sorry for the delays, a lot going on), you might need a rebase to catch up and it might involve some additional changes.

Que-tin:
> Most of these I'm good with, but I do have a bit of a stylistic preference in being more explicit around flex properties. So ideally we keep those but the rest look good to me. Given this PR was opened a while ago (sorry for the delays, a lot going on), you might need a rebase to catch up and it might involve some additional changes.


Will look into it as soon as I'm back from conference. 👍🏻

packages/diffs/src/style.css
Que-tin: Interesting. Will revert the changes then.

Out of curiosity where do you draw the line? 

I mean you also don't define display: block or inline on each element or the font-sizes, weights and family. What part of the cascade or UAs is "worth" ignoring and which part is worth redefining? 

Really interested to get some insights in the thought process behind this, never stumbled across such pattern. 😊

amadeus: There’s no hard and fast rule I have here, some of it stems from my experience working in react native where the default direction is column. There’s likely inconsistencies in this file already to begin with, but if intentional changes are happening I would prefer explicitness over ambiguity.

PR #731, opened 2026-05-21
[diffshub] Add Shiki theme switcher

necolas:
## Summary

  Adds a light/dark Shiki theme picker to the diffshub header and wires it through
  the diff viewer, the file tree, and the surrounding sidebar chrome so a single
  selection re-themes everything in view.

  - New header control with Light/Dark menus + Auto/Light/Dark color-mode toggle.
    Choices persist to localStorage via a small `usePersistedState` hook with an
    allowlist guard so stale keys can't push the picker into an unknown theme.
  - The selected Shiki theme drives the `trees` sidebar palette (file rows, git
    decorations, hover/selection chrome) and the diffshub sidebar wrapper
    (tabs row, Diff Stats panel, "Powered by" footer) — all via a shared
    `useResolvedTreeTheme` hook with a module-level cache so re-selecting is
    synchronous.
  - Sidebar primary text is sourced from `sideBar.foreground` (falling back to
    `editor.foreground`); muted text prefers `descriptionForeground`. Avoids the
    slack-ochin case where the editor surface is white but the sidebar is dark
    navy, which previously wrote `#000` over `#2D3E4C`.
  - `themeToTreeStyles` now rejects `list.hoverBackground` when its luminance
    sits closer to the sidebar fg than the sidebar bg (again, slack-ochin: a
    near-white hover wash over a dark navy sidebar erased row text). The rgba
    hover fallback is also picked from the sidebar's actual luminance instead
    of `theme.type`. Covered by focused unit tests.
  - Two `diffs` caches were keeping the previous theme's `:host` variables alive
    across theme swaps: `DiffHunksRenderer`/`FileRenderer`'s per-instance
    `renderCache.result` (collapsed files never re-tokenize, so their headers
    and indicator colors stayed stale) and `FileDiff.applyThemeState`'s
    `hasAdoptedThemeCSS` short-circuit on pool reuse. Both now pull fresh
    `themeStyles` from `WorkerPoolManager.getCurrentThemeStyles()` /
    always `upsertHostThemeStyle` once on adoption.

  ## Test plan

  - [x] Switch between several Shiki themes (including slack-ochin and a
        light theme) and confirm the diff viewer, file tree, and sidebar
        chrome all re-theme together
  - [x] Collapse a file, swap themes, confirm the collapsed header repaints
        without expanding the file
  - [x] Scroll past the virtualization window to recycle file containers,
        then swap themes, confirm recycled containers pick up the new theme
  - [x] Reload the page and confirm the previous light/dark picks rehydrate

PR #745, opened 2026-05-26
[diffshub] Add a Shiki theme switcher

necolas:
Clean history version of #731

## Summary

Adds a theme switcher to the diffshub code view so reviewers can render diffs
and the surrounding chrome in any Shiki theme, independent of the app's
global light/dark palette.

- **Theme picker** in the header (next to the gear): a light-theme list, a
  dark-theme list, and an Auto/Light/Dark color-mode toggle, all in one
  in-place dropdown that fits narrow viewports. Picks persist across reloads.
- **Themed chrome**: the sidebar, file tree, header, comment cards, and inline
  annotations all source their colors from the resolved Shiki theme. Foreground
  and muted-foreground are chosen by contrast against the resolved surface
  (not the theme's declared values) so text stays legible on mixed-palette
  themes (e.g. slack-ochin, ayu-dark).
- **System Monitor theme-cycle button** sweeps through every theme for quick
  previewing.
- Diff and chrome repaints are kept in the same frame on a theme swap, gated on
  theme hydration so client-side navigation doesn't tokenize against the
  default palette.
- **[trees]** `themeToTreeStyles` now rejects a theme's `list.hoverBackground`
  when its luminance would erase the hovered row's text, and derives the rgba
  hover fallback from the sidebar's actual luminance rather than the declared
  theme type.

## Commits

Organized into 7 focused commits (foundations first, then UI):

1. `[trees]` Reject `list.hoverBackground` when it would erase row text
2. `[diffshub]` Add a localStorage-backed persisted state hook
3. `[diffshub]` Add the Shiki theme catalog
4. `[diffshub]` Resolve Shiki themes into themed chrome tokens
5. `[diffshub]` Add a theme switcher to the code view header
6. `[diffshub]` Apply the resolved theme across the chrome
7. `[diffshub]` Add a theme-cycle button to the System Monitor

## Test plan

- [x] Pick light/dark themes from the header; confirm diff + chrome update together
- [x] Reload and confirm picks persist; navigate away and back, confirm no
      wrong-palette flash
- [x] Try mixed-palette themes (slack-ochin, ayu-dark) and confirm sidebar
      labels / comment labels stay legible
- [x] Run the System Monitor theme-cycle button, scroll quickly up and down to ensure theme is correctly applied to collapsed and open files

### DIFF
--- a/apps/docs/app/(diffshub)/(view)/_components/CodeViewHeader.tsx
+++ b/apps/docs/app/(diffshub)/(view)/_components/CodeViewHeader.tsx
@@ -389,7 +389,7 @@ function ThemeDropdown({
         {view === 'main' ? (
           <>
             <DropdownMenuItem
-              className="cursor-default p-1 focus:bg-transparent"
+              className="cursor-default p-0 focus:bg-transparent"
               onSelect={(event) => event.preventDefault()}
             >
               <ButtonGroup
--- a/apps/docs/app/(diffshub)/(view)/_components/CodeViewHeader.tsx
+++ b/apps/docs/app/(diffshub)/(view)/_components/CodeViewHeader.tsx
@@ -1,7 +1,12 @@
 import type { DiffIndicators } from '@pierre/diffs';
 import {
+  IconCheck,
+  IconChevronSm,
   IconCodeStyleBars,
   IconCollapsedRow,
+  IconColorAuto,
+  IconColorDark,
+  IconColorLight,
   IconDiffSplit,
   IconDiffUnified,
   IconExpandAll,
@@ -12,10 +17,27 @@ import {
   IconSymbolDiffstat,
 } from '@pierre/icons';
 import Link from 'next/link';
-import { type Dispatch, memo, type SetStateAction, useState } from 'react';
+import {
+  type CSSProperties,
+  type Dispatch,
+  memo,
+  type SetStateAction,
+  useLayoutEffect,
+  useMemo,
+  useRef,
+  useState,
+} from 'react';
 
 import { DiffUrlForm } from '../../_components/DiffUrlForm';
 import { DiffsHubLogo } from './DiffsHubLogo';
+import {
+  type ColorMode,
+  DARK_THEMES,
+  type DarkTheme,
+  LIGHT_THEMES,
+  type LightTheme,
+} from './themes';
+import { useThemeChromeStyle } from './useResolvedTreeThemeStyles';
 import { Button } from '@/components/ui/button';
 import { ButtonGroup, ButtonGroupItem } from '@/components/ui/button-group';
 import {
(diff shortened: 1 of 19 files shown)


mdo:
CleanShot 2026-05-26 at 14 49 28@2x

Looks like maybe our menus/overlays are a little janky with borders/shadows now?

CleanShot 2026-05-26 at 14 50 15@2x

Some border color issue on collapsed files?

necolas:
Something more like this?

image

amadeus:
Yeah that looks good! i wonder if we can fix the scrollbar to to inherit the themed colors? maybe try to inherit the annotation colors or something, but fall back to the defaults if they don't exist?

apps/docs/app/(diffshub)/(view)/_components/CodeViewHeader.tsx
```diff
@@ -245,3 +308,262 @@ export const CodeViewHeader = memo(function CodeViewHeader({
...
+}
+
+// Theme picker shown next to the gear icon. Avoids horizontal sub-menus
+// (which overflow on narrow viewports) by re-using the same DropdownMenu
+// content for three "views" — main, light theme list, dark theme list —
+// switched via local state. The user enters a list by clicking the
+// corresponding row, picks a theme, and is returned to the main view. The
+// menu auto-resets to the main view whenever it closes so the next open
+// always starts from the top.
+function ThemeDropdown({
```
amadeus, line 351: Should we add some sort of `reset` button to this if themes are not the default to allow quick reversion back to the default themes?

apps/docs/app/(diffshub)/(view)/_components/useThemeCycle.ts
```diff
@@ -0,0 +1,132 @@
...
+      if (step.mode === 'light') {
+        setLightTheme(step.theme);
+        setColorMode('light');
+      } else {
+        setDarkTheme(step.theme);
+        setColorMode('dark');
+      }
+      idx++;
+    };
+    const intervalId = window.setInterval(tick, stepSeconds * 1000);
```
amadeus, line 128: Should we call tick before setting this interval up so the use sees an immediate change? I definitely noticed my self clicking on the 3second and then clicking again because i didn't realize it was running or not

amadeus:
CleanShot 2026-05-27 at 12 38 07

one thing that kinda sucks, a ton of our components have built in CSS transitions, and this themeing cycle really kinda fucks that up. most of this i think is due to bullshit radix stuff, but i wonder if there's a way we can disable that on diffshub

apps/docs/app/(diffshub)/(view)/_components/themes.ts
```diff
@@ -0,0 +1,80 @@
...
+  'rose-pine',
+  'rose-pine-moon',
+  'slack-dark',
+  'solarized-dark',
+  'synthwave-84',
+  'tokyo-night',
+  'vesper',
+  'vitesse-black',
+  'vitesse-dark',
+] as const;
```
amadeus, line 73: Should probably figure out a way to directly export this from diffs at some point (mostly just a mental note for me)

amadeus (approved):
Code looks good to me! Only a few UX notes around reset/tick improvements, and then some of the random comments from the thread on design stuff! 🙏

apps/docs/app/(diffshub)/(view)/_components/themes.ts
necolas: We should think through making themes something generalized that can be plugged into diffs, trees, editor and used in app chrome that stitches them all together.

apps/docs/app/(diffshub)/(view)/_components/CodeViewHeader.tsx
necolas: Maybe in a later patch

apps/docs/app/(diffshub)/(view)/_components/useThemeCycle.ts
necolas: Done and updated

necolas:
@mdo I made some changes to how the border colors are generated for the files and app chrome. Looked through all the themes and think they're improved.

apps/docs/app/(diffshub)/(view)/_components/CodeViewHeader.tsx
necolas: Added in https://github.com/pierrecomputer/pierre/commit/30e5a8f91086fb6f2d3288a885347dc296c590be

necolas:
> i wonder if we can fix the scrollbar to to inherit the themed colors? 

Added in https://github.com/pierrecomputer/pierre/commit/930e8ca13d84715a034ea555b9d76c90a804c382

PR #747, opened 2026-05-28
[diffshub] Themify chrome: loading panel, scrollbar, and a reset link

necolas:
Follows on from the recent theme-switcher work by extending the active Shiki theme to two surfaces that were still pinned to the global light/dark palette, and adds a quick way back to the defaults.

### Theme the loading status panel

The "Fetching diff" / "Preparing diff" / error panel rendered on the global background, so under any non-default theme the loading screen flashed a mismatched surface before the diff came up. It now applies `useThemeChromeStyle` so its background and text follow the active theme, exactly like the header and sidebar. The hook's module-level cache keeps the resolved theme available through the first render, so there's no one-frame flash.

### Theme the main code-view scrollbar

`cv-scrollbar` derived its thumb from `var(--background)` and left its track transparent, but the `CodeView` host never gets `--background` re-themed — so under a non-default theme you'd get a near-black thumb and a dark gutter band. (The dropdown mini-scrollbars avoid this for free because their surface re-themes `--background`.)

Two new chrome variables fix it:

* `--diffshub-scrollbar-thumb-bg` — uses the default rule's own lighten/darken formula on the editor surface, so the thumb matches the theme-list mini-scrollbars.
* `--diffshub-scrollbar-track-bg` — the editor surface, so the reserved gutter paints with the diff instead of the host background.

`globals.css` keeps the prior formulas as fallbacks, so default chrome and any non-themed `cv-scrollbar` consumer are unchanged. The track border also reuses the themed diff separator so the gutter edge reads as one hairline with the file boundaries.

### Add a "Reset to default themes" link

After wandering through the Shiki catalog there was no quick way back to the pierre-soft defaults short of hunting for them in the light and dark lists. A muted link at the bottom of the theme picker's main view restores both slots at once. It only appears when at least one slot has drifted from the default pair, so it stays out of the way until it's useful.

### DIFF
--- a/apps/docs/app/(diffshub)/(view)/_components/CodeViewHeader.tsx
+++ b/apps/docs/app/(diffshub)/(view)/_components/CodeViewHeader.tsx
@@ -34,6 +34,8 @@ import {
   type ColorMode,
   DARK_THEMES,
   type DarkTheme,
+  DEFAULT_DARK_THEME,
+  DEFAULT_LIGHT_THEME,
   LIGHT_THEMES,
   type LightTheme,
 } from './themes';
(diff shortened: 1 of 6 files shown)


amadeus (approved):
Siiick 🙏

PR #750, opened 2026-05-29
[diffshub] Sync the iOS navbar tint to the theme

necolas:
iOS Safari tints its toolbar from <meta name="theme-color">, which the app never set — so the navbar didn't track the selected theme. Add a JS-managed theme-color meta in two layers:

- Mode-level base, sitewide: the pre-paint bootstrap script and ThemeProvider.applyTheme set the meta to the resolved mode's background (#ffffff / #0a0a0a, matching the global body bg iOS samples). Correct on first paint, updates on mode change. Applies to all three sites (diffs/trees/diffshub) since it lives in shared layout/provider code.
- Shiki upgrade, diffshub viewer only: ReviewUI overwrites the meta with the resolved editor background (e.g. synthwave's #262335) so the navbar matches the diff surface, and restores the mode-level color on unmount so other pages don't inherit the viewer's tint.

The meta is created imperatively — by the bootstrap script, then ThemeProvider/the viewer — rather than authored in JSX, so a single element is owned end-to-end by the same code that updates it. Keeping creation and mutation in one place (instead of split between React-managed markup and DOM writes) makes the tag's value predictable across first paint, mode changes, and route transitions.

### DIFF
--- a/apps/docs/app/layout.tsx
+++ b/apps/docs/app/layout.tsx
@@ -174,6 +174,22 @@ const themeBootstrapScript = `(${String(function applyInitialTheme() {
     root.classList.remove('light', 'dark');
     root.classList.add(resolvedTheme);
     root.style.colorScheme = resolvedTheme;
+
+    // Set the iOS navbar tint before first paint so it matches the resolved
+    // mode immediately. The meta is created here (not authored in JSX, which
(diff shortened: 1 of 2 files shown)


amadeus (approved):
👯

PR #757, opened 2026-06-01
feat(editor): Add rounded selection

ije:
Screenshot 2026-06-01 at 08 54 36

↓

image

Can be disabled by adding:
```js
import { Editor } from "@pierre/diffs/editor"
const editor = new Editor({ roundedSelection: false })
```

### DIFF
--- a/packages/diffs/src/editor/editor.ts
+++ b/packages/diffs/src/editor/editor.ts
@@ -1633,12 +1633,14 @@ export class Editor<LAnnotation> implements DiffsEditor<LAnnotation> {
     left: number,
     applyEolSpacing = true
   ) {
-    const spacing =
-      !applyEolSpacing ||
-      selection.end.line === ln ||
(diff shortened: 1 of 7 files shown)


packages/diffs/src/editor/editor.ts
```diff
@@ -648,12 +653,12 @@ export class Editor<LAnnotation> implements DiffsEditor<LAnnotation> {
               ) {
                 return;
               }
-              const lineIndex = Number(lineNumber) - 1;
+              const line = Number(lineNumber) - 1;
```
amadeus: Would it be worth doing `parseInt(lineNumber, 10)` on this and error/silently fail if `isNaN(line)`?

amadeus (approved):
lookin sharp!

packages/diffs/src/editor/editor.ts
ije: fixed

PR #759, opened 2026-06-01
refactor(editor): refactor search panel

ije:
- Refactor`SearchPanelWidget` class
- Render match highlight for search
Screenshot 2026-06-02 at 00 42 41

### DIFF
--- a/packages/diffs/src/editor/css.ts
+++ b/packages/diffs/src/editor/css.ts
@@ -7,9 +7,6 @@ export const editorCSS: string = /* CSS */ `
     50% { opacity: 0; }
     100% { opacity: 1; }
   }
-  [data-editor-overlay] {
-    display: contents;
-  }
   :host, /* for scroll anchor */
   [data-code], /* for editor overlay */
   [data-content] /* for wrap line measurement */
@@ -48,6 +45,10 @@ export const editorCSS: string = /* CSS */ `
     background-color: var(--diffs-line-bg);
     -webkit-user-select: none;
     user-select: none;
+  }
+
+  [data-editor-overlay] {
+    display: contents;
   }
   [data-caret], [data-selection-range], [data-match-range] {
     position: absolute;
--- a/packages/diffs/src/editor/css.ts
+++ b/packages/diffs/src/editor/css.ts
@@ -76,6 +76,23 @@ export const editorCSS: string = /* CSS */ `
     z-index: -10;
     background-color: var(--diffs-editor-selection-bg);
   }
+  [data-selection-corner] {
+    width: 100%;
+    height: 100%;
+    background-color: var(--diffs-bg);
+  }
+  [data-rtl]  {
+    border-top-left-radius: 3px;
+  }
+  [data-rtr] {
+    border-top-right-radius: 3px;
+  }
+  [data-rbl] {
+    border-bottom-left-radius: 3px;
+  }
+  [data-rbr] {
+    border-bottom-right-radius: 3px;
+  }
   [data-editor-overlay] {
     display: contents;
   }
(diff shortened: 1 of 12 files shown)


packages/diffs/src/editor/css.ts
```diff
@@ -7,9 +7,12 @@ export const editorCSS: string = /* CSS */ `
     50% { opacity: 0; }
     100% { opacity: 1; }
   }
```
amadeus, line 9: Not for this PR, but I would love to get this file moved into a normal css file so we can run it through our various linters and stuff. I think it would also be nice to figure out a way to figure out how to make it a re-usable stylesheet like we do here:

https://github.com/pierrecomputer/pierre/blob/main/packages/diffs/src/components/web-components.ts

packages/diffs/src/editor/editor.ts
```diff
@@ -764,15 +766,23 @@ export class Editor<LAnnotation> implements DiffsEditor<LAnnotation> {
...
-      this.#scrollToLine(this.#scrollingToLine, this.#scrollingToLineChar);
+      this.#scrollToLine(
+        this.#scrollingToLine,
+        this.#scrollingToLineChar,
+        this.#scrollingToLineNoFocus
+      );
       this.#scrollingToLine = undefined;
       this.#scrollingToLineChar = undefined;
-    } else if (this.#selections !== undefined && this.#selections.length > 0) {
+      this.#scrollingToLineNoFocus = false;
```
amadeus, line 790: How will this scrollToLine stuff interface with CodeView implementations?

packages/diffs/src/editor/searchPanel.ts
```diff
@@ -21,93 +16,138 @@ export interface SearchPanelOptions {
...
-    </svg>
-    `,
+        <line x1="3" y1="6" x2="10" y2="6" fill="none" stroke="currentColor" stroke-linecap="round" stroke-linejoin="round" stroke-width="2"></line>
+        <circle cx="12.5" cy="6" r="2.5" fill="none" stroke="currentColor" stroke-linecap="round" stroke-linejoin="round" stroke-width="2"></circle>
+        <line x1="15" y1="6" x2="17" y2="6" fill="none" stroke="currentColor" stroke-linecap="round" stroke-linejoin="round" stroke-width="2"></line>
+        <line x1="17" y1="14" x2="10" y2="14" fill="none" stroke="currentColor" stroke-linecap="round" stroke-linejoin="round" stroke-width="2"></line>
+        <circle cx="7.5" cy="14" r="2.5" fill="none" stroke="currentColor" stroke-linecap="round" stroke-linejoin="round" stroke-width="2"></circle>
+        <line x1="5" y1="14" x2="3" y2="14" fill="none" stroke="currentColor" stroke-linecap="round" stroke-linejoin="round" stroke-width="2"></line>
+      </svg>
+      `,
```
amadeus, line 154: We should make this icon part of our SVG sprite sheet, cc @mdo

amadeus (approved):
My review here is mostly high level, definitely requires a lot of additional context that I don't have around how the editor works. But in general looks good.

Most of my notes are around future stuff.

Also i guess side question, does editor work with CodeView? Reason I ask is because there's potentially some integration stuff we might need to talk about

packages/diffs/src/editor/editor.ts
ije: I think they are totally different. The `scrollToLine` of editor is used in a diff/file component when `editable`, CodeView scroll api works multiple diff/file components

packages/diffs/src/editor/searchPanel.ts
ije: i didn't put these svg icons in the sprite to keep the diffs index module smaller. the user may not need the editing feature. maybe an standalone editor svg sprite?

PR #763, opened 2026-06-01
[diffs] Ensure proper bun version when running publish

amadeus:
This should hopefully fix the derp from earlier today/this weekend.


amadeus:
@SlexAxton unclear if we should do this with more packages? I just added to diffs for now

SlexAxton (approved):

PR #766, opened 2026-06-02
pref(editor): Introduce `postponeBackgroundTokenizeToNextFrame` method for the `DiffsEditor`

ije:
when doing re-render(selection/code) postpone the background tokenizing to next frame. this can avoid UI freezing for a large file editing.


packages/diffs/src/components/File.ts
```diff
@@ -519,12 +522,20 @@ export class File<
...
         'File.render: attempting to call render after cleaned up'
       );
     }
+
+    const editor = this.editor;
+    if (editor != null) {
+      // postpone background tokenizing to next frame for avoiding UI freeze
+      // during render
+      editor.postponeBackgroundTokenizeToNextFrame();
+    }
```
amadeus: Mostly a nit, but could this just be simplified to a single line without an if statement?

```ts
this.editor?.postponeBackgroundTokenizeToNextFrame()
```

packages/diffs/src/components/FileDiff.ts
```diff
@@ -752,6 +752,14 @@ export class FileDiff<
         'FileDiff.render: attempting to call render after cleaned up'
       );
     }
+
+    const editor = this.editor;
+    if (editor != null) {
+      // postpone background tokenizing to next frame for avoiding UI freeze
+      // during render
+      editor.postponeBackgroundTokenizeToNextFrame();
+    }
```
amadeus: Mostly a nit, but could this just be simplified to a single line without an if statement?

```ts
this.editor?.postponeBackgroundTokenizeToNextFrame()
```

packages/diffs/src/editor/editor.ts
```diff
@@ -338,14 +246,17 @@ export class Editor<LAnnotation> implements DiffsEditor<LAnnotation> {
       return;
     }
     const contentEl = codeElement.children[1] as HTMLElement | undefined;
-    if (contentEl === undefined) {
+    if (contentEl === undefined || contentEl.dataset.content === undefined) {
```
amadeus, line 249: Question around code style, in the rest of diffs, I typically do: `== null` or `!= null` to test `undefined`-ness. I only use `=== undefined` if I need to differentiate between `null` or `undefined`. Not that this has to change, but mostly just a style convention incase you are curious.

packages/diffs/src/editor/editor.ts
```diff
@@ -818,6 +428,108 @@ export class Editor<LAnnotation> implements DiffsEditor<LAnnotation> {
     }
   }
 
+  postponeBackgroundTokenizeToNextFrame(): void {
+    const tokenizer = this.#tokenizer;
+    if (tokenizer !== undefined) {
+      tokenizer.pauseBackgroundTokenize();
+      requestAnimationFrame(() => {
```
amadeus, line 433: I have some utilities in diffs to manage request animationFrame to ensure things are all tied to a single callback `queueRender` and `dequeueRender` ( https://github.com/pierrecomputer/pierre/blob/editor/perf/tokenzier/packages/diffs/src/managers/UniversalRenderingManager.ts i should fix the names, they suck, but it is what it is).

Ideally though you'd create a function on the class itself, so you're always passing the same function in, that way you don't have to worry about double firing things, etc.

And then also cleanup can properly clean up these callbacks to not potentially have them firing after an unmount or whatever

packages/diffs/src/editor/editor.ts
```diff
@@ -981,6 +693,309 @@ export class Editor<LAnnotation> implements DiffsEditor<LAnnotation> {
     ];
   }
 
+  #listenContentElement(contentEl: HTMLElement): void {
+    const guttterEl = contentEl.previousElementSibling as HTMLElement | null;
```
amadeus: should this be `gutterEl` not `guttterEl`?

packages/diffs/src/editor/editor.ts
ije: sure sure! i will follow the code style in another PR.

packages/diffs/src/components/File.ts
ije: ✅

packages/diffs/src/components/FileDiff.ts
ije: ✅

packages/diffs/src/editor/editor.ts
ije: ✅

packages/diffs/src/editor/editor.ts
ije: nice! all(4) requestAnimationFrame callbacks in `editor.ts` are arrow functions, and they are passive to be called. In `resumeBackgroundTokenize` method we also check `isPaused` state, i think it's ok to use `requestAnimationFrame` here?

ije:
@amadeus i just fixed the typo, reduced some unnecessary `requestAnimationFrame` calls. will fix code style in another PR. please check again! 💚

PR #774, opened 2026-06-03
refactor(editor): realtime update on diff editing

ije:
- [x] **File:** Update line-annotation DOM in place on move/delete (no full content rendering)
- [x] **File:** Skip full rerendering when edits only append lines at end of document
- [x] **FileDiff:** Force diff style to 'split'
- [x] **FileDiff:** Use incremental/partial diff&rendering


packages/diffs/src/components/File.ts
```diff
@@ -487,28 +487,27 @@ export class File<
...
       this.file != null
     ) {
-      this.render({
-        file: this.file,
-        renderRange: this.renderRange,
-        lineAnnotations: newLineAnnotations,
-      });
+      this.setLineAnnotations(newLineAnnotations);
+      this.fileRenderer.setLineAnnotations(this.lineAnnotations);
+      this.renderAnnotations();
```
ije, line 503: just re-render annotations now.

packages/diffs/src/components/VirtualizedFile.ts
```diff
@@ -549,15 +549,30 @@ export class VirtualizedFile<
...
+    const { heights, checkpoints } = this.cache;
     const previousRenderRange = this.renderRange;
-    super.applyLayoutChange(textDocument, newLineAnnotations);
+
+    super.applyDocumentChange(textDocument, newLineAnnotations);
+
+    // reset the layout cache
     this.getSimpleVirtualizer()?.markDOMDirty();
-    this.resetLayoutCache(true);
+    this.layoutDirty = true;
```
ije, line 565: `resetLayoutCache` will clear the `renderRange` that causes full render even inserting one single line

packages/diffs/src/components/VirtualizedFileDiff.ts
```diff
@@ -679,6 +681,41 @@ export class VirtualizedFileDiff<
     this.virtualizer.instanceChanged(this, false);
   }
 
+  // Normally triggered by the editor when the document line count changes.
+  override applyDocumentChange(
```
ije, line 685: like `VirtualizedDiff`, we need to reset cache and update buffers after a line-count change

packages/diffs/src/editor/editor.ts
```diff
@@ -211,6 +230,7 @@ export class Editor<LAnnotation> implements DiffsEditor<LAnnotation> {
         enableGutterUtility: false,
         enableLineSelection: false,
         expandUnchanged: true,
+        diffStyle: 'split',
```
ije, line 233: force the diff style as 'split'

packages/diffs/src/editor/editor.ts
```diff
@@ -236,68 +256,30 @@ export class Editor<LAnnotation> implements DiffsEditor<LAnnotation> {
...
-      // normalize the render range
-      if (startingLine !== undefined && endLine !== undefined) {
-        renderRange = {
-          startingLine: startingLine,
-          totalLines: endLine - startingLine,
-          bufferBefore: 0,
-          bufferAfter: 0,
-        };
-      }
-    }
```
ije, line 291: no needed

packages/diffs/src/editor/editor.ts
```diff
@@ -1249,79 +1207,85 @@ export class Editor<LAnnotation> implements DiffsEditor<LAnnotation> {
...
-      gutterEl.style.gridRow = 'span ' + gutterEl.children.length;
-      contentEl.style.gridRow = 'span ' + contentEl.children.length;
+    if (didLineCountChange) {
+      let gridRow = contentEl.children.length;
+      for (const child of contentEl.children) {
+        const { bufferSize } = (child as HTMLElement).dataset;
+        if (bufferSize !== undefined) {
+          gridRow += parseInt(bufferSize) - 1;
+        }
+      }
```
ije, line 1266: use `data-buffer-size` when udpate grid row

packages/diffs/src/editor/editor.ts
```diff
@@ -1569,22 +1532,50 @@ export class Editor<LAnnotation> implements DiffsEditor<LAnnotation> {
...
+          if (
+            lineType !== undefined &&
+            isLineEditable(lineType) &&
+            lineNumber !== undefined
+          ) {
+            yFix = (line - lineNumber) * this.#metrics.lineHeight;
+            break;
+          }
+        }
+      }
```
ije, line 1561: add y fix if the `approximateLineY` is too small

packages/diffs/src/editor/lineAnnotations.ts
```diff
@@ -52,3 +59,205 @@ export function applyDocumentChangeToLineAnnotations<T>(
 
   return changed ? nextLineAnnotations : undefined;
 }
+
+export function renderLineAnnotations<LAnnotation>(
```
ije, line 63: the editor now renders changed annotations manually instead of doing full render

packages/diffs/src/renderers/DiffHunksRenderer.ts
```diff
@@ -242,8 +247,20 @@ export class DiffHunksRenderer<LAnnotation = undefined> {
...
+
   public clearRenderCache(): void {
+    const renderCache = this.renderCache;
     this.renderCache = undefined;
+    if (
+      renderCache != null &&
+      renderCache.isDirty === true &&
+      renderCache.diff.cacheKey != null
+    ) {
+      this.workerManager?.evictDiffFromCache(renderCache.diff.cacheKey);
```
ije, line 264: the render cache is updated by the editor, let's purge it from the worker manager cache

packages/diffs/src/renderers/DiffHunksRenderer.ts
```diff
@@ -308,22 +325,143 @@ export class DiffHunksRenderer<LAnnotation = undefined> {
     }
   }
 
+  public updateRenderCache(
```
ije, line 330: link `FileRenderer`, now the edit tokenizes the lines in background and async update the render cache(additionLines) after an edit

packages/diffs/src/renderers/DiffHunksRenderer.ts
```diff
@@ -1664,10 +1823,55 @@ function withContentProperties(
     properties: {
       ...lineNode.properties,
       ...contentProperties,
+      ...extendProperties,
+    },
+  };
+}
+
+function createPlainAdditionLineElement(
```
ije, line 1841: when the editor is tokenizing lines in background, we fill the changed lines as plain text in the render cache

packages/diffs/src/components/FileDiff.ts
```diff
@@ -992,40 +1002,37 @@ export class FileDiff<
...
+    shouldRerender?: boolean
+  ): void {
+    this.hunksRenderer.updateRenderCache(dirtyLines, themeType);
+    if (shouldRerender === true) {
+      this.render({ didEdit: true, renderRange: this.renderRange });
+    }
+  }
+
+  // normally triggered by the editor when the document line count changes
+  applyDocumentChange(
```
amadeus: this should be prefixed with public?

packages/diffs/src/components/FileDiff.ts
```diff
@@ -1917,6 +1905,69 @@ export class FileDiff<
     }
   }
 
+  // Update split-view `data-line-type` after an edit.
+  private updateLineType(renderRange: RenderRange): void {
+    if (this.options.diffStyle === 'unified') {
```
amadeus, line 1910: Not for the scope of this PR -- but i'm guessing unified would be a real pain to support, eh?

packages/diffs/src/editor/lineAnnotations.ts
```diff
@@ -52,3 +59,205 @@ export function applyDocumentChangeToLineAnnotations<T>(
...
+    return;
+  }
+
+  const deletionsAnnotationElements = createLineAnnotationElements(
+    deletionAnnotations,
+    leftContentElement,
+    leftGutterElement
+  );
+
+  requestAnimationFrame(() => {
```
amadeus, line 124: Kind of an extension of our discussion from earlier. I think all requestAnimationFrame calls should pass through our universal API, that way we can ensure better batching of execution overall from everything diffs related

Not necessarily for this PR, but i think we should make a run where every call to requestAnimationFrame in the editor codebase utilizes the shared utility for this.

packages/diffs/src/renderers/DiffHunksRenderer.ts
amadeus: Maybe a quick comment for this to make the `why` more obvious

packages/diffs/src/renderers/DiffHunksRenderer.ts
```diff
@@ -308,22 +325,143 @@ export class DiffHunksRenderer<LAnnotation = undefined> {
...
+              'data-char': char,
+              style: `--diffs-token-${themeType}:${fg};`,
+            },
+            children: [{ type: 'text', value: text }],
+          };
+        }),
+      };
+    }
+
+    if (changedAdditionLines.length > 0 && !diff.isPartial) {
```
amadeus: should we just throw an error early if `diff.isPartial`?

packages/diffs/src/renderers/DiffHunksRenderer.ts
```diff
@@ -1036,7 +1186,11 @@ export class DiffHunksRenderer<LAnnotation = undefined> {
           if (deletionLine != null) {
             const deletionLineDecorated = withContentProperties(
               deletionLineContent,
-              deletionLineDecoration.contentProperties
+              deletionLineDecoration.contentProperties,
+              {
+                'data-line': deletionLine.lineNumber,
+                'data-line-index': `${deletionLine.unifiedLineIndex},${splitLineIndex}`,
+              }
```
amadeus: Can you explain a bit about the necessity of the changes in this function?

packages/diffs/src/utils/updateDiffHunks.ts
```diff
@@ -0,0 +1,324 @@
...
+  parseDiffOptions?: CreatePatchOptionsNonabortable
+): HunkMetadataUpdate {
+  if (diff.isPartial) {
+    return applyHunkUpdateResult(
+      diff,
+      recomputeDiffHunks(diff, parseDiffOptions)
+    );
+  }
+
+  if (diff.deletionLines.length !== diff.additionLines.length) {
```
amadeus, line 63: This is probably always going to true, which means we'll nearly always fall through here, which is probably the least performant way of doing this, i wonder if there are ways we can be smarter about this. Not really a blocker for now, but i'll continue to ponder it.

ije: your are right, ideally we should keep `additionLines` immutable when editing, we can just use the `TextDocument` model of the editor, it use `pieceTable` algo keeps the original text immutable. but i don't want to introduce too many changes to the PR. let's optimize this in future update.

packages/diffs/src/renderers/DiffHunksRenderer.ts
ije: ✅

packages/diffs/src/editor/lineAnnotations.ts
ije: yea agree!

packages/diffs/src/components/FileDiff.ts
ije: it was supported, but the layout is not stable when line break so i disabled it for now(the editor will force update the diffStyle to 'split'). will re-support it with stable layout in another PR.

packages/diffs/src/components/FileDiff.ts
ije: ✅

packages/diffs/src/renderers/DiffHunksRenderer.ts
ije: it's used to fix `data-line` and `data-line-index` after editing, the `updateRenderCache` function used by the editor can not get right `data-line-index` value without rediff, for example if without these props we got inconsistent selected line: 

Screenshot 2026-06-09 at 11 03 34

ije: however i added a logic to only add these props if the render cache is dirty

packages/diffs/src/components/FileDiff.ts
amadeus: Word yeah, not necessary for now, i think the UI gets really tricky to reason about when limiting new edits vs old

packages/diffs/src/renderers/DiffHunksRenderer.ts
amadeus: Ahh, is this because the AST i generate has these values baked in? I do kinda wonder once we start building out more semantic diff stuff, if we can maybe not make this part of the highlight AST generation, and instead make it pure render time thing. Will think about this more

packages/diffs/src/utils/updateDiffHunks.ts
amadeus: Yeah, don't mean for anything to change right now, most just want to think about how we can incrementally update FileHunkMetadata

amadeus (approved):
lgtm!

PR #777, opened 2026-06-04
Add @pierre/theming: shared package

necolas:
## Summary

Introduces `@pierre/theming`, a dependency-light theming toolkit, and migrates
`@pierre/diffs`, `@pierre/trees`, and the docs apps to draw their colors from it.
A single theme name now resolves to the same surfaces everywhere, instead of each
package re-implementing its own resolution, color derivation, and selection logic.

## What's new: `@pierre/theme-kit`

A standalone package (zero runtime dependencies) that understands the Shiki / VS
Code theme format and owns four concerns:

- **Catalog / collection** (`createThemeCatalog`, `createThemeCollection`) — own
  theme names, metadata, defaults, and lazy `load()` loaders, so apps can list
  and pick themes without eagerly loading theme JSON.
- **Resolver** (`createThemeResolver`) — owns registration and resolved-theme
  caching, with typed errors (`DuplicateThemeError`, `UnregisteredThemeError`,
  `UnresolvedThemeError`).
- **Controller** (`createThemeController`) — framework-agnostic runtime state for
  the selected color mode + theme, including persistence.
- **Color primitives** (`@pierre/theme-kit/color`) — `normalizeThemeColors` plus
  transforms that derive accessible UI colors from a resolved theme.

Subpath exports: `.`, `/color`, `/themes` (Pierre + Shiki collections), `/react`.
Pierre and Shiki are modeled as ordinary catalog producers rather than special
cases in the resolver/controller.

## Adoption

- **diffs** — replaces the bespoke theme-resolution code with the shared resolver
  and derives highlighter styles from theme-kit color primitives.
- **trees** — builds tree styles from theme-kit tokens.
- **docs** — drives theming through a shared theme-kit controller + catalog
  (rebuilt theme provider; shared selection across the worker-pool context, the
  Shiki themes example, and the trees demo client).
- **diffshub prototype** — adds a self-contained `_theming` layer (theme
  source/selection hooks, derived chrome/diff/tree props, themed surface
  wrappers) and removes the superseded `themes`, `usePersistedState`, and
  `useResolvedTreeThemeStyles` helpers.

## Deliberate decisions

- diffs' git-status colors (addition/deletion/modified) intentionally stay on the
  diffs-local 2-link lookup rather than adopting theming's longer gutter chain,
  keeping the emitted CSS byte-identical to the pre-theming output. Adopting the
  longer chain is a separate, independently verified follow-up (documented inline
  in `getHighlighterThemeStyles.ts`).


apps/docs/app/(diffs)/_components/WorkerPoolContext.tsx
```diff
@@ -46,13 +46,11 @@ const PoolOptions: WorkerPoolOptions = {
...
-
 const HighlighterOptions: WorkerInitializationRenderOptions = {
-  theme:
-    SITE === 'diffshub'
-      ? { dark: 'pierre-dark-soft', light: 'pierre-light-soft' }
-      : DEFAULT_THEMES,
+  // diffshub used to override the default pair with the soft pierre themes;
+  // now that the canonical default IS the non-soft pair (shared via theme-kit),
+  // every site initializes the pool with the same defaults.
+  theme: DEFAULT_THEMES,
```
amadeus: I'm not sure about this change? Jacob pretty explicitly wanted diffshub to use our soft variants, despite them not being our defaults. The comment just seems to imply that we are making this change but unclear why? Is there a reason it can't stay the way it was?

packages/theming/src/modules/normalizeThemeColors.ts
```diff
@@ -0,0 +1,175 @@
...
+}
+
+// Writes `value` to `key` only when it is a real color, so an absent source key
+// stays absent rather than being coerced to undefined/''.
+function fill(
+  colors: Record<string, string>,
+  key: string,
+  value: string | undefined
+): void {
+  if (value != null && value !== '') colors[key] = value;
```
amadeus, line 159: Could it be worthwhile to call .trim() on value before doing the comparison?

packages/theming/src/modules/createThemeResolver.ts
```diff
@@ -0,0 +1,276 @@
...
+import { type DefaultExport, unwrapDefault } from './unwrapDefault';
+
+// A loader is a zero-argument async factory that returns either a ThemeLike
+// directly or a module-style `{ default: ThemeLike }` (the shape produced by
+// `import()` on a JSON theme file or a re-exporting ESM module).
+export interface ThemeLoader<TTheme extends ThemeLike = ThemeLike> {
+  (): Promise<TTheme | DefaultExport<TTheme>>;
+}
+
+export interface ThemeResolver<TTheme extends ThemeLike = ThemeLike> {
```
amadeus, line 21: Why not make this a class so implementation and annotations can live together?

packages/diffs/src/utils/getHighlighterThemeStyles.ts
```diff
@@ -24,26 +33,36 @@ export function getHighlighterThemeStyles({
...
 
-function getThemeVariables(
+// Emits the diffs git-status CSS variables (addition/deletion/modified colors)
+// for a resolved theme. This intentionally uses the diffs-local 2-link lookup
+// (gitDecoration.* → terminal.ansi*) and STOPS before the editorGutter.* tail
+// that @pierre/theme-kit's normalizeThemeColors adds, so the emitted string stays
+// byte-identical to the pre-theme-kit output. Adopting the gutter fallback for
+// diffs is a deliberate follow-up. A variable is omitted entirely when neither
+// source key is present, matching the previous behavior.
+function getDiffsGitVariables(
```
amadeus: Mostly a nit, but could probs just be `getGitVariables` since we can assume the diffs name already based on where we are in the code? I assume this was probably an AI naming convention which i find sometimes overly verbose

packages/diffs/src/highlighter/themes/themeResolver.ts
```diff
@@ -0,0 +1,17 @@
...
+
+import type { ThemeRegistrationResolved } from '../../types';
+
+// The single diffs-owned theme resolver. It replaces the former hand-rolled
+// ResolvedThemes (sync cache) + ResolvingThemes (in-flight dedupe) +
+// RegisteredCustomThemes (loader registry) trio with one shared
+// @pierre/theme-kit resolver: the resolver owns the cache, the concurrent-load
+// dedupe, and the name→loader registry internally.
+//
+// This module is intentionally NOT re-exported from the package index — the
```
amadeus, line 11: I tend to prefer to just export stuff if possible, why hide this?

docs/superpowers/plans/2026-06-06-theme-catalog-api.md
```diff
@@ -0,0 +1,404 @@
+# Theme Catalog API Implementation Plan
```
amadeus: Should we be committing this plan?

SlexAxton (approved):

apps/docs/app/(diffshub)/(view)/_components/_theming/react/useThemeSelection.ts
```diff
@@ -0,0 +1,91 @@
...
+const NOOP = () => {};
+const EMPTY_THEMES: readonly string[] = [];
+
+// A stable empty state for the fallback controller. useThemeController wraps
+// useSyncExternalStore, which compares snapshot identity — so getState MUST
+// return the same reference on every call or it would loop. Hence a module-level
+// singleton rather than a fresh object literal.
+const FALLBACK_STATE: ThemeControllerState = {
+  darkThemeName: '',
+  lightThemeName: '',
```
amadeus, line 38: Should we allow an empty string here?

apps/docs/app/(diffshub)/(view)/_components/ReviewUI.tsx
```diff
@@ -2,46 +2,41 @@
...
-import { useTheme } from '@/components/theme-provider';
+import {
+  docsThemeCatalog,
+  themeController,
+} from '@/components/themeController';
 
-const LIGHT_THEME_STORAGE_KEY = 'diffshub-light-theme';
-const DARK_THEME_STORAGE_KEY = 'diffshub-dark-theme';
+type LightThemeName = string;
+type DarkThemeName = string;
```
amadeus: Couldn't we use the types from the collections of theme-kit to have proper types for this?

apps/docs/app/(diffshub)/(view)/_components/useThemeCycle.ts
```diff
@@ -1,14 +1,12 @@
...
-  type ColorMode,
-  DARK_THEMES,
-  type DarkTheme,
-  LIGHT_THEMES,
-  type LightTheme,
-} from './themes';
+import { docsThemeCatalog } from '@/components/themeCatalog';
+
+type LightThemeName = string;
+type DarkThemeName = string;
```
amadeus: Similar note from the other file, could we strengthen these types to be the actual supported themes?

packages/diffs/package.json
```diff
@@ -64,6 +64,7 @@
   },
   "dependencies": {
     "@pierre/theme": "catalog:",
+    "@pierre/theme-kit": "workspace:*",
```
amadeus: i'm pretty dumb with this stuff, is `workspace:*` the correct way to define this?

packages/theming/src/modules/createThemeController.ts
```diff
@@ -0,0 +1,501 @@
...
+  name: string;
+}
+
+export interface ThemeResolutionError extends PendingThemeResolution {
+  error: unknown;
+}
+
+export type ThemeResolutionErrorContext = PendingThemeResolution;
+
+export interface ThemeController {
```
amadeus, line 49: Similar question from one of the other files -- is there a reason to avoid a class for this stuff and then have to separate implementation from annotation?

amadeus (commented):
Def some cool lookin stuff here!

I do have a concern about changing the default of diffshub, and also some higher level questions around the use of interfaces vs classes

apps/docs/app/(diffs)/_components/WorkerPoolContext.tsx
necolas: sure we can stick with the soft themes as defaults

packages/theming/src/modules/normalizeThemeColors.ts
necolas: Have you encountered values with whitespace before?

packages/diffs/package.json
necolas: It's what worked for now because the package hasn't been published. How do you guys publish packages? Is there some automation around that?

SlexAxton: bun publish handles this, we use it elsewhere. this is the correct way to define it.

packages/theming/src/modules/normalizeThemeColors.ts
amadeus: Nope, just thinking out loud if we should attempt to validate user input more specifically is all.

apps/docs/app/(diffs)/_components/WorkerPoolContext.tsx
necolas: Fixed

packages/theming/src/modules/normalizeThemeColors.ts
necolas: Got it. These values come from theme JSON and I haven’t seen whitespace issues in in practice yet. Adding `.trim()` would be speculative input-validation but easy to add later if a malformed theme ever shows up.

packages/theming/src/modules/createThemeResolver.ts
necolas: I like the factory pattern because these objects get destructured and passed around (controller methods into React hooks, the resolver into the fallback controller). Class methods can hit `this`-binding issues. Closures also give real privacy without `#fields` and are trivial to mock. Happy to add short comments at the implementations if the doc/impl split is the main concern.

packages/diffs/src/utils/getHighlighterThemeStyles.ts
necolas: Done

packages/diffs/src/highlighter/themes/themeResolver.ts
necolas: It's intentional because the public surface for diffs is the wrapper functions (`resolveTheme`, `getResolvedThemes`) that include the diffs-specific behavior the generic resolver doesn’t know about. Stuf like the worker-context guard, Shiki normalization, the bundled-theme fallback, and `theme.name` validation. Exporting the raw resolver would let callers bypass all of that even though that's probably not what we or they want

docs/superpowers/plans/2026-06-06-theme-catalog-api.md
necolas: Fixed

apps/docs/app/(diffshub)/(view)/_components/_theming/react/useThemeSelection.ts
necolas: This is just the fallback sentinel

apps/docs/app/(diffshub)/(view)/_components/useThemeCycle.ts
necolas: Removed these types, but matching the types to supported themes is complicated. Recovering literal names in the types would means threading literal generics through `createThemeCollection`/`createThemeCatalog`. The package itself should stay string-generic (it’s reusable and can’t know an app’s theme set), so this would be only for app-level type safety. I can look into it as a follow up

amadeus: Word, no need to change if it's super annoying, was mostly just curious.

packages/diffs/src/highlighter/themes/themeResolver.ts
amadeus: yeah that makes sense, sgsg

amadeus (approved):
lgtm! 🙏 thanks for cleaning up all this stuff!

PR #779, opened 2026-06-05
Pin GitHub Actions to commit SHAs

necolas:
Pins every third-party action to a full commit SHA and adds the tooling to keep that policy fresh and enforced. A version tag like `@v4` is mutable and whoever controls the action can repoint it at new code; SHA pinning is
a way to mitigate supply-chain risks of actions we don't own.

### Changes
- **Pin all external actions to SHAs** (`checkout`, `upload`/`download-artifact`,
  `cache`, `setup-bun`, `stickydisk`), each annotated with its `# vX.Y.Z`.
- **Dependabot** (`github-actions`, weekly, grouped) opens a single PR to bump
  the SHAs and their version comments. Uses `directories` with a `/.github/actions/*`
  glob so the CI workflows and every local composite action are covered.
- **`actions-pinned` CI job** fails any change that introduces an external
  action referenced by tag/branch instead of a 40-char SHA (local `./` actions
  exempt), so the policy can't silently regress.

### Notes
- The pinning behavior is unchanged from the version tags — SHAs were resolved
  from each action's current release.
- After merge: add **`Actions pinned to SHA`** to the `main` branch-protection
  required checks so the guard blocks rather than just advises. Dependabot only
  activates once `dependabot.yml` is on the default branch.

### DIFF
--- a/.github/actions/setup/action.yml
+++ b/.github/actions/setup/action.yml
@@ -22,12 +22,12 @@ runs:
         echo "version=$version" >> "$GITHUB_OUTPUT"
 
     - name: Setup Bun
-      uses: oven-sh/setup-bun@v2
+      uses: oven-sh/setup-bun@0c5077e51419868618aeaa5fe8019c62421857d6 # v2.2.0
       with:
         bun-version: ${{ steps.bun-version.outputs.version }}
 
     - name: Cache dependencies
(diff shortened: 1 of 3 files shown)


mdo (approved):

necolas:
Rebased on main and fixed format error

PR #787, opened 2026-06-08
feat(diffs/editor): Add marker support

ije:
preview: 

https://github.com/user-attachments/assets/f2d09f3e-0728-418c-b67e-c29953feb94e


packages/diffs/src/editor/editor.ts
```diff
@@ -70,31 +72,23 @@ import {
...
-
 export interface EditorOptions<LAnnotation> {
   /** Render rounded corners for selection ranges, default is true. */
   roundedSelection?: boolean;
   /** Show the clickable quick edit icon, default is disabled. */
   enabledQuickEdit?: boolean;
   /** Render the quick edit widget element. */
   renderQuickEdit?: (context: QuickEditContext<LAnnotation>) => HTMLElement;
+  /** Render custom marker message content. Falls back to a default when omitted. */
+  renderMarkerMessage?: (marker: Marker) => HTMLElement;
```
amadeus: Are we exposing this as a public API? If so we'd probably want to figure out a way to do this with react as well.

Also shouldn't block this merge, just to keep in mind.

packages/diffs/src/editor/editor.ts
```diff
@@ -1459,29 +1457,6 @@ export class Editor<LAnnotation> implements DiffsEditor<LAnnotation> {
     }
   }
 
-  #focus(position?: Position, preventScroll = true) {
```
ije, line 1462: sorry, did  some refactors make this messy

packages/diffs/src/editor/editor.css
```diff
@@ -90,18 +91,42 @@
...
+}
+[data-marker-warning] {
+  background-color: var(--diffs-editor-warning-fg, #cca700);
+}
+[data-marker-info] {
+  background-color: var(--diffs-editor-info-fg, #3794ff);
+}
+[data-marker-hint] {
+  background-color: var(--diffs-editor-hint-fg, #6a6a6a);
+}
```
amadeus, line 129: @mdo we should probably figure out how to use theme based colors for these. Also we should confirm if these are the colors we want as defaults.

Shouldn't block this PR, just something to follow up fix with.

amadeus (approved):

amadeus:
I had to rebase main into beta-1.3 because we reworked some vercel projects. so this will probably need another rebase

packages/diffs/src/editor/editor.ts
ije: we can remove the option, since the marker message can be `{ html: string }` now

PR #791, opened 2026-06-09
ci(docs): Build trees site variant in CI

necolas:
Build the trees site variant in CI so it is build-verified alongside diffs and diffshub. Previously only the diffs (default) and diffshub variants were built, so trees route or build regressions could merge unnoticed.


SlexAxton (approved):
Lots of trees stuff would still fail presumably, but this is probably better.

PR #792, opened 2026-06-09
Extract diffshub into a standalone app and gate docs route groups

necolas:
### Description

Extracts diffshub out of the multi-site `apps/docs` Next app into its own standalone app at `apps/diffshub`, wired to the shared `@pierre/theming` package via `useThemeController` (no `NEXT_PUBLIC_SITE`). The remaining docs sites (diffs, trees) now gate each route group with a `notFound()` layout guard so a given build only serves its own site's routes, and their nav/footer cross-link out to diffshub. CI builds the standalone diffshub app alongside docs.

### Motivation & Context

Next route groups don't prune at build time, so every product's routes — notably diffshub's catch-all — compiled into every docs-site build and leaked across sites. Splitting standalone apps out and gating the docs route groups fixes that leak and gives diffshub its own deployable app.


apps/diffshub/.gitignore
```diff
@@ -0,0 +1,41 @@
...
+
+# env files (can opt-in for committing if needed)
+.env*
+
+# vercel
+.vercel
+
+# typescript
+*.tsbuildinfo
+next-env.d.ts
```
amadeus, line 41: does diffshub need it's own .gitignore?

amadeus (approved):

.github/workflows/ci.yml
```diff
@@ -23,8 +23,9 @@ jobs:
       - name: Build docs (generates .source)
         run: bun ws docs build
 
-      - name: Build docs (diffshub)
-        run: NEXT_PUBLIC_SITE=diffshub bun ws docs build
+      # diffshub is now its own standalone app rather than a docs site variant.
+      - name: Build diffshub
+        run: bun ws diffshub build
```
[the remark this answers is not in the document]
necolas: Outdated

apps/diffshub/.gitignore
necolas: i think this is basically the standard next.js ignore

PR #809, opened 2026-06-11
[dev] Adopt moonrepo

SlexAxton:
# Replace custom script orchestration with moon

   Migrates the monorepo from hand-rolled task running (`scripts/ws.ts`, per-package
   npm scripts, `concurrently`, husky, artifact-passing CI) to
   [moon v2.3.3](https://moonrepo.dev), pinned in `.prototools` and enforced via
   `versionConstraint`. Every human/agent/CI entry point is now a moon task;
   `package.json` scripts are npm lifecycle hooks only.

   ## Command surface

   | Before | After |
   |---|---|
   | `bun ws <project> <task>` | `moon run <project>:<task>` / `moonx <project>:<task>` |
   | `bun run tsc` (root or package) | `moonx <project>:typecheck` / `moon run :typecheck` |
   | `bun run lint` / `format` / `lint:css` | `moon run root:lint` / `root:format` / `root:lint-css` |
   | `bun run wt …` | `moonx root:wt -- …` |
   | `test:e2e`, `dev:demo`, `icons:sprite`, … | kebab-case: `test-e2e`, `dev-demo`, `icons`, … |

   Task names are flat kebab-case with one meaning everywhere: `build` produces dist,
   `dev` is the persistent watcher, `test` is `bun test`, `tsc` is renamed `typecheck`.

   ## Task architecture

   - `.moon/tasks/bun-common.yml` — `test` + `typecheck` for every bun project, both
     with `deps: ['^:build']` (cross-package imports and types resolve through each
     dependency's built `dist`; this replaces every `build:deps` script and fixed a
     real race where `diffshub:test` could run before `theming:build` finished).
   - `.moon/tasks/tag-tsdown.yml` — `build`/`dev` for the tsdown packages; theming and
     trees override to append their post-build asserts.
   - `.moon/tasks/tag-publishable.yml` — `prepublish` guard chain
     (`assert-bun-version` + `build`) behind each package's `prepublishOnly`.
   - Per-project `moon.yml` for everything else (apps, benchmarks, e2e, publish).

   App `dev` tasks depend on `^:build` + `^:dev`: moon finishes dependency builds,
   then runs dependency watchers alongside the app server — `concurrently` is gone.
   Port offsets still come from `.env.worktree` (loaded via task `envFile`);
   `wt.ts`, `run-dev.sh`, and `load-worktree-env.mjs` survive unchanged in role.

   ## CI

   One affected-aware lane (plus the unchanged `actions-pinned` guard):

   - `moon ci --include-relations <explicit targets>` computes affected projects from
     the VCS diff and expands through graph relations, so a package change re-runs its
     dependents' builds/tests/typechecks — including trees e2e when affected.
   - Caching via Blacksmith sticky disks: `~/.proto`, bun's download cache, moon's
     portable cache (`hashes` + `outputs`), playwright browsers. No more artifact
     upload/download between jobs.
   - Toolchain bootstrap is proto-only (`moonrepo/setup-toolchain`, SHA-pinned); moon
     installs dependencies itself with a frozen lockfile. The bun-version awk parsing,
     `setup-bun`, and manual `bun install` are gone.
   - The playwright install is skipped (fail-open) when e2e is unaffected.

   ## Hooks

   moon-managed (`vcs.hooks`, worktree-compatible — hooks land in `.moon/hooks` with a
   worktree-local `core.hooksPath`; no install-time `prepare`). Pre-commit runs
   `moon exec :typecheck --affected --status staged` (now also checks dependents,
   which `precommit-tsc.ts` missed) + `lint-staged`. Pre-push keeps the git-lfs guard;
   generated hook scripts forward `"$@"` and stdin. husky and `precommit-tsc.ts` are
   deleted.

   ## Vercel

   Builds run via the GitHub integration, so each app ships a `vercel.json` whose
   `buildCommand` runs the app's moon build (moon comes from a pinned root
   devDependency; PATH covers both repo-root and app-dir Root Directory layouts).
   The two docs projects share one file and dispatch on `NEXT_PUBLIC_SITE`.
   docs `distDir` is now per-site in every mode (`.next/diffs` / `.next/trees`),
   making the two site builds disjoint and independently cacheable; they share a
   mutex only for the site-dependent `public/llms*.txt`.

   ## Deliberate behavior changes

   - `tsc` → `typecheck` everywhere.
   - `assert-bun-version` guards all five published packages (was diffs-only);
     storage-elements-next gains its first prepublish guard.
   - theming is now watched during docs/diffshub dev (was imported but unwatched).
   - path-store's `build` is renamed `build-demo` (it builds the demo site).
   - Type-aware `root:lint` now orders after package builds (it resolves `@pierre/*`
     types through dist; previously raced `tsdown --clean`).
   - CI consolidates six checks into two.
   - `runInCI` policy: graph-edge-free local tasks are `'always'` (work in CI-marked
     agent shells); graph-connected ones (`dev`/`prod`, e2e variants, publish guards)
     stay `'skip'` and need a `CI=` prefix in such shells (`CI= moonx docs:dev-diffs`).

   ## Verification

   - Full pipeline green locally, cold and warm (33 targets; warm rerun 5.2s,
     30 cache hits); affected-skip verified on isolated changes.
   - trees e2e: 58 passed through the real CI path (nested `moon run` webServer).
   - Worktree dev smoke: 4 watchers + next on offset port, hot reload on theming edit.
   - `bun publish --dry-run` × 5 fires the prepublish chains (trees' guard correctly
     blocks direct publish).
   - Hook smoke: staged type error blocks the commit; hooks ran on every commit here.
   - Gates: every `scripts` block is lifecycle-only; zero `bun ws`/`husky`/
     `concurrently` references in live code; all embedded invokers (playwright
     webServers, profiler/test spawns, trees publish flow) point at moon tasks.
   - Reviewed by an 8-agent pass; all confirmed findings fixed (cache-input gaps,
     broken spawns, lint/build race, Vercel PATH robustness, CI lane consolidation).

   ## Merge checklist (externally gated)

   - [ ] Vercel dashboard audit per project: Root Directory, Install Command,
         `NEXT_PUBLIC_SITE` env (and whether demo is a fourth project)
   - [ ] Preview deploys green for all docs/diffshub(/demo) projects on this PR
   - [ ] First Blacksmith run: stickydisk mounts + PR base detection (fallback:
         explicit `--base` in the workflow)
   - [ ] One real LFS push through the moon pre-push hook

   ## Follow-ups (deferred by design)

   - Depot remote cache when wall-clock or local cache hydration warrants it
     (stickydisk remains as L1).
   - docs `test-e2e` in CI — enable deliberately, not as migration fallout.
   - TypeScript project-reference syncing — separate project if typecheck time hurts.


amadeus (approved):
👯 lfg 👯

.agents/skills/testing-and-verification/SKILL.md
```diff
@@ -10,29 +10,31 @@ description:
...
+Useful check/fix pairs on the root project:
 
 ```bash
-bun run format:check
-bun run format
-bun run lint
-bun run lint:fix
-bun run lint:css
-bun run lint:css:fix
+moon run root:format-check
```
necolas, line 23: Is this different from how moon works in the monorepo (e.g, iirc, `moon :format-check` etc) or just another way of doing the same thing?

.agents/skills/tooling-and-dependencies/SKILL.md
```diff
@@ -32,18 +46,35 @@ This monorepo uses Bun's `workspaces.catalog` in the root `package.json`.
...
+- Tasks are defined in `.moon/tasks/*.yml` (inherited) and each project's
+  `moon.yml`. Repo-wide tooling (format, lint, icons, clean) lives on the `root`
+  project.
+- Run tasks from anywhere in the repo:
 
 ```bash
-bun ws <project> <task>
-bun ws <project> <task> --some --flag
+moon run <project>:<task>
+moonx <project>:<task>             # shorthand for moon exec
```
necolas, line 61: what's the difference between `moon exec <project>:<task>` and `moon <project>:<task>`?

SlexAxton: `moon run` === `moonx`

.agents/skills/testing-and-verification/SKILL.md
SlexAxton: it's a little differently setup. the short version is that our formatters and linters are so fast that it didn't make a ton of sense to not just run them from root every time. it's probably true in monorepo as well now, but still has configuration from when it wasn't true.

SlexAxton: that said `moonx :format-check` will run `format-check` in all the places it exists, which should just be `root` anyways. so it should work fine that way.

.github/workflows/ci.yml
```diff
@@ -7,138 +7,54 @@ on:
...
+        if: steps.e2e-affected.outputs.run == 'true'
+        run: bunx playwright@1.51.1 install --with-deps chromium
 
-      - name: Type check all workspaces
-        run: bun run ws "*" tsc
+      - name: Run affected tasks
+        run: >-
+          moon ci --include-relations :build docs:build-trees-site
+          path-store:build-demo :test :typecheck trees:test-e2e root:lint
+          root:lint-css root:format-check
```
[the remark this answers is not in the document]
SlexAxton: resolved

.agents/skills/testing-and-verification/SKILL.md
necolas: so `root:<task>` and `:<task>` are the same thing? or does `root` only run the task in the top-level `package.json`?

necolas: i don't really get why these commands are using `moon run` whereas we use `moonx` later on. it seems like `moonx` is shorthand for `moon exec`, and `moon run` is a higher-level wrapper around all that and not meant to be used for CI. i'm not 100% clear on when to use what, but that's what I've gathered

necolas (approved):

.agents/skills/testing-and-verification/SKILL.md
SlexAxton: pushed up some clarification in a new commit: https://github.com/pierrecomputer/pierre/commit/0e50e399f82ac788d63df8835d64e87f5aa1690c

PR #813, opened 2026-06-15
[diff/editor] Add auto-surround feature

ije:
- [x] handle auto-surround
- [x] `autoSurround` option

https://github.com/user-attachments/assets/e7b10fa1-cb3d-4638-81dc-b375aebd1c13


mdo:
@ije FYI, I rebased this to remove a bunch of the earlier `beta-1.3` commits.

PR #817, opened 2026-06-16
[diff/editor] Add Find-in-file replace support

ije:
- [x] Replace functional
- [x] Dom layout
- [x] UI deisign @mdo


necolas (approved):
lgtm

PR #818, opened 2026-06-16
[diffs/editor] Support unified diff

ije:
- [x] Allow 'unified' layout on an editor attach to the file diff component
- [x] Inline-edit doesn't trigger re-render now
- [x] refresh view when typing


mdo:
Tested this out locally in my own branch, so merging!

PR #820, opened 2026-06-16
[diffs] Removing BETA badge from Virtualization

amadeus:
Noticed this branch added these new fancy beta tags, but I already moved the Virtualization component out of beta on `main` by removing the various beta/api notices, and needed to add an additional change here since the beta tags are new.


mdo (approved):
Alpha move.

PR #821, opened 2026-06-16
[diffs] Editable CodeView

amadeus:
@ije this PR quickly adds an editable toggle to the CodeView demo and hooks it up. It kinda works but if you scroll down things stop working.

I think there's also a higher level API type question we need to resolve -- CodeView uses a shared options interface, so the way we are forcing options to be a certain configuration is not ideal in a CodeView environment and we might want to think about how to handle this.

### DIFF
--- a/apps/demo/src/codeViewDemo.ts
+++ b/apps/demo/src/codeViewDemo.ts
@@ -5,13 +5,17 @@ import {
   type CodeViewItem,
   type CodeViewOptions,
   type DiffLineAnnotation,
+  type DiffsEditableComponent,
   type DiffsThemeNames,
+  type FileContents,
   type FileDiffContentsLoader,
+  type FileDiffMetadata,
   type LineAnnotation,
   type ParsedPatch,
   type SelectedLineRange,
   type ThemesType,
 } from '@pierre/diffs';
+import { Editor } from '@pierre/diffs/editor';
 import type { WorkerPoolManager } from '@pierre/diffs/worker';
 
 import { FAKE_DIFF_LINE_ANNOTATIONS, type LineCommentMetadata } from './mocks/';
@@ -43,6 +47,28 @@ interface CodeViewDemoInstance {
   options: CodeViewOptions<CodeViewCommentMetadata>;
 }
 
+type CodeViewEditableInstance =
+  DiffsEditableComponent<CodeViewCommentMetadata> & {
+    file?: FileContents;
+    fileDiff?: FileDiffMetadata;
+  };
+
+interface CodeViewEditableContext {
+  item: CodeViewItem<CodeViewCommentMetadata>;
+  instance: CodeViewEditableInstance;
+}
+
+interface ActiveCodeViewEditor {
+  itemId: string;
+  viewer: CodeView<CodeViewCommentMetadata>;
+  items: CodeViewItem<CodeViewCommentMetadata>[];
+  instance: CodeViewEditableInstance;
+  editor: Editor<CodeViewCommentMetadata>;
+  dispose: () => void;
+  toggleInput: HTMLInputElement;
+  dirty: boolean;
+}
+
 type CodeViewDemoAnnotation =
   | DiffLineAnnotation<CodeViewCommentMetadata>
   | LineAnnotation<CodeViewCommentMetadata>;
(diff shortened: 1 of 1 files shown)


apps/demo/src/codeViewDemo.ts
```diff
@@ -111,35 +135,232 @@ export function renderDemoCodeView(
...
+    // just ignore their values while editing? CodeView intentionally shares
+    // most options, so ideally we don't toggle the entire CodeView when
+    // editing a single file
+    lineHoverHighlight: 'disabled',
+    enableLineSelection: false,
+    enableGutterUtility: false,
+    // I assume once you merge the unified support, this won't be a requirement anymore
+    diffStyle: diffStyle === 'unified' ? 'split' : diffStyle,
+    useTokenTransformer: true,
+    expandUnchanged: true,
```
amadeus, line 168: Do we HAVE to force expandUnchanged true?

apps/demo/src/codeViewDemo.ts
```diff
@@ -111,35 +135,232 @@ export function renderDemoCodeView(
...
+
+    // These settings are not compatible with editor... is there a way we can
+    // just ignore their values while editing? CodeView intentionally shares
+    // most options, so ideally we don't toggle the entire CodeView when
+    // editing a single file
+    lineHoverHighlight: 'disabled',
+    enableLineSelection: false,
+    enableGutterUtility: false,
+    // I assume once you merge the unified support, this won't be a requirement anymore
+    diffStyle: diffStyle === 'unified' ? 'split' : diffStyle,
```
amadeus, line 166: Obviously this wont be an issue with #818

apps/demo/src/codeViewDemo.ts
```diff
@@ -111,35 +135,232 @@ export function renderDemoCodeView(
...
+    useTokenTransformer: true,
```
amadeus, line 167: How come this is necessary?

apps/demo/src/codeViewDemo.ts
```diff
@@ -111,35 +135,232 @@ export function renderDemoCodeView(
...
       console.log('CodeView selected lines', selection);
     },
+
+    // These settings are not compatible with editor... is there a way we can
+    // just ignore their values while editing? CodeView intentionally shares
+    // most options, so ideally we don't toggle the entire CodeView when
+    // editing a single file
+    lineHoverHighlight: 'disabled',
+    enableLineSelection: false,
+    enableGutterUtility: false,
```
amadeus, line 164: I'd prefer not to toggle this globally for CodeView just to support editing, is there a way we can like let these options be whatever, but then disable the functionality while editing?

apps/demo/src/codeViewDemo.ts
ije: the editor depends on the `data-char` attr for selection rendering

apps/demo/src/codeViewDemo.ts
ije: the page may jitter when editing  with expand unchanged option. we can support this in the future?

apps/demo/src/codeViewDemo.ts
ije: yes, the editor will update the options and rerender if the option is not supported. you don't have to add these options.

apps/demo/src/codeViewDemo.ts
ije: and this only is required with worker pool

apps/demo/src/codeViewDemo.ts
```diff
@@ -121,35 +145,232 @@ export function renderDemoCodeView(
...
 }
 
 export function setCodeViewDiffStyle(diffStyle: CodeViewDiffStyle) {
+  deactivateCodeViewEditor();
   for (const codeView of codeViewInstances) {
-    codeView.options = { ...codeView.options, diffStyle };
+    codeView.options = {
+      ...codeView.options,
+      // Ideally we don't do this...
+      diffStyle: diffStyle === 'unified' ? 'split' : diffStyle,
```
ije, line 191: unified supported now

amadeus:
Closing this PR because I actually want to make the editing configuration a first class citizen, so this hack to get it working is not really relevant and should never merge.

PR #822, opened 2026-06-17
fix(diffs): Cut collapsed editor selections by line

necolas:
## Problem

Using the "cut" shortcut without a selection results in a silent copy. The contents of the line is copied to the clipboard but the text in the diff doesn't change.

## Fix

Route cut events through a dedicated cut path so caret-only cuts copy and remove the full logical line instead of behaving like copy.

I had AI create an editor clipboard integration test in lieu of e2e testing.


packages/diffs/src/editor/editor.ts
```diff
@@ -2351,6 +2355,205 @@ export class Editor<LAnnotation> implements DiffsEditor<LAnnotation> {
...
+      }
+    );
+    textDocument.setLastUndoSelectionsAfter(nextSelections);
+    this.#applyChange(
+      change,
+      nextSelections,
+      this.#applyChangeToLineAnnotations(change)
+    );
+  }
+
```
ije, line 2511: so many locs added for the fix🥹, maybe we can simplify a bit? i will look into

necolas: well cut was broken, multi-cursor cut wasn't working, and line-breaks in copy and cut weren't robust

ije (approved):
solid implement 🫡 i moved some methods to `selection.ts` as pure functions

necolas:
thanks, that's tidy!

PR #823, opened 2026-06-17
 fix(diffs): Handle IME composition input

necolas:
Fixes the following issue:
1. Open an editable diffs editor and place the caret in content.
2. Start a CJK IME or accented/dead-key composition.
3. Type a preview candidate, then commit it or press Esc to cancel.

Previously, each preview beforeinput was prevented and warned as an unknown input type, so the browser could not show native preview text and canceled composition text could still be inserted on compositionend.

Let insertCompositionText beforeinput events stay with the browser. Track composition updates and only commit non-canceled compositionend data into the editor model. Add regression coverage for preview, commit, warning noise, and canceled composition text.

### DIFF
--- a/packages/diffs/src/editor/editor.ts
+++ b/packages/diffs/src/editor/editor.ts
@@ -176,6 +176,7 @@ export class Editor<LAnnotation> implements DiffsEditor<LAnnotation> {
   #searchPanel?: SearchPanelWidget;
   #selectionAction?: SelectionActionWidget;
   #shouldIgnoreSelectionChange = false;
+  #isComposing = false;
   #isGutterMouseDown = false;
   #isContentMouseDown = false;
   #shiftKeyPressed = false;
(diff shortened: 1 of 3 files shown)


ije (approved):
Nice!!!!

necolas:
Fixed by the `getComposedRanges` issue by feature-checking the method and bailing out early. I did not use the `getRangeAt()` fallback. The editor renders inside a shadow root, and `getRangeAt()` doesn’t read across the shadow boundary like `getComposedRanges` does, so it would return the wrong range. Added a regression test that reproduces the throw without the guard.

Stale file contents on full rerenders seems to be pre-existing code in the highlight/render path, unrelated to this IME change. The suggested “update cached contents” fix risks regressing highlighting and can be tracked separately.

ije:
> Fixed by the getComposedRanges issue by feature-checking the method and bailing out early. I did not use the getRangeAt() fallback. The editor renders inside a shadow root, and getRangeAt() doesn’t read across the shadow boundary like getComposedRanges does, so it would return the wrong range. Added a regression test that reproduces the throw without the guard.

yep, it's vary trick to handle selection in shadow dom.

> Stale file contents on full rerenders seems to be pre-existing code in the highlight/render path, unrelated to this IME change. The suggested “update cached contents” fix risks regressing highlighting and can be tracked separately.

i think this is ok. the editing won't trigger full-rerender with the stable file.

PR #830, opened 2026-06-17
fix(diffs): Disable caret blink under reduced motion

necolas:
The editor caret animated with `blinking 1.2s infinite` regardless of the user's OS reduced-motion preference, so the caret kept pulsing even when motion was meant to be suppressed.

Add a `@media (prefers-reduced-motion: reduce)` rule that sets `animation: none` on `[data-caret]`. With the animation removed, `opacity` falls back to its default of `1`, so the caret renders as a solid bar (still gated by the existing focus/visibility rules) instead of disappearing or pulsing.

A CSS media query is used rather than the existing `prefersReducedMotion()` helper because the caret is a purely declarative CSS animation; the media query reacts live to OS preference changes without any JS listener. The helper stays reserved for the JS-driven scroll-behavior decision.

### DIFF
--- a/packages/diffs/src/editor/editor.css
+++ b/packages/diffs/src/editor/editor.css
@@ -81,6 +81,13 @@
   animation-delay: 0.8s;
   visibility: hidden;
 }
+/* Honor the OS reduced-motion setting: drop the blink so the caret renders as a
+   solid bar (opacity stays at its default of 1) instead of pulsing. */
+@media (prefers-reduced-motion: reduce) {
+  [data-caret] {
+    animation: none;
+  }
+}
 [data-selection-range] {
   z-index: -10;
   background-color: var(--diffs-editor-selection-bg);


packages/diffs/src/editor/editor.css
```diff
@@ -81,6 +81,13 @@
   animation-delay: 0.8s;
   visibility: hidden;
 }
+/* Honor the OS reduced-motion setting: drop the blink so the caret renders as a
+   solid bar (opacity stays at its default of 1) instead of pulsing. */
+@media (prefers-reduced-motion: reduce) {
+  [data-caret] {
+    animation: none;
+  }
+}
```
mdo, line 90: Could we save some repetition and invert the media query for the same effect?

```suggestion
[data-caret] {
  width: 2px;
  background-color: var(
    --diffs-bg-caret-override,
    var(
      --diffs-editor-cursor-fg,
      light-dark(
        color-mix(in lab, var(--diffs-fg) 50%, var(--diffs-bg)),
        color-mix(in lab, var(--diffs-fg) 75%, var(--diffs-bg))
      )
    )
  );
  visibility: hidden;
}
/* Effectively reduce motion if user has set a preference */
@media (prefers-reduced-motion: no-preference) {
  [data-caret] {
      animation: blinking 1.2s infinite;
      animation-delay: 0.8s;
  }
}
```

mdo (approved):
Happy to see merged either way, though.

packages/diffs/src/editor/editor.css
amadeus: Wouldn't visibility: hidden mean you never see the cursor?

ije (approved):
nice!

packages/diffs/src/editor/editor.css
necolas: I created a task to revisit the CSS

PR #832, opened 2026-06-17
perf(diffs): Skip redundant hunk recompute on line-count edits

necolas:
Editing a diff was re-diffing the whole file twice for every keystroke that adds or removes a line (Enter, Backspace that merges lines, typing over a multi-line selection).

For such edits the editor calls updateRenderCache and then applyDocumentChange in the same pass, and both rebuild hunk metadata with a full recompute. The recompute in updateRenderCache runs against a half-updated line array and is immediately discarded by applyDocumentChange, which recomputes from the authoritative document text — so it is pure wasted work.

I don't think it can be made incremental: updateDiffHunks falls back to a full recompute whenever the addition and deletion line counts differ. Since applyDocumentChange already does the authoritative recompute, the editor now tells updateRenderCache to skip the hunk recompute on these edits. The per-line token and text updates still run, so the rendered output should be unchanged.

### DIFF
--- a/packages/diffs/src/components/FileDiff.ts
+++ b/packages/diffs/src/components/FileDiff.ts
@@ -1045,9 +1045,14 @@ export class FileDiff<
   public updateRenderCache(
     dirtyLines: Map<number, Array<HighlightedToken>>,
     themeType: 'dark' | 'light',
-    shouldRefreshView?: boolean
+    shouldRefreshView?: boolean,
+    skipDiffRecompute?: boolean
   ): void {
-    this.hunksRenderer.updateRenderCache(dirtyLines, themeType);
+    this.hunksRenderer.updateRenderCache(
+      dirtyLines,
+      themeType,
+      skipDiffRecompute
+    );
     if (shouldRefreshView === true) {
       if (this.options.diffStyle === 'split') {
         if (this.fastRefreshTimeout != null) {
--- a/packages/diffs/src/editor/editor.ts
+++ b/packages/diffs/src/editor/editor.ts
@@ -1454,7 +1454,10 @@ export class Editor<LAnnotation> implements DiffsEditor<LAnnotation> {
     fileInstance.updateRenderCache(
       dirtyLines,
       tokenizer.themeType,
-      isDiff && !didLineCountChange
+      isDiff && !didLineCountChange,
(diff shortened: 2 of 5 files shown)


amadeus (approved):
good catch, i actually have a comment buried somewhere in a much older PR that I need to think about how we can update FileDiffMetadata more incrementally, since it would be a lot more efficient.

necolas:
@amadeus Curious to hear your thoughts about possible next steps in this area too. I had AI suggest ideas for how to fix this at a deeper level, including the unresolved TODO in `applyDocumentChange`:

```
public applyDocumentChange(textDocument, newLineAnnotations?) {
  this.hunksRenderer.applyDocumentChange(textDocument);
  const renderDiff = this.hunksRenderer.getRenderDiff();
  if (renderDiff != null) this.fileDiff = renderDiff;
  // TODO(@ije): can we avoid full-render here?
  this.rerender();
  this.interactionManager.setSelectionDirty();
  …
}
```

1. Separate token-sync from hunk-recompute (removes the flag). Make updateRenderCache token-content-only, and make hunk recompute an explicit call the editor makes exactly once per edit — incremental for content-only, full-from-document for line-count. Then the call graph itself guarantees “one recompute per edit,” the deferred-tokenize path simply never calls recompute, and the boolean disappears. Moderate churn (splits one interface method into two across File/FileDiff/renderers), low conceptual risk.

2. Thread the change delta down (the bigger lever — kills the open TODO). The editor already has a structured TextDocumentChange (startLine, lineDelta) but hands components only the whole textDocument and dirtyLines. That’s why applyDocumentChange re-splits the entire file and rerender() rebuilds everything — the component was never told what changed. Pass the delta down and both the recompute and the render can be localized to the affected region. That’s the actual path to removing the remaining O(file) cost, not just the duplicate.

amadeus:
I don't know enough about the other editing flows to say for sure what the right answer is there. I would need to dig in a bit more.

I was just thinking more from the perspective of -- `incrementally updating FileDiffMetadata would be much more efficient than pulling out all the content and re-computing the diff`, but it's also probably not trivial

ije (approved):
super!

ije:
The `applyDocumentChange` triggers full-render is a big performance issue here. i tried to use the new `refreshDiffView` method that should be faster but it causes view glitch.

PR #836, opened 2026-06-18
agents(git-commits): Lead commit descriptions with the user flow

necolas:
A reviewer reading a fix or feature first wants to know what was happening and how to reproduce or experience the change. The skill only asked for "motivations and changes", so descriptions tended to open with the technical cause instead.

Add a "Writing the Description" recipe that puts the flow first and the cause and fix second, require plain language in the subject and first paragraph, and tell authors to leave test counts and other process narration out of the message.


necolas:
@amadeus tried updating the commit skill to prefer clear language up front

@mdo messages for bugs should start with clearer repo steps

PR #841, opened 2026-06-18
fix(diffs): re-measure character width after web font loads

necolas:
Steps to reproduce:
1. Open the editor on a page that uses a custom monospace web font.
2. Load it with a cold cache so the editor renders before the font file finishes downloading.
3. Click in the code to place a caret.

The caret and selection sit left of the real text and the gutter is too narrow. The editor measures the '0' character width on first render and uses it to size the gutter and place every caret. Before the web font loads, that width comes from the fallback font. getComputedStyle returns the same font-family string before and after the font arrives, so init() never re-measures and the wrong width sticks.

Re-measure the '0' width once document.fonts.ready resolves. If it changed, drop the cached widths and offsets and repaint the carets, selections, and markers so they match the loaded glyphs. Do nothing when the width is unchanged or the font was already loaded.

The re-measure is scheduled once per mount; cleanUp() clears the guard so a reused editor instance schedules it again for a font that may not be loaded yet.

### DIFF
--- a/packages/diffs/src/editor/editor.ts
+++ b/packages/diffs/src/editor/editor.ts
@@ -192,6 +192,7 @@ export class Editor<LAnnotation> implements DiffsEditor<LAnnotation> {
   #scrollingToLineChar?: number;
   #scrollingToLineNoFocus = false;
   #retainSearchPanelFocus = false;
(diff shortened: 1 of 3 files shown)


ije (approved):
nice catch!

PR #843, opened 2026-06-18
feat(diffs): Add programmatic undo and redo to the editor

necolas:
The editor could only undo and redo through keyboard shortcuts. Apps that drive the editor (toolbar buttons, AI edits, custom shortcuts) had no way to trigger or check undo state.

Expose `undo()`, `redo()`, `canUndo`, and `canRedo` on the Editor class. The methods reuse the existing command path, so programmatic and keyboard undo behave the same and consumers still get the `onChange` callback after each one.


fat (approved):

fat:
🤝

ije (approved):
nice! i updated the docs as well

necolas:
Thank you!

PR #845, opened 2026-06-18
More edit CSS fixes

mdo:
- Remove some custom color changes for edit mode—unsure these were needed? Changing the line number foreground for example felt wrong.
- Consolidate colors for markers (both underline and popups)
- Add edit mode and markers to the `/playground` page
- Update playground button groups to use `size="icon"` in a few places
- Generate a new warning color token from themes

### DIFF
--- a/packages/diffs/src/editor/editor.ts
+++ b/packages/diffs/src/editor/editor.ts
@@ -2292,8 +2292,7 @@ export class Editor<LAnnotation> implements DiffsEditor<LAnnotation> {
 
     const { ch, lineHeight } = this.#metrics;
     const y = this.#getLineY(line) + wrapLine * lineHeight;
-    const css = `width:${width}px;transform:translateX(${left}px) translateY(${y}px);`;
-    const cacheKey = `${type}-${line}/${wrapLine}-${left}-${width}${extraDataset ?? ''}`;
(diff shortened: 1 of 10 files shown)


mdo:
@ije Please see [a4ac762](https://github.com/pierrecomputer/pierre/pull/845/commits/a4ac76210cb7df4dbd9ad14e3c0fd3617618f98a) for a potential fix on switching themes.

mdo:
And also @ije see [5db1de6](https://github.com/pierrecomputer/pierre/pull/845/commits/5db1de6ef68895b465d565fb7f6e5efd897b427f) for a potential fix on the word wrap + split diff bug. Unsure if this is the right move, totally vibe coded by Claude, so take with grain of salt.

packages/diffs/src/editor/editor.ts
```diff
@@ -1856,8 +1856,36 @@ export class Editor<LAnnotation> implements DiffsEditor<LAnnotation> {
     }
   }
 
+  // Anchor the editor overlay to the editable column's origin so the
```
fat: feel like we should be able to do this with raw css or something… 🤔

mdo: Yeah this is something @ije needs to revisit for sure—it's flagged in Discord around the `display: contents` issue in split + wrap.

ije (approved):
LGTM. Added `#contentOffset` metrics for split diff

PR #847, opened 2026-06-18
fix(diffs): memoize and round DOM text measurements

necolas:
Steps to reproduce:
1. Open the diffs editor on a line containing an emoji, such as const greeting = "😀 hello".
2. Move the caret along that line or extend a selection over it.
3. Watch the Performance panel: every caret or selection update forces a synchronous layout. For each non-ASCII run the editor inserts a hidden span, reads getBoundingClientRect(), and removes it, with no caching, so the same runs are re-measured on every render. That path also returns the raw sub-pixel width while the canvas and ASCII paths round, so offsets are quantized inconsistently across runs on the same line.

Fix:
- Memoize domMeasureTextWidth() by measured text, capped at 4096 entries with FIFO eviction, so repeats skip the reflow.
- Clear the cache in init() when the font changes, since cached widths are font-specific.
- Round the DOM-measured width to match canvasMeasureTextWidth and the ch metric.

Add regression tests for the rounding, one measurement per distinct run, and cache invalidation on font change.

### DIFF
--- a/packages/diffs/src/editor/textMeasure.ts
+++ b/packages/diffs/src/editor/textMeasure.ts
@@ -1,9 +1,22 @@
 import { h, round } from './utils';
 
+// Upper bound on cached DOM text-width measurements. The cache only holds
+// non-ASCII runs (emoji, ZWJ sequences, variation selectors), so it stays
+// small for ordinary code, but capping it prevents unbounded growth on
+// emoji-heavy documents. Past the cap the oldest entry is evicted.
+const DOM_WIDTH_CACHE_LIMIT = 4096;
+
 export class Metrics {
   #root?: HTMLElement;
   #canvasCtx?: CanvasRenderingContext2D;
   #font?: string;
+
+  // Memoizes domMeasureTextWidth() results keyed by the (tab-expanded) text it
+  // measured. That path inserts a span and reads getBoundingClientRect(), which
+  // forces a synchronous layout every call; without this cache, caret and
+  // selection updates on emoji-heavy lines re-measure the same runs on every
+  // render. Cleared in init() whenever the font changes (see below).
+  #domWidthCache = new Map<string, number>();
 
   /** Width of the '0' character. */
   ch: number = -1;
@@ -54,6 +67,8 @@ export class Metrics {
       this.#font = font;
       this.#canvasCtx.font = font;
       this.ch = this.canvasMeasureTextWidth('0');
+      // Cached DOM widths were measured against the previous font.
+      this.#domWidthCache.clear();
     }
     const nextTabSize = parseInt(tabSize, 10);
     if (!Number.isNaN(nextTabSize)) {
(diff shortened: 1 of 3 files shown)


packages/diffs/src/editor/textMeasure.ts
```diff
@@ -83,12 +98,17 @@ export class Metrics {
...
+   * this is slow because it cause a reflow, use it for non-ascii text;
+   * results are memoized per text so repeated measurements skip the reflow
    */
   domMeasureTextWidth(text: string): number {
     if (this.#root === undefined) {
       throw new Error('Metrics not initialized');
     }
+    const cached = this.#domWidthCache.get(text);
+    if (cached !== undefined) {
+      return cached;
```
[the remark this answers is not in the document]
necolas: The cache shares `ch`'s lifecycle: `ch`, `lineHeight`, `tabSize`, and `#font` are all computed only in `init()` (new content element), so a same-root font change leaves all of them stale, not just the cache. Since `ch` drives ASCII caret/selection positioning, the editor is already mismeasured in that scenario independent of this cache.

Keying the cache by font wouldn't address it either: the only font state available without a per-call layout read is `#font`, which is itself refreshed only in `init()` (it short-circuits for the same root), so it would be equally stale.

What this PR does change is that the DOM path went from re-measuring on every call to cached. To keep that from regressing the reflow case, c031f4df clears the cache from `#handleLayoutResize()`, alongside the sibling DOM-measurement caches (`#wrapLineOffsetsCache`, `#lineYCache`, `#lastAccessedCharX`) already invalidated there on a width change. A web font finishing loading reflows the text → resize fires → cache cleared → widths re-measure.

Fully reacting to live font changes (recomputing `ch`/metrics on `document.fonts` load) is a separate, pre-existing gap, out of scope here.

fat (approved):
🔥

packages/diffs/src/editor/textMeasure.ts
```diff
@@ -1,10 +1,26 @@
...
 
+  // Memoizes domMeasureTextWidth() results keyed by the (tab-expanded) text it
+  // measured. That path inserts a span and reads getBoundingClientRect(), which
+  // forces a synchronous layout every call; without this cache, caret and
+  // selection updates on emoji-heavy lines re-measure the same runs on every
+  // render. Cleared via clearTextWidthCache() whenever the inherited font may
+  // have changed: on a font change in init(), and on a layout reflow the
+  // editor reports through its resize handler (e.g. a web font finishing
+  // loading without the content element being replaced).
+  #domWidthCache = new Map<string, number>();
```
ije: nice!

packages/diffs/src/editor/textMeasure.ts
```diff
@@ -83,12 +101,17 @@ export class Metrics {
...
    * measure the width of the text using the DOM
-   * this is slow because it cause a reflow, use it for non-ascii text
+   * this is slow because it cause a reflow, use it for non-ascii text;
+   * results are memoized per text so repeated measurements skip the reflow
    */
   domMeasureTextWidth(text: string): number {
     if (this.#root === undefined) {
       throw new Error('Metrics not initialized');
     }
+    const cached = this.#domWidthCache.get(text);
```
ije: i think the cache key should be `text` plus `font`

ije: i fixed

ije (approved):

PR #848, opened 2026-06-18
fix(diffs): move caret and transpose by whole graphemes

necolas:
Steps to reproduce:
1. In the diffs editor, edit a line with an emoji, e.g. "a😀".
2. Put the caret after "a" and press Right. The caret lands inside the emoji instead of moving past it; a later insert or delete then splits it into a lone surrogate.
3. Put the caret at the end of "a😀" and press Ctrl-T. Instead of swapping to "😀a", the emoji is split into mojibake.

Cause: mapCursorMove stepped left/right by one UTF-16 code unit and transpose swapped single code units. The document is UTF-16, so an emoji takes two units; both operations landed on or split a surrogate pair, even though the module already segments graphemes for word operations.

Fix: move horizontally by whole grapheme clusters and transpose whole graphemes, via a shared getLineGraphemeStarts helper built on Intl.Segmenter (also reused by deleteWordBackward). ASCII behavior is unchanged. Add regression tests for left/right move, shift-select, and the mid-line, end-of-line, and cross-line transpose branches.

### DIFF
--- a/packages/diffs/src/editor/selection.ts
+++ b/packages/diffs/src/editor/selection.ts
@@ -149,26 +149,29 @@ export function mapCursorMove(
       const lineLength = textDocument.getLineLength(line);
       character = Math.min(character, lineLength);
       if (shortcut === 'left') {
-        character--;
-
-        if (character < 0) {
-          if (line === 0) {
-            character = 0;
-          } else {
-            line = Math.max(0, line - 1);
-            character = textDocument.getLineLength(line);
-          }
+        if (character > 0) {
+          // Step left by a whole grapheme so the caret never lands inside an
+          // emoji or other multi-code-unit character.
+          character = stepCharacterByGrapheme(
+            textDocument,
+            line,
+            character,
+            false
+          );
+        } else if (line > 0) {
+          line = line - 1;
+          character = textDocument.getLineLength(line);
         }
-      } else {
-        character++;
-        if (character > lineLength) {
-          if (line === lineCount - 1) {
-            character--;
-          } else {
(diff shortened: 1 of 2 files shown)


packages/diffs/src/editor/selection.ts
```diff
@@ -1644,6 +1676,39 @@ function findClusterBreak(
   return 0;
 }
 
+// Lists the start column of every grapheme cluster on a line (always starting
+// at 0). Used to step the caret and to transpose by whole graphemes so a
+// surrogate pair (emoji) or combining sequence is never split.
+function getLineGraphemeStarts(lineText: string): number[] {
+  const graphemeStarts = [0];
+  const segmenter = new Intl.Segmenter(undefined, { granularity: 'grapheme' });
```
fat, line 1684: Oh wow blessed firefox supports this now

fat (approved):
Seems chill to me - maybe worth testing on firefox still bc of the grapheme shit but still should work prob.

Also curious if this hurts perf at all

packages/diffs/src/editor/selection.ts
ije: Image

ije:
> Seems chill to me - maybe worth testing on firefox still bc of the grapheme shit but still should work prob.

in latest firefox, it works fine

> Also curious if this hurts perf at all

yes the pref will be affected when non-ascii chars present. we has to fallback to use `getBoundingClientRect` that even cases reflow. but i think it is acceptable since we cache(LUR) the dom measure result

PR #850, opened 2026-06-19
fix(diffs): Pluralize the "0 unmodified lines" hunk label

necolas:
A collapsed hunk separator shows how many unchanged lines are hidden. When that count was zero, the label read "0 unmodified line" instead of "0 unmodified lines".

The suffix added an "s" only when the count was greater than one, so zero fell through to the singular form. Select the form with the platform's English plural rules, which treat every count other than one as plural.


SlexAxton (approved):
Hell yea

amadeus (approved):

PR #859, opened 2026-06-19
fix(diffs): use current selection for selection action

necolas:
Steps to reproduce:
1. Open edit mode with the selection action enabled.
2. Drag to select several characters inside a word.
3. Click the gutter lightning-bolt icon, then the wrap button.
4. Only the first (or last) character is wrapped, not the whole selection.

The gutter icon is a cached DOM element reused across renders for the same line, but its click handler closed over the selection captured when the icon was first created. During a drag that is the first single-character selection (the first letter for a forward drag, the last for a backward drag), so the action ran against a stale range instead of the user's final selection.

Read the current primary selection at click time so the action always operates on what the user has selected. Add a regression test covering the forward and backward drag cases.


ije (approved):
LGTM!!!

PR #865, opened 2026-06-22
[diff/editor] Suppport custom clipboard provider

ije:
The browser's default `paste` event listener can freezing the renderer process. This pr added `clipboard` option for the Editor allows you to handle the `paste` event. Highly recommended to use native clipboard API if you are building an electron app.


necolas:
Needs a rebase as the beta branch got a force-pushed update

PR #867, opened 2026-06-22
fix(diffs): keep viewport colors correct after an offscreen delete

necolas:
Steps to reproduce:
1. Scroll a long file so the visible lines sit inside an unclosed block comment or template literal.
2. Delete the line directly above the viewport.
3. The visible lines lose their comment color and render as if they were plain code.

The tokenizer read the loop's grammar state before the offscreen flush rebuilt the cached state stack up to the viewport's first line, so it captured INITIAL. The visible lines were then tokenized as if outside the construct — corrected by a later background pass, or never when the viewport reaches the end of the document and no background pass is scheduled.

Seed the loop state after the offscreen flush instead. This reuses the state the flush already computed, so it adds no tokenization work and only changes the delete-reaches-viewport case; far-above edits with a gap stay seeded from INITIAL and are corrected by the background pass as before.

### DIFF
--- a/packages/diffs/src/editor/tokenzier.ts
+++ b/packages/diffs/src/editor/tokenzier.ts
@@ -305,7 +305,6 @@ export class EditorTokenizer {
     let line = canReuseCachedStates
       ? changedLineRanges[changedRangeIndex][0]
       : viewStart;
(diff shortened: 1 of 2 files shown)


ije (approved):
LGTM

PR #869, opened 2026-06-22
fix(diffs): align caret and selection on lines below a wrapping edit

necolas:
Turn on word wrap in an editable diff and type in a line until it grows long enough to wrap onto another row. Now click a line below it, or select text there: the caret and the selection highlight land a row too high, sitting on the wrapped line's extra row instead of the line you clicked.

The editor caches each line's vertical position and only refreshed it when the number of lines changed. Wrapping a line adds a visual row without changing the line count, so the cached positions of the lines below it stayed stale, and the caret and selection drawn from them rendered a row too high. Refresh the cached positions after an edit whenever wrap is on, matching how the wrap offsets are already invalidated.

This corrects only the vertical position of overlays on the lines below a wrap. Selecting a word on the wrapped line itself still mispaints the highlight horizontally; that is a separate bug.

### DIFF
--- a/packages/diffs/src/editor/editor.ts
+++ b/packages/diffs/src/editor/editor.ts
@@ -2990,10 +2990,13 @@ export class Editor<LAnnotation> implements DiffsEditor<LAnnotation> {
       onChange(fileRef, newLineAnnotations ?? this.#lineAnnotations);
     }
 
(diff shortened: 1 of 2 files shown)


ije (approved):
LGTM!

PR #870, opened 2026-06-22
perf(diffs): edit the piece table in place instead of rebuilding it

necolas:
Repro:

1. Open the editor on a large file, or make many scattered edits so the document accumulates many internal fragments.
2. Keep typing; each keystroke gets progressively slower.

Every edit (typing, deleting, applyEdits) rebuilt the editor's whole text structure from scratch, so a single edit cost time proportional to how fragmented the document had become; in a heavily edited file each keystroke paid for the entire document.

That structure (a piece table) keeps text as a list of "pieces" indexed by a tree. The tree is now a treap: a binary search tree that also keeps each node's random priority in heap order, so it stays balanced without an explicit rebalancing pass. An edit now splits the tree at the edit, drops the removed part, and merges the new text in, touching one root-to-leaf path (O(log P)) instead of rebuilding all P pieces. Reads and editor behavior are unchanged.

### DIFF
--- a/packages/diffs/src/editor/pieceTable.ts
+++ b/packages/diffs/src/editor/pieceTable.ts
@@ -45,11 +45,14 @@ class TextBuffer {
   }
 }
 
-// A node in the balanced piece tree.
+// A node in the balanced piece tree. `priority` keeps the tree balanced as a
+// treap (see PieceTable): the tree is a max-heap on priority while staying a
+// binary search tree on document offset.
 class PieceNode {
   left: PieceNode | null = null;
   right: PieceNode | null = null;
   parent: PieceNode | null = null;
+  priority = 0;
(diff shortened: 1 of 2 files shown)


packages/diffs/src/editor/pieceTable.ts
```diff
@@ -45,11 +45,14 @@ class TextBuffer {
   }
 }
 
-// A node in the balanced piece tree.
+// A node in the balanced piece tree. `priority` keeps the tree balanced as a
+// treap (see PieceTable): the tree is a max-heap on priority while staying a
```
ije, line 49: cool！maybe cleanup the jsdocs a bit 🤔

ije (approved):

PR #872, opened 2026-06-22
fix(diffs): keep edits visible when toggling a display option

necolas:
Steps to reproduce:
1. Open a diff in edit mode with word wrap on.
2. Edit a line, for example rename a function.
3. Toggle word wrap, or change any display option (theme, diff style, line numbers).

The edit disappears and the line shows its original text again. It only reappears after the next keystroke.

Changing a display option forces a full re-render that rebuilds the line DOM from the host's original file contents. The editor keeps its edited document across that re-render but never repainted the line text, so the view showed the pre-edit source while the model still held the edit.

After the re-render, repaint any editable row whose visible text no longer matches the editor document, re-tokenizing those lines from the document so the edit stays on screen.


ije (approved):

necolas:
The cacheKey fix works for the playground. My concern is that an editor host that doesn’t set a cacheKey on its files silently loses the visible edit on any display-option change. It looks like by using cacheKey the edits are kept alive inside the highlight render cache. So the edit staying on screen rides on (a) the host supplying a cacheKey and (b) that cache being retained across the re-render. cacheKey was documented as a worker-pool/caching hint, but it would also become a requirement for edits to persist. It's a coupling that's easy to miss, and nothing tests it.

If we’d rather the editor be correct regardless of host config, we can try reconciling the rendered rows against the `TextDocument` after a rebuild. I can revisit that approach to try and avoid re-renders when scrolling.

ije:
agree, no `cacheKey` guarantee is bad. i reset the changes.

necolas:
created a new PR which should avoid rendering on scroll https://github.com/pierrecomputer/pierre/pull/877

PR #873, opened 2026-06-23
Some editor nits and updates

mdo:
Copy edits, design updates, and more.


packages/diffs/src/editor/editor.css
```diff
@@ -149,14 +149,12 @@
...
   }
-  [data-selection-action] {
-    caret-color: currentColor;
-  }
   [data-content]:focus ~ [data-editor-overlay] [data-caret] {
     visibility: visible;
   }
 }
 [data-editor-widget] {
+  --diffs-widget-fg: var(--diffs-fg);
```
necolas, line 170: I think this is leaking into the marker severity popup, which relied on the fallback in `color: var(--diffs-widget-fg, #fff)` to render white text. The error/info/hint severity blocks set only `--diffs-marker-fg`, not `--diffs-widget-fg`. So that variable used to be `undefined` and fallback to white but now it resolves to `--diffs-fg` which is `light-dark(#000, #fff)`.

Repro:
1. Playground in light mode with markers enabled
2. hover an error/info/hint marker - you get black text on a saturated red/blue/gray chip instead of white. Dark mode and warning unaffected.

Before:
Image

After:
Image

If this isn't intended, I think the fix is to set the white explicitly in the severity rule (e.g. `[data-marker-popup][data-marker-severity] { color: #fff }` with warning keeping `#000`) so it no longer depends on the now-always-defined variable.

packages/diffs/src/editor/editor.ts
```diff
@@ -2469,112 +2480,75 @@ export class Editor<LAnnotation> implements DiffsEditor<LAnnotation> {
...
+        selectionActionElement,
+        overlayElement
+      );
     }
-    icon.style.transform = `translateY(${top}px) translateX(${left}px)`;
-    renderCtx.elements.set(cacheKey, icon);
+
+    // Anchor just below the selection's head line, mirroring the marker hover
+    // popover's geometry.
+    const [left, wrapLine] = this.#getCharX(head.line, head.character);
```
necolas, line 2632: This anchors at the selection head’s column with no horizontal clamping. The old gutter icon was pinned at the left margin and always visible.

Repro: select text whose head sits far right on a long, wrapped or non-wrapped line (worse for non-wrapped). The popover overflows the content’s right edge / needs horizontal scrolling to see.

Image

apps/docs/app/(diffs)/_examples/LiveEditing/LiveEditing.tsx
```diff
@@ -153,6 +160,26 @@ export function LiveEditing({
...
+              Edit mode
+            </div>
+          </Button>
+          <Switch
+            checked={mode === 'edit'}
+            onCheckedChange={(checked) => setMode(checked ? 'edit' : 'review')}
+            onClick={(e) => e.stopPropagation()}
+            className="pointer-events-none mr-3 place-self-center justify-self-end"
+            aria-label="Edit mode"
+          />
```
necolas, line 182: Is the Button and Switch doing the same thing?

apps/docs/app/(diffs)/_examples/LiveEditing/LiveEditing.tsx
```diff
@@ -153,6 +160,26 @@ export function LiveEditing({
...
+            <div className="flex items-center gap-2">
+              <IconPencil className="-ml-0.5" />
+              Edit mode
+            </div>
+          </Button>
+          <Switch
+            checked={mode === 'edit'}
+            onCheckedChange={(checked) => setMode(checked ? 'edit' : 'review')}
+            onClick={(e) => e.stopPropagation()}
+            className="pointer-events-none mr-3 place-self-center justify-self-end"
```
necolas, line 181: This can still be controlled via keyboard (pointer-events-none). Maybe this should be aria-hidden / tabIndex -1 so it's a visual-only thing?

apps/docs/app/(diffs)/_edit/HistoryDemo.tsx
```diff
@@ -276,47 +312,131 @@ export function HistoryDemo({ prerenderedFile }: HistoryDemoProps) {
...
 
   return (
     <div className="not-prose" ref={wrapperRef}>
+      <div className="mb-4 flex flex-wrap items-center gap-2">
+        <Button
+          variant="outline"
+          onClick={undoAll}
+          disabled={!canUndo}
+          aria-label="Reset to original"
+          title="Reset to original"
```
necolas: if the "undo all" text is now visible, we probably dont need the aria-label/title (which have different text)

packages/diffs/src/editor/editor.ts
```diff
@@ -2469,112 +2480,75 @@ export class Editor<LAnnotation> implements DiffsEditor<LAnnotation> {
...
+    if (this.#selectionAction === undefined) {
+      // The popover element is created once and kept open across selection
+      // changes, so its action handlers read the live primary selection rather
+      // than the snapshot taken at creation time (extending the selection by
+      // keyboard would otherwise leave them acting on the original range).
+      const getActiveSelection = (): EditorSelection =>
+        this.#selections?.at(-1) ?? primarySelection;
+      const selectionActionElement = renderSelectionAction({
+        textDocument,
+        selection: primarySelection,
```
necolas: This is selection frozen at popover-creation time. We should make selection a live getter too

packages/diffs/src/editor/editor.ts
```diff
@@ -2469,112 +2480,75 @@ export class Editor<LAnnotation> implements DiffsEditor<LAnnotation> {
...
-        // final selection. Read the current primary selection at click time so
-        // the action always operates on what the user actually has selected.
-        const activeSelection = this.#selections?.at(-1) ?? selection;
-
-        const cleanUp = () => {
+    if (this.#selectionAction === undefined) {
+      // The popover element is created once and kept open across selection
+      // changes, so its action handlers read the live primary selection rather
+      // than the snapshot taken at creation time (extending the selection by
+      // keyboard would otherwise leave them acting on the original range).
```
necolas: The "created once" comment hold for keyboard selection changes only; but a new drag seems to tear it down and pointerup recreates it

packages/diffs/src/editor/editor.ts
mdo: This might get fixed by #879, should be using same positioning. Might merge that and then review again.

ije: Screenshot 2026-06-24 at 18 01 46

ije: fixed in [b5068f3](https://github.com/pierrecomputer/pierre/pull/873/commits/b5068f3fe07c7da18f5b4506133cc930a1603c96)

apps/docs/app/(diffs)/_examples/LiveEditing/LiveEditing.tsx
mdo: Ya it's a stylistic thing—I use a CSS grid to stack the switch on the button. Not great, but _fine_ lol. We do this in a lot of places, so we could refactor at some point.

packages/diffs/src/style.css
```diff
@@ -238,6 +238,15 @@
...
   }
 
+  input,
+  button {
+    margin: 0;
+    padding: 0;
+    font-family: inherit;
+    font-size: inherit;
+    line-height: inherit;
+  }
```
necolas, line 246: This will zero padding on buttons inside the editor’s shadow DOM. The reset is in `@layer` base and editor chrome is injected unlayered, so search inputs/buttons (styled via `[data-input-box]`/`[data-search-icon]`) correctly override it. But a consumer `<button>` in the selection-action popover has no editor rule, so `button { padding: 0 }` replaces the UA default. Buttons in the popover will need explicit padding now. Worth checking is any buttons now render cramped.

necolas (approved):

PR #874, opened 2026-06-23
[diffs] Filename Header Suffix Slot API

amadeus:
Ability to slot in content after the file name in diffs/files/etc.

CleanShot 2026-06-23 at 12 45 07


amadeus:
@codex review

PR #877, opened 2026-06-23
fix(diffs): keep editor edits visible across host re-renders

necolas:
Steps to reproduce:
1. Open a diff in edit mode and edit a line (e.g. rename a symbol).
2. Toggle any display option - word wrap, theme, diff style, line numbers.
3. The edit vanishes and the line shows its original text again; it only reappears on the next keystroke.

When an editor is attached, its document is the source of truth for the content, but the host passes a static fileDiff. A display-option change forces a full re-render that rebuilds the diff rows from that fileDiff, so the in-progress edits are painted over with the original content - and inserted or deleted lines are lost.

After a full re-render, when the editor's document survived it and the rendered rows no longer match it, re-render the diff from the document: rerenderFromDocument re-derives the diff and clears the render cache so the rebuild re-highlights from the edited contents. One pass restores text, syntax colors, and line count, with no per-row reconciliation. The diff is mutated in place so its cacheKey is kept and the editor's document and undo history survive.

Gated to full re-renders (a scroll's partial render reuses the existing rows) and to components with a document-backed re-render (FileDiff; the plain File has no such path yet). Host-agnostic: this holds whether or not the host sets a cacheKey. Adds a regression test covering the toggle, further typing, an inserted line via both render APIs, and downstream block-comment highlighting - all with no cacheKey.

### DIFF
--- a/packages/diffs/src/editor/editor.ts
+++ b/packages/diffs/src/editor/editor.ts
@@ -591,13 +591,17 @@ export class Editor<LAnnotation> implements DiffsEditor<LAnnotation> {
       }
     }
 
-    if (
+    // Whether this sync replaces the document with a freshly parsed one (a new
+    // file, language, or cache key) versus reusing the existing one. A reused
+    // document keeps any edits the host's file contents do not have, which the
+    // rebuilt line DOM below must be reconciled against.
+    const documentReplaced =
       this.#textDocument === undefined ||
       this.#fileInfo === undefined ||
       this.#fileInfo.name !== fileOrDiff.name ||
       this.#fileInfo.lang !== fileOrDiff.lang ||
-      this.#fileInfo.cacheKey !== fileOrDiff.cacheKey
-    ) {
+      this.#fileInfo.cacheKey !== fileOrDiff.cacheKey;
+    if (documentReplaced) {
       let contents = '';
       if ('contents' in fileOrDiff) {
         contents = fileOrDiff.contents;
@@ -638,7 +642,16 @@ export class Editor<LAnnotation> implements DiffsEditor<LAnnotation> {
       }
     }
 
+    // A full re-render swaps in a new content element, so comparing identity
+    // detects one. This is reliable for FileDiff, which rebuilds the column
+    // (a new node) on a full render and reuses it on a partial one (scrolling).
+    // File reuses its content element in place, so this would not fire for a
+    // File full render - but File has no rerenderFromDocument path, so the
+    // re-render gate below never applies to it. If File ever gains one, this
+    // detection must be revisited.
+    let fullRerender = false;
     if (this.#contentElement !== contentEl) {
+      fullRerender = true;
       this.#gutterElement = gutterEl;
       this.#contentElement = extend(contentEl, {
         contentEditable: 'true',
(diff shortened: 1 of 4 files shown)


packages/diffs/src/editor/editor.ts
```diff
@@ -1832,6 +1853,78 @@ export class Editor<LAnnotation> implements DiffsEditor<LAnnotation> {
...
+      ) {
+        firstDriftedLine = lineIndex;
+      }
+    }
+
+    if (firstDriftedLine === Infinity) {
+      return;
+    }
+
+    const tokensByLine = tokenizer.tokenizeLines(firstDriftedLine, lastLine);
```
ije: Here i think we need to full tokenizing in background since the file contents have been reset in the File/Diff instance, you can follow the `#emitThemeChange` method that do full-tokenizing in bg and reneer lines that in current render range.

necolas:
OK I've tried reworking this one more time to fix at the library level. A lot of the problems come from trying to patch up the host copy. This commit makes the editor’s document the source of truth, so on a host-driven full re-render we now re-render the diff from the document instead of patching the rebuilt rows.

Assuming this latest iteration works, there are two options to compare:

* This approach (library-side): the editor owns the content and re-renders from it. No host changes; correct for any consumer. Cost: some editor <-> component plumbing (rerenderFromDocument + a divergence check) and a re-render on the toggle.
* Your cacheKey fix (host-side): ~2 lines, no editor changes but each editor host has to set a stable cacheKey, or it silently reverts.

I think you're best positioned to make the right call, and I'm fine with either approach. What do you think?

necolas:
Will ask @amadeus to take a look too since approach touches diffs

packages/diffs/src/components/File.ts
```diff
@@ -630,14 +627,16 @@ export class File<
 
     try {
       const pre = this.getOrCreatePreNode(fileContainer);
-      if (
+      // A full render rebuilds the line DOM from `file`; a partial render only
+      // shifts the rendered window. The editor needs to know which happened so
+      // it repaints edited rows only after a full rebuild.
+      const didFullRender =
```
amadeus: Wonder about the name of this variable, since calculating it won't have performed the full render, the full render happens on line 639. 

So in some ways it means, partialRenderFailed or `shouldFullRender` (not names i am advocating for, just trying to explain the logic.

Or we should invert the value like FileDiff and do `didPartiallyRender` and remove the two `!`  and make it an `&&`

ije:
i added a small change that removes `contentRebuilt` argument. we can just simple check whether the `content` element is changed.

ije (approved):

packages/diffs/src/components/File.ts
necolas: Good catch. That `didFullRender` (and its comment) were leftovers from the `contentRebuilt` arg we just removed — its only consumer was the `syncRenderViewToEditor(didFullRender)` call. Since `File` isn't on the document-re-render path, I dropped the variable and inlined the condition, so `File.render` is back to its upstream form (no diff on `File.ts` now). Sidesteps the naming question entirely — thanks for flagging it.

PR #878, opened 2026-06-23
[diffs/editor] Fix diff rerener in edit mode

ije:
Fixes diff re-rendering while the editor is active. Edits were causing jitter, broken caret/selection rendering, incorrect scroll positioning, and stale unified-diff rows because refresh logic, render-range bookkeeping, and document-boundary handling did not account for how diff views virtualize and recompute hunks.

**Rendering & refresh**
- Split unified and split refresh paths (`refreshUnifiedDiffView` / `refreshSplitDiffView`) and debounce both at 150ms after edits.
- Unified edits now refresh row spans, hunk separators, and manager state — not just gutter/content HTML.
- `applyDocumentChange` updates diff metadata in place instead of forcing a full re-render on every keystroke.

**Editor state & cache**
- Remove the `rerenderFromDocument` / DOM-divergence path. The editor now defaults `cacheKey` to the file name (or `prevName->name` for renames) so editing state survives scroll-driven partial re-renders without walking the DOM. @necolas 
- `isLineVisible` checks for a rendered line element instead of relying on `renderRange`, which broke caret/selection overlays in virtualized diffs.
- Defer caret `scrollIntoView` to the next animation frame to reduce scroll jitter when adding/removing lines.
- Fix `moveCursorToDocEnd` / `getDocumentBoundarySelection` for diff documents with trailing blank lines, no final newline, or a single-line buffer.

**Virtualization & hunk recompute**
- Add `getLayoutLineCount` so render-range clamping uses the live diff size even when layout checkpoints are stale after edits.
- Preserve `renderRange` across in-edit layout resets; recompute buffer spacers when the document grows below the rendered window.
- Add `recomputeDiffHunksForEdit` with top-aligned addition sentinels so delete-all / sparse-blank edits keep editable rows at the top of split view and retain a caret host row for empty documents.


necolas:
The codex-flagged issues look real to me

ije:
@necolas added `renderSeparators`

necolas:
Sorry this needs another forced rebase onto the latest beta branch

packages/diffs/src/editor/editor.ts
```diff
@@ -1990,42 +1934,6 @@ export class Editor<LAnnotation> implements DiffsEditor<LAnnotation> {
...
-      const lineIndex = lineNumber - 1;
-      if (
-        lineIndex >= textDocument.lineCount ||
-        el.textContent !== textDocument.getLineText(lineIndex)
-      ) {
-        return true;
-      }
-    }
-    return false;
-  }
```
ije, line 2051: now the editor uses the file name as the cache key if not provided, so we don't need to walk the dom when scrolling triggers rerender.

packages/diffs/src/editor/editor.ts
```diff
@@ -3588,22 +3509,25 @@ export class Editor<LAnnotation> implements DiffsEditor<LAnnotation> {
...
-        // to mock the 'contenteditable' behavior
         if (this.#primaryCaretElement !== undefined) {
-          this.#primaryCaretElement.scrollIntoView({
-            block: 'nearest',
-            inline: 'nearest',
+          requestAnimationFrame(() => {
+            this.#primaryCaretElement?.scrollIntoView({
+              block: 'nearest',
+              inline: 'nearest',
+            });
```
ije, line 3576: apply scroll at next frame to fix the scroll jitter

packages/diffs/src/editor/editor.ts
```diff
@@ -3941,16 +3857,6 @@ export class Editor<LAnnotation> implements DiffsEditor<LAnnotation> {
...
-    }
-    const { startingLine, totalLines } = this.#renderRange;
-    if (line < startingLine) {
-      return false;
-    }
-    if (totalLines === Infinity) {
-      return true;
-    }
-    return line < startingLine + totalLines;
+    return this.#getLineElement(line) !== undefined;
```
ije, line 3909: let's check if the line element exists, use renderRange may break the selection rendering in diff view

packages/diffs/src/editor/selection.ts
```diff
@@ -1369,9 +1369,12 @@ export function getDocumentFullSelection(
...
   textDocument: TextDocument<unknown>,
-  atEnd: boolean
+  atEnd: boolean,
+  endNewLineTrimmed?: boolean
 ): EditorSelection {
-  const line = atEnd ? textDocument.lineCount - 1 : 0;
+  const line = atEnd
+    ? textDocument.lineCount - (endNewLineTrimmed === true ? 2 : 1)
+    : 0;
   const character = atEnd ? textDocument.getLineLength(line) : 0;
```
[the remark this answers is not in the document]
ije: fixed

packages/diffs/test/editorDisplayOptionResync.test.ts
```diff
@@ -161,14 +167,15 @@ describe('diff editor: display-option toggle mid-edit', () => {
...
+      // A forced re-render briefly rebuilds rows from the host's diff source,
+      // then re-renders from the editor document so the edit remains visible.
       expect(lineText(container, 1)).toBe('alphaX');
     } finally {
       await fixture.cleanup();
     }
   });
 
-  test('accepts further typing after the toggle without duplicating the edit', async () => {
+  test('accepts further typing after the toggle from the rebuilt line text', async () => {
```
necolas: AI flagged this test as failing in isolation because the outcome of a display-option toggle mid-edit depends on whether the shared highlighter is cold or warm:

```
bun test packages/diffs/test/editorDisplayOptionResync.test.ts -t "accepts further typing"
→ 0 pass / 1 fail   (got "alphaYX\nCHANGED\n", expected "alphaY\nCHANGED\n")
```

The file only passes because the order matches: Test 1 runs cold and asserts the edit is preserved (alphaX); Test 2 runs warm and asserts the edit is gone (typing yields alphaY). The two tests and their comments assert opposite behaviors:

* Test 1: "re-renders from the editor document so the edit remains visible" → alphaX
* Test 2: "the view shows the pre-edit source … applies against that rebuilt text" → alpha

`documentReplaced` is gated on `cacheKey`, and the `cacheKey` defaulting races the async highlight that drives `syncRenderViewToEditor`. This is the full-re-render (toggle) case, i.e., the original "edit reverts on word-wrap toggle" bug. In the warm case (typical once the app has rendered any file), the in-flight edit is reverted on a display-option toggle.

The suggested fix is to decide the single intended behavior for toggle-mid-edit (preserve vs. revert) and make it deterministic regardless of highlighter warmth; then switch the resync suite to `beforeEach(disposeHighlighter)` (as `editorDiffEmptyDocument.test.ts` already does) and assert that one behavior.

packages/diffs/src/components/VirtualizedFile.ts
```diff
@@ -626,44 +626,13 @@ export class VirtualizedFile<
   // normally triggered by the editor when the document line count changes
   override applyDocumentChange(
     textDocument: DiffsTextDocument,
-    newLineAnnotations?: LineAnnotation<LAnnotation>[],
-    shouldUpdateBuffer = false
```
necolas, line 630: We lost the `shouldUpdateBuffer` parameter here and the buffer-spacer recompute, while `VirtualizedFileDiff` keeps and enhances it. The editor still passes the flag as the 3rd arg so for a non-diff editable File, an edit that grows the document below the rendered window no longer updates the buffer before re-render, which might cause a scroll/caret-position regression.

packages/diffs/src/utils/updateDiffHunks.ts
```diff
@@ -47,26 +47,63 @@ export function recomputeDiffHunks(
...
+export function shouldTopAlignAdditionRecompute(
+  diff: FileDiffMetadata,
+  additionLines: string[]
+): boolean {
+  return (
+    additionLines.length > 0 &&
+    additionLines.length < diff.deletionLines.length &&
+    hasOnlyBlankAdditionContents(additionLines)
+  );
+}
```
necolas, line 92: Changing delete-all behavior might clash with https://github.com/pierrecomputer/pierre/pull/907

packages/diffs/src/components/FileDiff.ts
```diff
@@ -1081,17 +1071,17 @@ export class FileDiff<
...
+        clearTimeout(this.refreshDiffViewTimeout);
       }
+      this.refreshDiffViewTimeout = setTimeout(() => {
+        this.refreshDiffViewTimeout = undefined;
+        if (this.options.diffStyle === 'split') {
+          this.refreshSplitDiffView();
+        } else {
+          this.refreshUnifiedDiffView();
+        }
+      }, 250);
```
necolas: Did you mean to change the timeout interval from 100 to 250?

packages/diffs/src/components/VirtualizedFile.ts
ije: i removed this, currently we do full rerender when `applyDocumentChange` is called, that updates the buffers. `File` component still uses the `shouldUpdateBuffer` argument since the editor managed the rerender

packages/diffs/src/components/FileDiff.ts
ije: yep, now refresh unified layout also debounced

packages/diffs/src/utils/updateDiffHunks.ts
ije: hmm, can we merge this pr first? 🤔

packages/diffs/test/editorDisplayOptionResync.test.ts
ije: fixed in 185b81c, testing is restored

PR #888, opened 2026-06-24
[cleanup] remove unused/unsupported packages (truncate, storage-elements)

SlexAxton:
none of these are actively developed or used anywhere. we can go find them in git history if we ever need them again.


amadeus (approved):
🫂

PR #893, opened 2026-06-25
Add additional mts and cts typescript extensions

scaryrawr:
Node supports running typescript natively.

By default (from [nodejs docs](https://nodejs.org/api/typescript.html#determining-module-system)):

- `.ts` files use the package.json type
- `.cts` files run as CommonJS
- `.mts` files run as ES modules


ije (approved):
LGTM

scaryrawr:
Thanks @ije ! I saw some PRs target a beta branch. I'm wondering if I should be requesting into something other than `main`? 😅

ije:
i guess is fine on main @amadeus

amadeus:
main is fine for a change like this!

PR #897, opened 2026-06-26
[docs] Sync diff editor with app theme in system mode

necolas:
The diff editor's "system" themeType follows the OS prefers-color-scheme (its shadow root declares `color-scheme: light dark`), so it drifted from the app whenever the app theme differed from the OS. Resolve "system" to the app's scheme from `@pierre/theming` so the editor stays in sync; the switcher stays editor-only and "light"/"dark" remain independent overrides.

Rename the `useTheme()` binding and every consumer to the controller's vocabulary so the docs match the library, not just one file:

```
resolvedTheme -> resolvedColorScheme
theme         -> colorMode
setTheme      -> setColorMode
systemTheme   -> systemColorScheme
themes        -> colorModes
```

Also rename the playground's local switcher state (`themeType` -> `colorMode`, the diffs `themeType` option key is unchanged) and the pre-paint bootstrap script's locals. Pure rename — no behavior change.

### DIFF
--- a/apps/docs/app/(diffs)/playground/PlaygroundClient.tsx
+++ b/apps/docs/app/(diffs)/playground/PlaygroundClient.tsx
@@ -94,7 +94,7 @@ type EditorMode = 'review' | 'edit';
 // Default values for URL param comparison
 const DEFAULTS = {
   diffStyle: 'split',
-  themeType: 'system',
+  colorMode: 'system',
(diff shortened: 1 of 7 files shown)


ije (approved):
nice!

mdo (approved):
Woohoo!

PR #900, opened 2026-06-26
fix(diffs): move focus when a diff line number is clicked

### LINKED ISSUE #898: Line-number clicks prevent moving focus
### Package

@pierre/diffs 1.2.4, React `CodeView`

### Summary

We need to wrap the diff in a focusable element is so that we can implement keyboard navigation for the diff selection, similar to lazygit:

https://github.com/user-attachments/assets/3b307593-8c6d-4e52-98c6-036083e1c045

However, currently we have an issue whereby clicking a diff line number updates the selected line range _but focus does not move into the diff area_. Observe `document.activeElement` here:

https://github.com/user-attachments/assets/f06e86f3-e8d1-4fcd-aeed-e432e714aa78

### What seems to cause it

Naturally I would expect click to bubble up to the focusable element that surrounds the diff, but it doesn't because of the `event.preventDefault()` in `pointerdown` here:

https://github.com/pierrecomputer/pierre/blob/0d40be4920992b2a1fb0a64153970dcbdb01f22a/packages/diffs/src/managers/InteractionManager.ts#L761

Preventing default on `pointerdown` cancels the browser's native focus update.

### Reproduction

1. Render `CodeView` inside a focusable host wrapper.
2. Enable line selection.
3. Focus some other element.
4. Click a diff line number.

necolas:
Steps to reproduce:

1. Wrap a CodeView in a focusable element (the host wrapper).
2. Enable line selection.
3. Focus some other element on the page.
4. Click a diff line number.

The line range is selected, but focus never moves into the diff, so the host wrapper can't receive the keyboard events it needs to drive navigation of the selection.

The line-selection pointerdown called preventDefault, which also cancels the browser's native focus change. Line numbers already use `user-select: none` and `touch-action: none`, so that preventDefault only blocked focus. Each `pre` also had `tabindex="0"`, so any focus that did move landed on the `pre` instead of the wrapper, even though the `pre` is not the scroll container and has no keyboard behavior.

Remove both. A click now moves focus to the nearest focusable ancestor, the host wrapper, which can own keyboard navigation.

Fix #898

### DIFF
--- a/packages/diffs/src/utils/setWrapperNodeProps.ts
+++ b/packages/diffs/src/utils/setWrapperNodeProps.ts
@@ -55,7 +55,6 @@ export function setPreNodeProperties(
     pre.removeAttribute('data-diff-type');
   }
   pre.setAttribute('data-overflow', overflow);
-  pre.tabIndex = 0;
   // Set CSS custom property for line number column width
   pre.style.setProperty(
     '--diffs-min-number-column-width-default',
--- a/packages/diffs/src/managers/InteractionManager.ts
+++ b/packages/diffs/src/managers/InteractionManager.ts
@@ -758,7 +758,11 @@ export class InteractionManager<TMode extends InteractionManagerMode> {
       return;
     }
 
-    event.preventDefault();
+    // Intentionally no preventDefault here. The line-number gutter uses
+    // `user-select: none`, so a drag that starts on it never begins a native
+    // text selection. Calling preventDefault would also cancel the browser's
+    // native focus change, which stops a focusable host wrapper from receiving
+    // focus (and the keyboard events it needs) when a line number is clicked.
     const { lineNumber, eventSide, lineIndex } = pointerInfo;
 
     if (event.shiftKey && this.selectedRange != null) {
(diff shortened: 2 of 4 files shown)


packages/diffs/src/utils/setWrapperNodeProps.ts
```diff
@@ -55,7 +55,8 @@ export function setPreNodeProperties(
     pre.removeAttribute('data-diff-type');
   }
   pre.setAttribute('data-overflow', overflow);
-  pre.tabIndex = 0;
+  // The pre is intentionally not focusable; see createPreElement for why.
+  pre.removeAttribute('tabindex');
```
amadeus: We probably don't need this if we are fixing the create logic?

amadeus (approved):
🙏 thanks for jumping on this. A lot of the current stuff was either shit i copied over from old pierre diffs or just stuff i didn't think very deeply about.

If you have other good ideas on stuff that we should fix for a11y/keyboard stuff, def feel free to ad more fixes

packages/diffs/src/utils/setWrapperNodeProps.ts
necolas: fixed

amadeus (approved):

PR #903, opened 2026-06-26
fix(diffs): keep merge conflict actions working after resolving one

necolas:
In the merge conflict view, accepting one conflict (current, incoming, or both) updates the diff. Afterward every remaining "Accept" button stops responding: clicking any of them does nothing.

UnresolvedFile wires its merge-conflict-action click handler into the InteractionManager, but the shared FileDiff.syncInteractionOptions() helper rebuilt those options without that handler. The helper re-runs on every render and during hydration, so resolving a conflict re-renders the diff and drops the handler.

Override syncInteractionOptions() in UnresolvedFile to keep the handler attached across re-renders, and route setOptions() through it so the two paths can no longer drift apart.

### DIFF
--- a/packages/diffs/src/components/UnresolvedFile.ts
+++ b/packages/diffs/src/components/UnresolvedFile.ts
@@ -158,15 +158,22 @@ export class UnresolvedFile<
 
     this.options = options;
     this.hunksRenderer.setOptions(this.getHunksRendererOptions(options));
+    this.syncInteractionOptions();
+  }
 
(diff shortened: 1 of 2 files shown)


amadeus:
I wonder when this broke, cause it definitely wasn't when we shipped it 🤔

amadeus (approved):
looks like i broke this here: https://github.com/pierrecomputer/pierre/pull/718

thanks for the fix!

PR #904, opened 2026-06-26
fix(docs): reset agent demo scroll when switching files

necolas:
On the homepage editor demo, scroll a changed file down, then open a different file from the Changes list: it opens scrolled to the previous file's offset instead of at the top.

The demo reuses one FileDiff surface across files (never remounted, so the prerendered markup survives), and the scroll lives on the host `.aui-surface-wrap` container, which nothing reset on a file switch. Reset its scrollTop to 0 whenever the active file changes, in a layout effect so the new diff never paints at the old offset.


ije (approved):
🙏

PR #906, opened 2026-06-29
[diffs/editor] Fix some selection bug

ije:
- Fix cursor movement handling for backward multi-line selections
- Fix selection merging for multiple cursors with `shift` key pressed

### DIFF
--- a/packages/diffs/src/editor/selection.ts
+++ b/packages/diffs/src/editor/selection.ts
@@ -131,16 +131,14 @@ export function mapCursorMove(
       shortcut === 'start' ||
       shortcut === 'end'
     ) {
+      const caret = getCaretPosition(selection);
+      line = caret.line;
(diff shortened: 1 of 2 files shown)


necolas:
LGTM, nice fix.

I pushed a test-only commit (70ccda0) on top of your branch; no source changes.

While reviewing, the new direction handling in `mergeOverlappingSelections` had two branches that weren't exercised by tests:

- the most-recent overlapping selection is already **backward** → the merged range should keep that backward direction (the `direction = latest.selection.direction` path)
- a later **bare caret** lands on the merged start → the direction is *derived* as backward (the `DirectionNone` re-derivation path)

Both exist so the merged caret stays where it was most recently placed, so I added one test for each to pin that focus-preservation behavior down.

necolas (approved):

PR #907, opened 2026-06-29
fix(diffs): keep editing working after deleting a diff's new side

necolas:
To reproduce:

1. Open a diff in the editor and switch to edit mode.
2. Select everything on the new-file side and delete it.
3. Type a character, press Enter, and keep typing.

The caret jumps to the top of the file and then to the bottom instead of staying on the line being edited. In a unified diff the empty editable row also renders on top of a deletion line, newly typed lines fail to render or land on top of one another, and switching from split to unified after the delete throws "deletionLine and additionLine are null".

The editor treats a document that is empty or ends in a newline as having a trailing empty line the caret can sit on, but the diff drops that line: splitFileContents collapses it during the recompute and cleanLastNewline drops it again during re-highlight. The editable column then has one fewer row than the editor's line count, so the caret line has no element (drawn at the top) and a line typed onto it is appended after the deletion buffer (drawn at the bottom). In unified the dropped row also has no content slot, which crashed the renderer.

Render an editable row for the trailing line and rebuild the diff when an edit adds or removes it, place the caret from the previous line when a row has no element yet, and render an empty row instead of throwing when the highlight lacks a row the diff still emits. This replaces the empty-document sentinel, which matched a blank deletion line and drew the editable row in the middle of the file.

### DIFF
--- a/packages/diffs/src/editor/editor.ts
+++ b/packages/diffs/src/editor/editor.ts
@@ -213,6 +213,13 @@ export class Editor<LAnnotation> implements DiffsEditor<LAnnotation> {
   #searchPanel?: SearchPanelWidget;
   #selectionAction?: SelectionActionWidget;
   #shouldIgnoreSelectionChange = false;
+  // Set by select-all, cleared on the next keydown/pointerdown. Select-all
+  // mirrors the whole document onto the native selection so WebKit fires a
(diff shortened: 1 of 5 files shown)


necolas:
@codex

necolas:
Still working on this...

ije:
i think i already fixed this in #878 LOL

PR #909, opened 2026-06-29
Blend the edit mode backgrounds

mdo:
This adds some functionality to blend the selection color of edit mode with the diff line background.

### DIFF
--- a/packages/diffs/src/editor/editor.ts
+++ b/packages/diffs/src/editor/editor.ts
@@ -2866,7 +2866,18 @@ export class Editor<LAnnotation> implements DiffsEditor<LAnnotation> {
       radius: 'rtl' | 'rbl' | 'rbr'
     ) => {
       const top = this.#getLineY(line) + wrapLine * lineHeight;
-      const css = `width:${ch}px;transform:translateX(${left}px) translateY(${top}px);`;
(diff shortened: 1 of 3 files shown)


packages/diffs/src/editor/editor.ts
```diff
@@ -3655,6 +3662,25 @@ export class Editor<LAnnotation> implements DiffsEditor<LAnnotation> {
     return undefined;
   }
 
+  // Painted background color of a line, read from the [data-line]::after layer
+  // (the line element itself is transparent in edit mode). Returns undefined when
+  // that layer is transparent (e.g. context lines).
+  #lineBackgroundColor(line: number): string | undefined {
```
ije, line 3705: is it possible to use color-mix in editor.css instead of dynamically adding `--diffs-selection-corner-bg`?

mdo: I don't think so, because the selection is on another layer, so it has no access to the line color. `data-editor-overlay` is where the selection range lives. That said, I do think we'll need to cache some of the values if we decide to go this route.

necolas (approved):
lgtm once agents are appeased

ije (approved):
LGTM

PR #910, opened 2026-06-29
fix(diffs): Refresh reshaped split edits safely

necolas:
Fixes split-diff edit-mode rerendering issues surfaced by the Codex review on this branch.

**Reproduction**
1. Open a split editable diff (e.g. `/playground` with Edit mode).
2. Edit an added line back to the old-side text so its hunk disappears or changes shape.

Before this change the editor stayed on the split fast-refresh path, which only patches row attributes — leaving stale added/deleted rows in the DOM.

**Changes**
- Detect when recomputing content hunks changes the rendered hunk shape, and use a full refresh (clearing the render cache) for that case instead of the in-place fast patch.
- Reset the editor's caret/selection geometry caches when a split content edit rebuilds rows, so the caret doesn't measure a detached row and jump.
- `fastRefreshDiffView` self-escalates to a full rebuild when a column's rendered row count diverges, instead of silently leaving stale rows.
- Guard the async editor-sync continuation so a highlighter promise from an older render can't resync a detached or replaced editor.
- Document `getDiffRenderShapeKey` (the fast-vs-full gate) and add regression coverage: split reshape that removes a hunk, a same-split-count reshape, and fast-path reachability.

**Scope note**
Rebased to drop the "preserve editor history across layout toggles" commit — that workaround is being handled via the cacheKey approach in #878. This PR is now only the split-reshape / async-sync correctness fix.

### DIFF
--- a/packages/diffs/src/components/File.ts
+++ b/packages/diffs/src/components/File.ts
@@ -178,6 +178,7 @@ export class File<
   protected enabled = true;
 
   protected editor: DiffsEditor<LAnnotation> | undefined;
+  private renderViewSyncVersion = 0;
 
   constructor(
     public options: FileOptions<LAnnotation> = { theme: DEFAULT_THEMES },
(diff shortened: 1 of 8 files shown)


necolas:
Still working on this...

apps/docs/app/(diffs)/playground/PlaygroundClient.tsx
```diff
@@ -667,12 +667,10 @@ export function PlaygroundClient({ prerenderedDiff }: PlaygroundClientProps) {
...
-  // The editor attaches to the diff's editable (new-file) side. Recreate it
-  // when the diff layout or edit mode changes so it re-attaches to the freshly
-  // relaid-out surface with a clean document instead of reusing a torn-down
-  // instance (mirrors LiveEditing's editor lifecycle).
-  // eslint-disable-next-line react-hooks/exhaustive-deps -- deps intentionally force a fresh editor; the factory takes no inputs
-  const editor = useMemo(() => new Editor({}), [diffStyle, editorMode]);
+  // Keep the editor instance stable across display-option changes. FileDiff
+  // re-syncs the attached editor after layout re-renders; recreating it here
+  // would drop the TextDocument undo/redo stack on split/unified toggles.
+  const editor = useMemo(() => new Editor({}), []);
```
ije: nice

ije (approved):
LGTM

ije:
@necolas #878 will ensure the cacheKey with the file name if it's not provided. we better should not add workarounds for the layout/options update. in further update, we need to cache TextDocuments using the cacheKey. so when switching file the undo stack remains.

i also updated the playground/home apps in #878, please take a look

PR #922, opened 2026-06-30
fix(diffs): place selection action above bottom-up selections

mdo:
Render the selection-action popover above the selection for backward (bottom-up) selections instead of covering its first line, and preserve the selection direction across a refocus so the caret/popover no longer jump to the bottom.


necolas (approved):
lgtm once agents are appeased

mdo:
@codex review

packages/diffs/src/editor/editor.ts
```diff
@@ -185,10 +196,29 @@ export class Editor<LAnnotation> implements DiffsEditor<LAnnotation> {
...
+  // and re-running getComputedStyle on every render.
+  #scrollContainer?: HTMLElement;
+  #scrollContainerSource?: HTMLElement;
+  // Cached screen-space rects for the selection-action popover's viewport
+  // check, refreshed only on scroll/resize (rather than on every selection
+  // update) since #updateSelectionActionPopover can run once per keystroke.
+  #cachedCodeRect?: DOMRect;
+  #cachedScrollContainerRect?: DOMRect;
+  #viewportRectsDirty = true;
+  #viewportRectListenersDisposes?: (() => void)[];
```
ije: so many states added to editor for the action UI improve, may be move these to slectionAction.ts?

mdo: Yeah I think this PR stays on hold for a bit, it's not an urgent thing to get into the release. After the other major branches merge, I'll swing back to this.

mdo:
Follow-up from the reviews here and latest `beta-1.3` changes—this extends the edge-avoidance logic to the marker hover popup, fixes a virtualization edge case, and tightens things up a bit.

**What changed**
- Pulled `choosePopoverPlacement` out into a shared `popoverPlacement.ts` and hooked the marker hover popup up to it too, so hovering a squiggle near the bottom of the screen now flips the popup above instead of clipping off-screen.
- The selection-action popover could try to anchor its fallback position to a line that isn't even rendered (outside the virtualized window), which produced garbage coordinates for long backward selections. Now it just checks visibility first.
- Added a little hysteresis (4px) to the flip decision so a selection sitting right on the boundary doesn't flicker between above/below on every keystroke.
- The popover now re-checks its placement on scroll/resize instead of staying stuck on whatever side it picked last.
- Double-checked that IME composition doesn't break the direction-preservation fix — turns out it's already covered.

**Bonus find:** the existing `refocus keeps a backward selection backward` test wasn't worth it — it passed even with the real fix ripped out, because leftover focus-handling state was silently swallowing the test's event before it ever hit the code being tested. Fixed that too.

Added tests for all of the above (marker popup placement, the fallback-visibility fix, hysteresis, composition). Full `diffs` suite (804 tests) + lint/format/typecheck all green.

ije:
Screenshot 2026-07-02 at 08 53 15
The selection action widget position on a backward selection shifted 2 lines


**update**: fixed in 51512b6

packages/diffs/src/editor/editor.ts
ije: moved these states to  `PopoverManager` class in 42c3e9d

mdo:
@codex review

packages/diffs/src/editor/selectionAction.ts
```diff
@@ -44,19 +53,48 @@ export class SelectionActionWidget {
...
     );
+    // Measured synchronously once so the very first reposition (right after
+    // construction) already has an accurate height instead of waiting for the
+    // ResizeObserver's async callback; later size changes (e.g. the
+    // consumer's content changing while the selection stays open) are picked
+    // up reactively below.
+    this.#height = this.#popover.offsetHeight;
+    this.#resizeObserver = new ResizeObserver(() => {
+      this.#height = this.#popover.offsetHeight;
+    });
```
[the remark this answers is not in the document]
ije: fixed in 471928e

ije (approved):
this works great to me

PR #924, opened 2026-06-30
Fix reset button on live editing example

mdo:
Reset is broken currently on the /edit page first example.

### DIFF
--- a/packages/diffs/src/editor/editor.ts
+++ b/packages/diffs/src/editor/editor.ts
@@ -557,14 +557,11 @@ export class Editor<LAnnotation> implements DiffsEditor<LAnnotation> {
     // read-only surface would keep showing the editor's stale selected line.
     this.#resetState();
 
-    // Drop the attached document so a re-attach rebuilds from the host's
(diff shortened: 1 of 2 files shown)


packages/diffs/src/editor/editor.ts
```diff
@@ -543,7 +543,21 @@ export class Editor<LAnnotation> implements DiffsEditor<LAnnotation> {
...
     this.#resetState();
+
+    // Drop the attached document so a re-attach rebuilds from the host's
+    // contents rather than reusing the edited buffer. The document survives host
+    // re-renders (option toggles) but not a full detach, so remounting the
+    // surface on the same editor discards prior edits - how the docs "Live
+    // editing" Reset reverts to the original.
+    this.#fileInstance = undefined;
+    this.#fileInfo = undefined;
+    this.#textDocument = undefined;
```
necolas: this is very similar to the change in https://github.com/pierrecomputer/pierre/pull/878/changes/f52f0ff6700af80541f57112b5620e1a1a1ffc5c

necolas (approved):

PR #929, opened 2026-07-01
[docs/diffs]: Update agent UI component

ije:
- “Add to chat” captures the live editor selection when clicked, then renders an attachment header like filename (start-end) with a file icon.
- Large snippet attachments are scrollable inside their code block.


mdo (approved):
Nice, thanks!

PR #937, opened 2026-07-02
fix(diffs): Preserve annotations across line edits

necolas:
1. Open a diff editor with an annotation on an added line.
2. Delete that line, or delete the newline immediately above it.
3. Undo and redo the edit, or insert another line above it.
4. The annotation disappears or fails to return with history.

The annotation mapper only used the aggregate line delta for an edit batch. That made it guess deleted ranges, drop annotations on deleted lines, and skip net-zero batches where one edit added lines and another removed them.

Record per-edit line changes on TextDocumentChange and map annotations through each line edit in order. Keep annotations by moving them to the nearest surviving line, and refresh touched annotation lists so undo and redo restore the right annotation state.


necolas:
https://github.com/user-attachments/assets/c2e9c600-3a6c-4c94-89d6-ebd6ac67ffe5

ije (approved):
lgtm

PR #951, opened 2026-07-09
fix(diffs): Keep redo correct after typing at an insert's left edge

necolas:
Type a character, move the caret to the far left with an arrow key, then type another character. The buffer shows the right result (for example "ba"), and undo clears it, but redo puts the characters back in the wrong order ("ab"). Any consumer reading the document contents then sees the corrupted text.

Consecutive inserts merge into one undo step so a run of typing undoes at once. The merge test only checked that the new insert mapped back to the same base offset as the previous one. For a pure insert both edges of the inserted text map back to that single offset, so typing before the previous character looked identical to continuing after it; the two inserts merged as if appended, but the buffer held them reversed, and redo replayed the merged run in append order.

Require the next insert to begin exactly at the end of the previously inserted text in after-edit offsets (the end of the previous inverse edit) before merging, so a caret that jumped left starts a new undo step instead.

### DIFF
--- a/packages/diffs/src/editor/editStack.ts
+++ b/packages/diffs/src/editor/editStack.ts
@@ -212,10 +212,14 @@ export function shouldCoalesceEditStackEntry<LAnnotation>(
       !nextForward.text.includes('\n') &&
       nextInverse.text.length === 0;
     if (previousWasInsert && nextIsInsert) {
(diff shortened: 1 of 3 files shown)


ije (approved):
nice catch

PR #952, opened 2026-07-09
fix(diffs): Clear match highlights when the search is dismissed

necolas:
Reproduce (empty query):

1. Open a file in the editor and press Cmd/Ctrl+F.
2. Type a matching term, e.g. "foo"; its matches highlight.
3. Delete the query so the search box is empty.

Wrong: the highlights stay painted over an empty box.

Reproduce (Escape from the content):

1. Press Cmd/Ctrl+F and search a term so its matches highlight.
2. Click into the editor text, then press Escape.

Wrong: the panel closes but the highlights stay painted.

Emptying the query returned from updateMatches before calling onUpdate([]), the only channel that clears the editor's #matches and repaints, so the highlights were never dropped. Fold the empty-query case into the existing zero-match branch, which already calls it.

The content-focused Escape handler tore the panel down with cleanup(), which removes the panel DOM but skips the onClose teardown. Add a close() method (cleanup plus onClose) and call it from that handler so #matches is cleared and the overlay repaints, matching the close button and the search input's own Escape.

### DIFF
--- a/packages/diffs/src/editor/editor.ts
+++ b/packages/diffs/src/editor/editor.ts
@@ -1262,7 +1262,7 @@ export class Editor<LAnnotation> implements DiffsEditor<LAnnotation> {
       addEventListener(contentEl, 'keydown', (e) => {
         if (e.key === 'Escape') {
           e.preventDefault();
-          this.#searchPanel?.cleanup();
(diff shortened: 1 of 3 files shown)


ije (approved):
ltgm, thanks

PR #954, opened 2026-07-09
fix(diffs): Move wrapped carets by visual rows

necolas:
Reproduce the bug:
1. Open the diffs editor with word wrap enabled.
2. Put the caret on the middle visual row of one long logical line that wraps onto at least three rows.
3. Press ArrowDown.
4. The caret jumps to the next logical line instead of the next wrapped visual row. Cmd+Left, Cmd+Right, Ctrl+A, Ctrl+E, Home, and End also use logical line boundaries instead of the current visual row.

The editor already measured wrap offsets for caret rendering and soft-line deletion, but cursor movement did not receive those offsets. Pass wrap offsets into the selection movement mapper when wrap mode is active, and use them for vertical moves and line-boundary shortcuts.

### DIFF
--- a/packages/diffs/src/editor/editor.ts
+++ b/packages/diffs/src/editor/editor.ts
@@ -1306,7 +1306,11 @@ export class Editor<LAnnotation> implements DiffsEditor<LAnnotation> {
           mvShortcut !== undefined &&
           textDocument !== undefined
         ) {
-          const cursorMoveOptions = this.#getCursorMoveOptions();
+          const cursorMoveOptions: CursorMoveOptions | undefined = this.#isWrap
+            ? {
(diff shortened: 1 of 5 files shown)


packages/diffs/src/editor/editor.ts
```diff
@@ -1302,13 +1306,28 @@ export class Editor<LAnnotation> implements DiffsEditor<LAnnotation> {
           mvShortcut !== undefined &&
           textDocument !== undefined
         ) {
+          const cursorMoveOptions: CursorMoveOptions | undefined = this.#isWrap
+            ? {
```
ije, line 1310: added a small change: write `#getCursorMoveOptions` logic inline

ije (approved):

PR #956, opened 2026-07-09
fix(diffs): Correct line count when edits split or form a CR/LF

necolas:
Reproduce:
1. Open a document whose content is "a\r\nb" (two lines).
2. Insert "X" between the \r and the \n (offset 2) to make "a\rX\nb".
3. Ask the document for its line count.

It reports 2 lines instead of 3. In that state positionAt(2) returns {line:0,character:2} instead of {line:1,character:0}, getLineText(2) throws "Line index out of range", and a search for "X" returns the wrong range. The mirror case is also wrong: inserting a "\r" just before an existing "\n" reports an extra line instead of merging the two into one \r\n break.

Each piece derived its line-break count from the shared buffer's offsets, which record a \r\n as one break after the \n. An edit that split a \r\n across a piece boundary, or formed one across two pieces, was then miscounted, corrupting the line count and every position, line-text, and search result that relies on it. The add buffer had the same flaw: appending text that begins with "\n" after a buffer ending in "\r" recorded two breaks for the resulting \r\n.

Count each piece's breaks as if its text stood alone, then reconcile boundary pairs in the tree by folding a \r ending one piece with a \n starting the next into a single break. Drop the stale lone-\r offset in the add buffer when an append merges it into a \r\n. Trim the trailing line break in getTextSlice once at the slice end instead of per piece, so an interior \r split across pieces is preserved.

### DIFF
--- a/packages/diffs/src/editor/pieceTable.ts
+++ b/packages/diffs/src/editor/pieceTable.ts
@@ -16,12 +16,19 @@ class Piece {
     public readonly offset: number,
     public readonly length: number,
     public readonly lineOffsetStart: number,
-    public readonly lineOffsetEnd: number
+    public readonly lineOffsetEnd: number,
+    // Line breaks counted as if this piece's text stood on its own: `\r`, `\n`,
+    // and `\r\n` each count once, and a `\r` at the very end counts as a lone
+    // break even when the shared buffer paired it with a following `\n` that
+    // now lives in a different piece. Pairs that straddle a piece boundary are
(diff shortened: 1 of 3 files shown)


ije:
how can i insert "X" between the \r and the \n (offset 2) to make "a\rX\nb" in the edit file?  the fix is great! but seems introduce some extra complex, as we do not expose the piecetable class as API.

fat (approved):

PR #927, opened 2026-07-01
Add playwright tests for diffs and edit mode

mdo:
## Add browser E2E tests for diffs & edit mode

Adds a Playwright E2E suite for `@pierre/diffs`, covering behavior the jsdom
unit tests can't reach: shadow-DOM rendering and real `contentEditable` editing.
Mirrors the existing `trees` E2E setup — a Vite server serves the built `dist`
and Playwright drives real Chromium against static HTML fixtures.

### Coverage
- **Diff rendering**: split/unified toggle, hunk expansion, accept/reject
- **Edit mode**: typing, `onChange`, undo/redo, additions-editable vs
  deletions-read-only, full-file rewrite
- **Find & replace**, **multi-cursor + indent/outdent**, **markers** (hover +
  bounds), **line selection + gutter utility**, **annotations**,
  **theme switching** (selection-stable), **selection actions** (in-bounds)
- **Merge conflicts** and **added/deleted file states**

### Notable fix
- Editor now falls back to `ShadowRoot.getSelection()` when
  `Selection.getComposedRanges` is unavailable (older browsers, WebViews, and
  the pinned CI Chromium), so a click reliably seeds the caret and typing works.

### Infra
- `diffs:test-e2e` + `diffs:test-e2e-server` moon tasks; `diffs:test-e2e` wired
  into CI
- Reserves worktree port `4175` (`scripts/wt.ts`, `scripts/README.md`)
- Shared `test/e2e/e2e-globals.d.ts` for fixture `window` globals

## Running the tests

Run the full suite: `moon run diffs:test-e2e`

View the fixtures in your own browser — start the server and open the index:

- `moon run diffs:test-e2e-server`
- Open http://127.0.0.1:4175/test/e2e/fixtures/

The index page lists every fixture with a description, its ready-flag, and the
spec that drives it.

PR #957, opened 2026-07-09
Updated Playwright tests, modify edit mode popover flipping

mdo:
Follow up to #927.

- Refines selection action test
- Adds another selection action test to confirm flipping behaviors
- Modifies popover flipping detection for instances where short files cause placement issues


mdo:
@ije Could use your eyes on the source changes here, please.

ije (approved):
Lgtm

necolas (approved):

PR #959, opened 2026-07-09
fix(diffs): Clear web font text width cache

necolas:
1. Render the diffs editor with a custom monospace web font that uses font-display: swap.
2. Before the font file finishes loading, move the caret or make a selection on a line containing emoji.
3. Let the web font finish loading without resizing the editor.
4. Notice the caret and selection x positions still use fallback-font DOM widths until an unrelated resize.

Metrics.remeasureCharacterWidth only reported a refresh when ch changed, so font-load completion with the same ASCII advance skipped both cache invalidation and repaint. Clear cached DOM widths when present and return true for that invalidation too, so DOM-measured runs are remeasured with the loaded font.

### DIFF
--- a/packages/diffs/src/editor/textMeasure.ts
+++ b/packages/diffs/src/editor/textMeasure.ts
@@ -74,15 +74,18 @@ export class Metrics {
 
   /**
    * Re-measure the '0' character width against the font that is loaded right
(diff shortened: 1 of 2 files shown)


ije (approved):
lgtm

PR #960, opened 2026-07-09
fix(diffs): Stop structural edits scanning to EOF

necolas:
1. Open a large TypeScript file in the diffs editor.
2. Let tokenization populate the file, then press Enter near the top.
3. Keep typing near that edit and observe each background pass restart and scan the tail instead of stopping after grammar state reconverges.

Line-count edits truncated cached grammar end states and scheduled the background job without changed ranges, leaving it no prior state to compare with. Shift cached end states across inserted or deleted lines and carry changed ranges into background tokenization so it settles once state matches the shifted tail again.


necolas:
@codex

ije (approved):
nice

PR #961, opened 2026-07-09
fix(diffs): Keep column-zero carets from matching prior brackets

necolas:
1. Open a diff file in the editor with a `{` at the end of one line and its matching `}` later in the file.
2. Place the caret at column 0 on the line immediately after the `{`.
3. The editor highlights the previous line's `{` and its matching `}`, even though the caret is not beside either bracket.

The matcher treated the previous line's last character as adjacent to column-zero carets. Keep the previous-character lookup line-local so column-zero carets can only match a bracket at the current position.


ije (approved):
good fix!

PR #963, opened 2026-07-09
fix(diffs): honor Shift+Enter in search input

necolas:
1. Open the diffs editor with at least three repeated matches.
2. Press Cmd/Ctrl+F, search for the repeated text, and press Enter once to move from the first match to the second.
3. Press Shift+Enter while focus remains in the find input.
4. The selection should move back to the first match; before this fix, it advanced to the third match instead.

The find input treated every Enter keydown as forward navigation. Pass the event's shiftKey into the existing search navigation helper so Shift+Enter uses the previous-match path, matching Cmd/Ctrl+Shift+G and the previous button.

Add a jsdom regression test for the panel that asserts Enter moves forward and Shift+Enter returns to the previous range.

### DIFF
--- a/packages/diffs/src/editor/searchPanel.ts
+++ b/packages/diffs/src/editor/searchPanel.ts
@@ -302,7 +302,7 @@ export class SearchPanelWidget {
           close();
         } else if (e.key === 'Enter') {
           e.preventDefault();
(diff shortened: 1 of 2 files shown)


ije (approved):
LGTM

PR #965, opened 2026-07-10
fix(diffs): Treat trailing tabs like spaces for word deletion

necolas:
1. Open a diffs editor document containing `hello<Tab>world`.
2. Put the caret immediately after the tab, before `world`.
3. Run deleteWordBackward.
4. Observe that the editor deletes only the tab and leaves `helloworld` instead of deleting `hello<Tab>` and leaving `world`.

The word-deletion scan skipped assigning the first trailing whitespace cluster only when that cluster was a literal space. Reuse the computed whitespace class so a single trailing tab follows the same path, while multi-tab whitespace runs still delete as their own group.

### DIFF
--- a/packages/diffs/src/editor/selection.ts
+++ b/packages/diffs/src/editor/selection.ts
@@ -1997,7 +1997,7 @@ function resolveDeleteWordBackwardRange(
     if (match !== undefined && nextMatch !== match) {
       break;
     }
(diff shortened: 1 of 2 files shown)


fat (approved):

ije (approved):
ltgm!

PR #966, opened 2026-07-10
fix(diffs): Reselect auto-surrounded quote text

necolas:
1. In the diffs editor, start with a document whose selected text is a single double quote.
2. Type another double quote so auto-surround replaces the selection with three double quotes.
3. The editor should reselect the original middle quote, but it reselects the newly inserted opening quote.

Detect exact auto-surround replacements before falling back to the generic substring search. This keeps the preserved selection at offset one when the selected text is also the opening delimiter, while leaving non-wrapper replace behavior on the existing indexOf fallback.


fat (approved):

ije (approved):
lgtm

PR #967, opened 2026-07-10
fix(diffs): Keep NFD graphemes intact in editor measurement

necolas:
To reproduce:
1. Open a diffs editor with line wrapping enabled in a narrow pane.
2. Add a long line containing decomposed text, such as repeated `Café` (or `Cafe\u0301`) words where the accent is U+0301.
3. Resize the pane until a soft wrap falls inside one `e\u0301` grapheme (e.g., `é` is the first symbol on the wrapped line), then move the caret vertically across that wrap.

The caret can land between `e` and the combining accent, and the wrapped line can split that grapheme across visual rows.

Detect Unicode combining marks in needsDomTextMeasurement so the existing grapheme segmenter drives caret snapping and wrap offset measurement for NFD text. Add predicate, snapping, measurement-offset, and wrapped navigation regressions.

### DIFF
--- a/packages/diffs/src/editor/textMeasure.ts
+++ b/packages/diffs/src/editor/textMeasure.ts
@@ -1,10 +1,12 @@
 import { getGraphemeSegmenter, h, round } from './utils';
 
 // Upper bound on cached DOM text-width measurements. The cache only holds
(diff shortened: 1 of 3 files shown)


fat (approved):

ije (approved):
nice!

PR #973, opened 2026-07-12
fix(diffs): Preserve controlled editor selection

jakeleventhal:
## What changed

- Reuse the attached file instance when a controlled React parent echoes the
  editor's current name, language, and contents with a new cache key.
- Add a React regression test covering both the controlled-value echo and a
  genuine external content replacement.

## Why

`useFileInstance` previously passed every controlled `file` update to
`instance.render`. When an `Editor.onChange` consumer stored the returned file
and generated a new cache key, the adapter treated that echo as an external
document replacement. The editor rebuilt its document and dropped selection
and focus after each change.

The adapter now preserves editor-owned state for same-document echoes while
continuing to render real external changes normally.

## Validation

- `moon run root:format root:lint`
- `moonx diffs:typecheck`
- `moonx diffs:test` (1055 tests)

### DIFF
--- a/packages/diffs/src/react/utils/useFileInstance.ts
+++ b/packages/diffs/src/react/utils/useFileInstance.ts
@@ -132,7 +132,16 @@ export function useFileInstance<LAnnotation>({
       newOptions
     );
     instanceRef.current.setOptions(newOptions);
-    void instanceRef.current.render({ file, lineAnnotations, forceRender });
+    const editorFile = contentEditable ? editor?.getFile() : undefined;
+    const renderFile =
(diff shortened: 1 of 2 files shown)


jakeleventhal:
@SlexAxton @amadeus before i start spam addressing AI comments, can you please let me know what you think of direciton/intent?

ije:
you can use the `editor.getState()` and `editor.setState` (or `editor.setSelections`) APIs to preserve selection. i'm currently working on a document cache API that holds all opened textDocument. (this should work on the vanillaJS components as well)

ije:
see https://github.com/pierrecomputer/pierre/pull/976

PR #985, opened 2026-07-14
test(diffs): align getPlacementBounds tests with code-box clip

mdo:
#957 made `getPlacementBounds` clip popovers to `[data-code]`'s own box, but two Bun unit tests still asserted the old bounds and failed.

- Update the scroll-container and window-fallback tests to stub a code box taller than the scrollport, isolating the viewport→code-relative mapping they check.
- Add a test for the new behavior: a short file clamps the bounds to the code box.


fat (approved):
gracias

PR #986, opened 2026-07-14
[diffs/edit] Sanitize malformed position components in normalizePosition

fat:
## What

Fixes the worst bug found by the editor-behavior test audit: an edit whose range carries a `NaN` (or fractional) `line`/`character` component silently **replaced the entire document** with the inserted text.

Repro before this fix:

```ts
const d = new TextDocument('inmemory://x', 'hello\nworld', 'plain');
d.applyEdits([{ range: { start: { line: 0, character: NaN }, end: { line: 0, character: NaN } }, newText: 'X' }]);
d.getText(); // "X"  — everything else gone
```

## Why

`normalizePosition` clamped with bare `Math.max(0, Math.min(...))`, which passes `NaN` straight through, and fractional indexes survived the clamps only to break the integer-keyed line-offset lookup. Either way the resolved offset became `NaN`, and a `NaN`-ranged edit degenerates into a whole-document replace.

## Fix

`normalizePosition` sanitizes components before clamping, consistent with its existing silent-clamp policy for out-of-range values: `NaN` and `-Infinity` act as 0, fractions floor, `+Infinity` clamps to the max. Every position-taking API funnels through it (`offsetAt`, `#resolveEdit`), so all edit paths are covered in one place. `+Infinity` behavior is byte-identical to before (already clamped correctly by accident).

## Tests

- The two `test.failing` pins for this bug flip to permanent regression tests with exact expected outputs (insertion lands at the sanitized position, no content lost).
- The `Infinity` divergence pin is unchanged and still passes.
- Full diffs suite: 1278 pass / 0 fail; typecheck and oxlint clean.
- `test/README.md` known-bug inventory: 32 → 30.

First PR of the failing-test burn-down; next up is the history-equivalence cluster (7 pins).

### DIFF
--- a/packages/diffs/src/editor/textDocument.ts
+++ b/packages/diffs/src/editor/textDocument.ts
@@ -316,14 +316,26 @@ export class TextDocument<LAnnotation> {
   }
 
   normalizePosition(position: Position): Position {
-    const line = Math.max(0, Math.min(position.line, this.lineCount - 1));
+    const line = TextDocument.#clampIndex(position.line, this.lineCount - 1);
     return {
       line,
-      character: Math.max(
-        0,
-        Math.min(position.character, this.getLineLength(line))
+      character: TextDocument.#clampIndex(
+        position.character,
+        this.getLineLength(line)
       ),
     };
+  }
+
+  // Math.min/max pass NaN through, and a fractional index breaks the
+  // integer-keyed line-offset lookup — either way the resolved offset becomes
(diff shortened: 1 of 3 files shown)


necolas (approved):

amadeus (approved):
lgtm. could there be a valid reason to throw in this situation? or is there something about the API where a user could provide bad data? But probably not a big deal overall

fat:
mm don't think so

PR #987, opened 2026-07-14
[diffs/edit] Programmatic edits join the undo timeline like local edits

fat:
## What

Fixes the largest bug cluster from the editor-behavior audit (7 pinned bugs, one root cause): history entries were frozen while `applyEdits(..., updateHistory=false)` bypassed the EditStack, so any programmatic edit left the stack holding stale offsets — undo corrupted text, redo replayed at wrong positions, and undo-to-exhaustion could never restore the original document.

## Design decision

Programmatic edits are **not a separate semantic class**: a mixed programmatic/local sequence must behave exactly like the same sequence applied all-locally. Concretely, after any sequence, undo-to-exhaustion restores the original text and redo-to-exhaustion the final text, with per-step states matching the all-local run.

```ts
// before: undo corrupts ('ilot?' …); after:
doc.applyEdits(insert('!'), true);   // local  -> "base!"
doc.applyEdits(insert('>> '));       // programmatic -> ">> base!"
undoAll(doc);                        // ">> base!" -> "base!" -> "base"
redoAll(doc);                        // -> ">> base!?" — byte-exact both ways
```

## How

- `TextDocument.applyResolvedEdits` always creates/coalesces/pushes an EditStack entry; `updateHistory` now only gates caller-supplied selection + undo-boundary **metadata** on the entry. Redo clears on every recorded edit, tracked or not.
- `Editor.applyEdits` always tracks; its `updateHistory` param is kept for API compatibility but deprecated.
- **No recursion / no bypass leak**: `undo()`/`redo()` always wrote through the private `#applyResolvedEditsToBuffer` and are untouched; that private method is the only non-tracking door. History reset remains document rebuild with a fresh stack.
- **Composition unaffected**: previews never touch the document (browser composes in DOM; `compositionend` commits once through the tracked path) — verified end-to-end in review.

## Tests

- The 7 equivalence pins flip to permanent regression tests with **assertions unchanged** — they encode all-tracked reference runs, independently re-derived by a snapshot-based simulator during review (25/25 values matched).
- The 2 old-model companion pins are rewritten to the new contract; exactly **one** other test in the suite pinned the bypass (`applyEdits default does not record undo`) and is flipped with a cross-reference.
- Docs updated (`applyEdits` example + History section) — they described the old flag.
- Full diffs suite: 1278 pass / 0 fail (run twice, identical); typecheck + oxlint clean. README known-bug inventory: 32 → 25.

## Review notes

The riskiest surfaces were checked specifically: recursive recording via undo/redo replay, bypass leaking publicly, composition previews entering history, redo-clear semantics on the coalesce path, metadata-less entries in selection restore, and version-counter drift — all clean. Second PR of the failing-test burn-down (after #986).


ije:
@codex review

ije (approved):
LGTM

PR #990, opened 2026-07-15
[diffs/edit] Fix PieceTable CRLF metadata and resulting search range drift

ije:
> **PieceTable CRLF line metadata** (5, spanning piece-table and search
  coverage) — line-break bookkeeping goes stale when a CRLF pair is split or
  assembled across edits (deleting exactly the `\n` of a pair, inserting between
  the `\r` and `\n`, assembling CRLF from two separate inserts, plus a
  CRLF-biased fuzz oracle), and the same stale metadata drives search astray
  (shifted or missing match ranges) even though `getText()` stays correct.

Fixed PieceTable CRLF metadata and resulting search range drift.

- Piece/tree metadata now reconciles CRLF pairs across piece seams while preserving lone \r breaks.
- Prevented unsafe coalescing across separately appended CRLF boundaries.


ije:
@codex review

necolas (approved):

PR #991, opened 2026-07-15
[diffs/edit] Fix batch edit ordering sensitivity

ije:
> **Batch edit ordering sensitivity** (1) — accepting a batch containing a
  delete and an insert at the same offset depends on the caller's array order:
  delete-first throws an overlap error, insert-first succeeds.

Fixed batch edit ordering sensitivity.
Equal-start edits now sort by end offset, placing inserts before deletes regardless of caller order.


ije:
@codex review

necolas (approved):

PR #993, opened 2026-07-15
[diffs/edit] Fix the CRLF position round-trip bug

ije:
> **Position round-trip losing a column** (1) — a position-from-offset
  computation can return a position strictly inside a CRLF pair (a character
  beyond the line's own length) that the inverse offset-from-position
  computation clamps away, so the round trip silently loses a column.

Normalized both positionAt and positionsAt outputs.


ije:
@codex review

necolas (approved):

PR #994, opened 2026-07-15
[diffs/edit] Fix search-replace capture expansion with lookaround

ije:
> **Search-replace capture expansion with lookaround** (3) — replacement-text
  expansion re-executes the pattern against only the matched slice, so
  lookaround context outside the slice is lost: a lookbehind whose context sits
  before the slice, a lookahead whose context sits after the slice, and a
  lookahead that re-matches shorter on the slice all fall back to inserting the
  literal (unexpanded) replacement text.

Fixed search-replace capture expansion with lookaround.
- Re-executes regex against the full line at the original match offset.
- Validates the exact match range before expanding captures.


ije:
@codex review

necolas (approved):

PR #995, opened 2026-07-15
[diffs/edit] Normalize inverted editor selections

ije:
> **Inverted selection after normalization** (1) — setting selections normalizes
  positions but never reorders a start-after-end pair, storing an inverted
  selection that violates the start-before-or-equal-end invariant downstream
  code assumes.

Normalized endpoints are reordered and direction is flipped to preserve focus.


ije:
@codex review

necolas (approved):

PR #999, opened 2026-07-16
[diffs] add folding support

ije:
Screenshot 2026-07-17 at 01 21 41

- Add `folding` option for `File` and `FileDiff` options. (Default is `false`)
- Works in edit mode
- Support both split and unified diffs layouts
- Support A11y

## Usage

```ts
new File({
  folding: true,
});

new FileDiff({
  folding: true,
});
```


ije:
@codex review

tomerzcod7:
Hey guys, thanks for working on this - we use your library extensively and this is one feature we've been needing.
Please let me know if this is not the place for a comment like this.

We have a common use case, where we use the React File component as a read-only viewer, particularly for large, pretty-printed JSON files. We’d love to use the new nested folding controls there without enabling edit mode.
Any chances you would be open to support folding for read-only File component as well? maybe via the FileOptions?
Alternatively, is there an intended way to attach the experimental Editor for folding while preventing content edits?

We would love to adopt an experimental version and would be happy to test it and provide feedback.

ije:
> Any chances you would be open to support folding for read-only File component as well?

this makes sense to me, will support!

PR #1000, opened 2026-07-16
[diffs/edit] Fix edit stack coalescing

ije:
> **EditStack coalescing** (3) — coalescing decisions compare a new edit against
  whatever sits on top of the undo stack purely by geometry, with no state reset
  after undo/redo: an undo can expose a stale top entry that new typing then
  fuses into; an undo-boundary marker stops blocking merges once it is undone;
  and backspace followed by forward-delete at the same pivot coalesces into a
  single undo step instead of getting an undo stop when the delete direction
  flips.

- Undo/redo now closes the active coalescing group, preventing stale or boundary-hidden entries from merging.
- Delete entries retain backspace/forward-delete direction, so direction flips create undo stops.


ije:
@codex review

PR #979, opened 2026-07-14
[diffs/edit] No more required props (except `useTokenTransform`

amadeus:
Made it so we can preserve all line selection, annotation and line highlights while in edit mode. 

Of course a dev could still choose to change specific props on a per file basis if necessary.

- [x] - Test
- [x] - Update docs to reflect new world

PR #1005, opened 2026-07-20
[diffs/edit] Use editor gutter interaction handler for line selection

ije:
https://github.com/pierrecomputer/pierre/pull/979 fixed `enableLineSelection` and other file/diffs options that did have good compatibility with the editor. however the editor has the own gutter interaction handler.

this pr uses editor gutter interaction when a editor attached without changing the file/difss options, and the previous line selection remain.

### DIFF
--- a/packages/diffs/src/components/File.ts
+++ b/packages/diffs/src/components/File.ts
@@ -538,6 +538,7 @@ export class File<
   public attachEditor(editor: DiffsEditor<LAnnotation>): () => void {
     this.editor?.cleanUp();
     this.editor = editor;
+    this.interactionManager.setEditorAttached(true);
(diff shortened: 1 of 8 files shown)


amadeus:
So I had made a bit of an intentional design decision here. If enableLineSelection was true, then you’d lose the editor line selection functionality. You also don’t need to use that prop to make line selections because it can be done programmatically or via gutter utility.

I think this makes for a cleaner separation of concerns, and if devs want the edit line selection they could toggle the prop when editing

ije:
i see! i create another pr for fixing the selection rounded corner rendering with selected lines in https://github.com/pierrecomputer/pierre/pull/1007

PR #1006, opened 2026-07-20
[editor] Improve EditContext API

amadeus:
Overall this branch converts the 1:1 editor context API into a function factory context to make edit more developer friendly to implement.

### TL;DR

So it was kinda buggin me that to use edit in react involved this architecture:

```tsx
<EditProvider>
  <FileDiff edit />
</EditProvider>
<EditProvider>
  <File edit />
</EditProvider>
```

Basically this created a scenario where you had to create an edit context for every File/FileDiff. It also created situations where if you want to toggle edit on or off, and didn't want to pay the cost for an editor instance when it wasn't used, you'd essentially have to remount FileDiff (changing parent/child relationships in react is always an automatic re-mount).

Of course, part of the reason this is necessary is because we don't want to force people to import/use the Editor codebase if they aren't going to use it. 

When I worked through adding support for CodeView, I took a slightly different approach -- you create an `editFactory` function that then CodeView would automatically use to create Editor instances on demand for items.

I feel like this is a much cleaner solution for the React API so I did a bit of an agentic port to rework EditContext to supply a factory function, and utilize that to automatically generate instances on demand.

I also changed the `contentEditable` prop to `edit` which feels more on brand for how we are making the feature.

So now, when building out a page that uses diffs, that might also selectively edit, you can just do
```tsx
function createEditor(options) {
  return new Editor(options);
}

// Ideally you place this context high up in your app's layout,
// so anything that needs it can just get it
<EditContext createEditor={createEditor}>
  <File edit />
  <FileDiff edit />
  {/* ...etc */}
</EditContext>
```

### The Longer Version:

- **Breaking:** Replace `<EditProvider editor={editor}>` with `<EditProvider createEditor={factory}>`, export `CreateEditor`, `EditProviderProps`, and `useCreateEditor`, and remove `useEditor`.
- **Breaking:** Rename `contentEditable` to `edit` on `File`, `FileDiff`, `MultiFileDiff`, and `PatchDiff`, and add per-surface `editOptions`; `UnresolvedFile` remains non-editable.
- Create an independent editor for every surface and edit session without remounting the underlying DOM or imperative instance; exiting cleans the editor, re-entry starts fresh history, and factory or option replacements apply to the next session.
- Move React `CodeView` editor creation into the nearest `EditProvider`, add CodeView-level `editOptions`, preserve per-item editors across virtualization, and keep vanilla `CodeViewOptions.createEditor` unchanged.
- Route React `CodeView` changes through item-aware `onItemEditChange` and `onItemEditComplete` callbacks, with completion limited to changed sessions ending through edit-off, collapse, or removal.
- Validate missing or invalid editor factories lazily when an editable surface or rendered CodeView item needs attachment, and clean newly created editors when attachment fails.
- Remove the ignored `controlledSelection` and `onSelectedLinesChange` keys from React CodeView's `options` object; controlled prop-based and uncontrolled selection remain supported.
- Prevent deferred `Editor.onAttach` callbacks from outliving their edit sessions, preserve correct delivery across virtualization, and make shared render scheduling idempotent.
- Mount one shared editor factory provider at the docs app root, migrate all diffs and trees examples to `edit` and memoized `editOptions`, fix Live Editing state across layout changes, and refresh the Agent UI starter/generated artifacts.
- Expand React lifecycle coverage for every editable surface, simultaneous siblings, factory and option replacement, StrictMode, virtualization, attachment failures, cleanup, selection, and CodeView edit completion.


amadeus:
@codex review

packages/diffs/src/react/utils/useFileDiffInstance.ts
necolas: useIsomorphicLayoutEffect

necolas (approved):

PR #1008, opened 2026-07-20
[diffs/edit] Add search-match viewport culling

ije:
This pr fixes search-match overlay renders all matches with no viewport culling by:

- Binary-searches offset-sorted matches for each rendered line segment.
- Skips virtualized offscreen matches and matches hidden inside collapsed diff gaps.


ije:
@codex review

amadeus (approved):

PR #1010, opened 2026-07-20
[diffs/edit] Clamp selection rendering to render window

ije:
Selection bounds are intersected with the half-open render window once before iteration, preserving original endpoint geometry.
Added a **10,000-line** regression test with a seven-line middle window.


ije:
@codex review

amadeus (approved):

PR #1012, opened 2026-07-21
diffs: add starlark file extensions

lf-:
If shiki ever grows Starlark support, this should be revised to use that starlark support, but Starlark is a strict subset of Python anyway.

ref for the file extensions:
https://bazel.build/concepts/build-ref
https://buck2.build/docs/rule_authors/package_files/
https://buck2.build/docs/bxl/

### Description

Add support for highlighting Starlark files like those from Bazel/Buck2/Copybara.

### Motivation & Context

I like my diffs for my build files to be readable.

### Type of changes

      being opened
      that this PR is being opened
- [ ] Documentation update

### Checklist

- [x] I have read the
- [x] I have updated the documentation accordingly (if applicable)

wrote this with claude then realized it was all out of order and redid it by hand lmao

### DIFF
--- a/packages/diffs/src/utils/getFiletypeFromFileName.ts
+++ b/packages/diffs/src/utils/getFiletypeFromFileName.ts
@@ -243,6 +243,19 @@ export const EXTENSION_TO_FILE_FORMAT: ExtensionFormatMap = {
   py: 'python',
   pyw: 'python',
   pyi: 'python',
+  // FIXME: these should be changed to a Starlark file type if/when one
+  // appears in shiki. For now, Starlark is highlighted like the Python subset
+  // it is.
+  BUCK: 'python',
+  BUILD: 'python',
+  MODULE: 'python',
+  PACKAGE: 'python',
+  WORKSPACE: 'python',
+  bazel: 'python',
+  bxl: 'python',
+  bzl: 'python',
+  star: 'python',
+  sky: 'python',
   qml: 'qml',
   qmldir: 'qmldir',
   qss: 'qss',


amadeus:
@codex review

packages/diffs/src/utils/getFiletypeFromFileName.ts
```diff
@@ -243,6 +243,19 @@ export const EXTENSION_TO_FILE_FORMAT: ExtensionFormatMap = {
   py: 'python',
   pyw: 'python',
   pyi: 'python',
+  // FIXME: these should be changed to a Starlark file type if/when one
+  // appears in shiki. For now, Starlark is highlighted like the Python subset
+  // it is.
+  BUCK: 'python',
+  BUILD: 'python',
```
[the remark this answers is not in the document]
amadeus: Yeah this issue is due to my laziness around our file name detection (i.e. i need to probably extend this to properly test the filename), but probably shouldn't be apart of this PR.

packages/diffs/src/utils/getFiletypeFromFileName.ts
```diff
@@ -243,6 +243,19 @@ export const EXTENSION_TO_FILE_FORMAT: ExtensionFormatMap = {
...
+  MODULE: 'python',
+  PACKAGE: 'python',
+  WORKSPACE: 'python',
```
amadeus, line 253: I'm a bit worried about including 

BUILD, MODULE, and PACKAGE, since those feel pretty generic. The other ones definitely seem bazel related. What do you think about removing those three since there's a custom API you can use to extend for the more ambiguous ones?

Another possibility might be to say, what if we just remove all of these direct filenames, until i can improve that part of the system to properly track BUCK files in a sub directory?

lf-: Some thoughts:

Bazel is super popular, so BUILD is most likely always going to be Bazel, and false positives are probably worth the benefit for these vs the possibility to have a false positive. Approximately one MODULE file appears in a Bazel repo; so it's lower value compared to BUILD.

PACKAGE is a buck specific thing and is, of the options, the least credible just given that Buck is a lot less popular than Bazel (c.f. the time [i tried to add it to linguist lmao](https://github.com/github-linguist/linguist/pull/7562)). This is the one I would put into consumers rather than the defaults; I guess the other thought I have is that the consumers could override the decision if they *don't* like the defaults for BUILD or similar.

It's definitely worth distinguishing direct filenames from extensions in the recognizer.

packages/diffs/src/utils/getFiletypeFromFileName.ts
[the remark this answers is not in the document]
HackAttack: #1153

packages/diffs/src/utils/getFiletypeFromFileName.ts
```diff
@@ -243,6 +243,19 @@ export const EXTENSION_TO_FILE_FORMAT: ExtensionFormatMap = {
...
+  bazel: 'python',
+  bxl: 'python',
+  bzl: 'python',
```
HackAttack: WORKSPACE.bzlmod is accepted as a transitional file during Bzlmod migration
```suggestion
  bzl: 'python',
  bzlmod: 'python',
```

packages/diffs/src/utils/getFiletypeFromFileName.ts
HackAttack:
```suggestion
  WORKSPACE: 'python',
  Tiltfile: 'python',
```

packages/diffs/src/utils/getFiletypeFromFileName.ts
```diff
@@ -243,6 +243,19 @@ export const EXTENSION_TO_FILE_FORMAT: ExtensionFormatMap = {
   py: 'python',
   pyw: 'python',
   pyi: 'python',
+  // FIXME: these should be changed to a Starlark file type if/when one
+  // appears in shiki. For now, Starlark is highlighted like the Python subset
+  // it is.
+  BUCK: 'python',
+  BUILD: 'python',
+  MODULE: 'python',
```
HackAttack: I think only MODULE.bazel is valid actually
```suggestion
```

HackAttack (commented):
I had done a similar change before finding this PR, in support of https://github.com/modem-dev/hunk/pull/813. Needs tests, feel free to take what I had generated:
```typescript
test('highlights Bazel and Starlark files with the Python grammar', () => {
  expect(getFiletypeFromFileName('defs.bzl')).toBe('python');
  expect(getFiletypeFromFileName('tools/defs.bzl')).toBe('python');
  expect(getFiletypeFromFileName('rules.star')).toBe('python');
  expect(getFiletypeFromFileName('copy.bara.sky')).toBe('python');
  expect(getFiletypeFromFileName('BUILD.bazel')).toBe('python');
  expect(getFiletypeFromFileName('pkg/nested/MODULE.bazel')).toBe('python');
  expect(getFiletypeFromFileName('WORKSPACE.bzlmod')).toBe('python');
  expect(getFiletypeFromFileName('Tiltfile')).toBe('python');
});

test('resolves Bazel package files at any depth', () => {
  expect(getFiletypeFromFileName('BUILD')).toBe('python');
  expect(getFiletypeFromFileName('pkg/BUILD')).toBe('python');
  expect(getFiletypeFromFileName('a/b/c/WORKSPACE')).toBe('python');
  expect(getFiletypeFromFileName('third_party/BUCK')).toBe('python');
});
```
They won’t all pass without #1153 though.

PR #1014, opened 2026-07-21
[diffs/edit]: keep virtual EOF caret anchored after line reinsertion

ije:
Reproduction steps:
0, Open the demo page, enable editing, and jump to the last line.
1, Press Enter to insert a new line.
2, Press Delete to remove the newly added final line.
3, **[Bug]** Press Enter again. The page scrolls to the top, and the caret does not remain focused on the newly inserted line.

this PR fix that issue by:
- Reset cached line geometry after every line-count change.
- Prevent a removed EOF row from being reused to position the caret at the top.
- Add regression coverage for Enter → Delete → Enter in a 1,000-line virtualized file.


ije:
@codex review

amadeus (approved):

PR #1017, opened 2026-07-22
[edit] Edit Focus Polish Improvements

amadeus:
Mostly the was to improve some API paper cuts around line focusing with the editor

- Extend `Editor.focus()` with explicit line/character targeting and a `'first-visible'` mode that preserves scroll position.
- Resolve visible editor rows across standard diffs, element/window Virtualizers, and CodeView, including sticky headers and scroll-container fallbacks.
- Keep `onAttach` reliable across queued rerenders and virtualized recycling without targeting detached DOM or stealing focus from other controls.
- Add opt-in autofocus examples to every editable playground mode and document the React and vanilla APIs.
- Add some documentation around editor focusing and related APIs
- Expand regression coverage for viewport resolution, editable-row filtering, attachment lifecycle, and CodeView option forwarding.

### DIFF
--- a/packages/diffs/src/components/FileDiff.ts
+++ b/packages/diffs/src/components/FileDiff.ts
@@ -1862,6 +1862,9 @@ export class FileDiff<
       previousContainer ??
       document.createElement(DIFFS_TAG_NAME);
     const containerChanged = previousContainer !== nextContainer;
+    if (previousContainer != null && containerChanged) {
+      this.editor?.__captureFocusForDOMReplacement();
+    }
     if (containerChanged) {
       this.emitPostRender(true);
     }
@@ -1943,6 +1946,7 @@ export class FileDiff<
     // If we have a new parent container for the pre element, lets go ahead and
     // move it into the new container
     else if (this.pre.parentNode !== shadowRoot) {
+      this.editor?.__captureFocusForDOMReplacement();
       shadowRoot.appendChild(this.pre);
       this.appliedPreAttributes = undefined;
     }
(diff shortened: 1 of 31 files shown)


amadeus:
@codex review

packages/diffs/src/types.ts
```diff
@@ -1046,6 +1046,11 @@ export interface DiffsEditableComponent<
    * Return the scroll container element.
    */
   getScrollContainer?: () => HTMLElement | undefined;
+  /**
+   * Return an explicit viewport that bounds visible editor rows. Components
+   * without one fall back to their nearest scrollable ancestor or document.
+   */
+  getEditorViewport?: () => HTMLElement | Document | undefined;
```
ije, line 1053: why not just use `getScrollContainer`?, they are basically same

ije: otherwise LGTM

packages/diffs/src/editor/editor.ts
```diff
@@ -1074,6 +1104,7 @@ export class Editor<LAnnotation> implements DiffsEditor<LAnnotation> {
     } else if (
       this.#selections !== undefined &&
       this.#selections.length > 0 &&
+      this.#contentHasFocus &&
       !this.#retainSearchPanelFocus
```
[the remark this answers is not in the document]
ije: seems this is real

packages/diffs/src/types.ts
amadeus: Ahh this is interesting, i see that there's a bug with getScrollContainer, because it doesn't account for CodeView... i'll do a bit of digging, good callout

amadeus:
@codex review

ije (approved):
LGTM

packages/diffs/src/components/FileDiff.ts
```diff
@@ -2212,6 +2236,7 @@ export class FileDiff<
     const unifiedAST = this.hunksRenderer.renderCodeAST('unified', result);
     const deletionsAST = this.hunksRenderer.renderCodeAST('deletions', result);
     const additionsAST = this.hunksRenderer.renderCodeAST('additions', result);
+    this.editor?.__captureFocusForDOMReplacement();
```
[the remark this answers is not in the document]
amadeus: This is not relevant. the beta release does not care about backwards compatibility.

PR #1021, opened 2026-07-23
fix(diffs): Wrap annotation content in every overflow mode

necolas:
Annotation prose did not wrap in the default scroll overflow mode. React consumers had to add `white-space: normal` to each annotation by hand, or the text overflowed the diff column.

The diffs stylesheet lives in the shadow root, so slotted annotation content gets its white-space by inheritance from the shadow-DOM `[data-annotation-content]` parent. That parent normalized white-space only under `[data-overflow="wrap"]`. In scroll mode it inherited `white-space: pre` from the `pre` element, so the slotted prose did not wrap.

Set `white-space: normal` on `[data-annotation-content]` in every overflow mode. Annotations hold prose, not code, so they never need `pre`. One rule fixes every render path (React, imperative, SSR) and keeps them consistent.

Also remove the now-unnecessary `white-space: normal` workaround from the docs annotations demo, and add a browser regression test for the computed white-space in scroll mode.

### DIFF
--- a/apps/docs/app/(diffs)/_examples/Annotations/Annotations.tsx
+++ b/apps/docs/app/(diffs)/_examples/Annotations/Annotations.tsx
@@ -182,7 +182,6 @@ function CommentForm({
         <div
           className="max-w-[95%] sm:max-w-[70%]"
           style={{
-            whiteSpace: 'normal',
             margin: 20,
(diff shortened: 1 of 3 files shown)


amadeus (approved):
good catch!

PR #1029, opened 2026-07-27
fix(diffs/edit): Realign cached rows after line-count edits

ije:
## Reproduction steps

1. Near the top, press `Enter` to type new lines
2. Scroll to the end of the document
3. Observe that the final content line is missing and an extra blank lines appears at EOF

## Why

Before commit be23fff6 (“Stop structural edits scanning to EOF”), inserting a newline accidentally caused background tokenization to rewrite every row through EOF.
That expensive full-tail scan unintentionally repaired the stale render cache. The optimization correctly stopped tokenizing once syntax state reconverged—often after one line—so the old cache bug became visible.
It was hidden when rendering reached EOF, editing at EOF, or never scrolling beyond the edited window.

Example:
```
Cached:    A B C D <blank EOF>
Insert X:  A B X C <blank EOF> <blank EOF>
                  ^ D was never shifted
```

The old applyDocumentChange only resized the array; it didn’t move unchanged cached rows to their new indexes.


packages/diffs/src/renderers/FileRenderer.ts
```diff
@@ -376,11 +376,46 @@ export class FileRenderer<LAnnotation = undefined> {
...
+      for (let i = 0; i < suffix; i++) {
+        result.code[nextLines.length - 1 - i] =
+          previousCode[previousLines.length - 1 - i];
+      }
+      // Tokenized rows beyond the old EOF already use post-edit indexes.
+      for (let i = previousLines.length; i < nextLines.length; i++) {
+        result.code[i] ??= previousCode[i];
+      }
+      for (let i = prefix; i < nextLines.length - suffix; i++) {
+        result.code[i] ??= {
```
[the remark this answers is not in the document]
ije: fixed [21ebe61](https://github.com/pierrecomputer/pierre/pull/1029/commits/21ebe61036dd1eeb3e9132f43a70bb1392c93a8d)

packages/diffs/src/components/File.ts
```diff
@@ -583,9 +583,15 @@ export class File<
 
   public updateRenderCache(
     dirtyLines: Map<number, Array<HighlightedToken>>,
-    themeType: 'dark' | 'light'
+    themeType: 'dark' | 'light',
+    _shouldRefreshView?: boolean,
```
amadeus: What's this prop for? Looks like its unused?

Is this to match FileDiff function signature? If so, wondering if we should just turn this into an object that we destructure instead?

ije: it's designed for requesting `FileDiff` view refresh, i renamed it to `shouldRefreshDiffsView` as an option.

```ts
updateRenderCache: (
    lines: Map<number, Array<HighlightedToken>>,
    themeType: 'dark' | 'light',
    options?: {
      /**
       * Whether to refresh the diffs view.
       * Deferred background-tokenize passes always pass false.
       */
      shouldRefreshDiffsView?: boolean;
      /**
       * Whether the line count has changed in flight.
       * True only during an edit pass whose line count changed,
       * deferred background-tokenize passes always pass false.
       */
      lineCountChangeInFlight?: boolean;
    }
)
```

PR #1033, opened 2026-07-28
Improvements to Edit API

amadeus:
So the idea of this PR is to improve the develop experience around editing.

Editing a file requires `useTokenTransform` enabled and I was not a fan of having that be a requirement that developers have to manage given it can really balloon dom node count and has a measurable highlighting impact.

However, after doing a bit more digging and thinking about it, I realized that all highlighting for edit already goes through the main thread highlighter, and that particular highlighter can easily take whatever customization props we need (vs WorkerPool which tries to save rendering options and has to re-highlight all visible files whenever a setting changes.

Because of this, I came up with the idea that we can stop requiring this prop to be set by the developer, and just silently send it through when in edit mode. If the file was previously highlighted without `useTokenTransform`, we can just kick off a new highlight using the main thread highlighter to rectify the DOM and then do our normal edit highlight infrastructure. Devs can still force the prop to be always on to avoid the initial re-highlight of the file on edit.

This does mean that while in edit mode on a file or diff, it's ALWAYS using the main thread highlighter, regardless. I think this is actually a good thing anyways, because we get into potentially weird territory with trying to allow the edit highlighter AND the worker pool at the same time, and we ensure they don't fight with each other.

So in summary: `useTokenTransform` becomes an internal edit mode responsibility. Devs don't have to think about it, and all the weird previous architecture that was forcing new options is no longer needed.

### DIFF
--- a/apps/docs/app/(diffs)/docs/Edit/constants.ts
+++ b/apps/docs/app/(diffs)/docs/Edit/constants.ts
@@ -9,8 +9,8 @@ const options = {
   unsafeCSS: CustomScrollbarCSS,
 } as const;
 
-// The editor requires the token transformer, so enabling it in the server
-// render keeps hydration from rerendering the surface after the editor
+// Enabling the token transformer in the server render keeps the markup
+// editor-ready, so hydration does not rerender the surface when the editor
 // attaches. Mirrors `(diffs)/_edit/constants.ts`.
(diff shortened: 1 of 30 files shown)


amadeus:
@codex review

ije (approved):
lgtm

PR #1038, opened 2026-07-31
[diffs/edit] refactor `onChange` editor option

ije:
Add `change` argument for the `onChange` editor option allow to get operation edits:

```ts
new Editor({
  onChange: (file, lineAnnotations, event) => {
    // [NEW]
    // 'event.changes' is an array containing all edits
    const changes = event.changes

    changes.forEach((change) => {
      console.log("Text inserted/replaced:", change.text)
      console.log("Range of the edit:", change.range) // { startLineNumber, startColumn, endLineNumber, endColumn }
      console.log("Offset of the change:", change.start, change.end)
    })
  }
})
```


packages/diffs/src/types.ts
```diff
@@ -1253,10 +1319,5 @@ export interface DiffsTextDocument {
  * emit them through its own `onItemEditChange` option.
  */
 export interface CodeViewCreateEditorOptions<LAnnotation> {
-  onChange: (
-    file: FileContents,
-    lineAnnotations?:
-      | LineAnnotation<LAnnotation>[]
-      | DiffLineAnnotation<LAnnotation>[]
-  ) => void;
+  onChange: (event: EditorChangeEvent<LAnnotation>) => void;
```
amadeus: Could we make this non-breaking by simply adding a 3rd `changes` argument?

amadeus: Or even better, what if we just passed the entire `event` through as the third argument? Then we expose everything and don't have to worry too much about potential future needs and it gives devs some flexibility?

packages/diffs/src/edit/index.ts
```diff
@@ -4,3 +4,9 @@ export type {
   PersistStateStorage,
 } from '../editor/stateStorage';
 export * from '../editor/textDocument';
+export type {
+  DiffsEditor,
+  EditorAttachEvent,
+  EditorChange,
+  EditorChangeEvent,
+} from '../types';
```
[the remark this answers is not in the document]
ije: no needs to expose `ResolvedTextEdit`

packages/diffs/src/types.ts
ije: ok, but let's refactor it in future update

ije:
@codex review again

PR #1041, opened 2026-07-31
[diffs/edit] fix editor state persisting

ije:
**Scroll position and selections are persisted in he Agent UI demo in home page.**

- Persist both `scrollTop`(been deleted in 517a4d17) and `scrollLeft` for `Virtualizer`.
- Keep `CodeView` persistence horizontal-only.
- Reset the viewport to `{ top: 0, left: 0 }` when persistence is enabled but no stored state exists.
- Restore persisted state for editable `FileDiff` surface as well, before only support `File` instance.

https://github.com/user-attachments/assets/7f868d85-ca07-4a1f-ac81-d56d67c040f8

**Fix safari scroll bug on unified diffs layout.**

https://github.com/user-attachments/assets/ae16a84b-72a7-4dbd-a515-d5c4d76771d6

### DIFF
--- a/packages/diffs/src/components/FileDiff.ts
+++ b/packages/diffs/src/components/FileDiff.ts
@@ -14,6 +14,7 @@ import {
   THEME_CSS_ATTRIBUTE,
   UNSAFE_CSS_ATTRIBUTE,
 } from '../constants';
+import { isSafari } from '../editor/platform';
 import {
   type GetHoveredLineResult,
   type GetLineIndexUtility,
@@ -2613,12 +2614,31 @@ export class FileDiff<
     const ast = this.hunksRenderer.renderCodeAST('unified', hunksResult);
     const gutterChildren = getElementChildren(ast?.[0]);
     const contentChildren = getElementChildren(ast?.[1]);
+
+    // Replacing both column subtrees within a single layout pass makes WebKit
+    // scroll the viewport on its own during that layout: while typing in a
+    // unified diff with collapsed regions, Safari yanked the editor toward the
+    // top of the render window on every debounced refresh. Forcing a layout
+    // after each column replacement keeps the passes separated, which
+    // empirically suppresses that scroll entirely; the snapshot below is a
+    // second line of defense that puts the viewport back if a WebKit build
+    // still moves it. Both are scoped to Safari so other engines never pay
+    // the extra layout nor risk clobbering a concurrent user scroll.
+    const restoreViewportScroll = isSafari();
+    const viewportScroll = restoreViewportScroll
+      ? this.getViewportScroll()
+      : undefined;
+
     for (const [el, astChildren] of [
       [columns.gutter, gutterChildren],
       [columns.content, contentChildren],
     ] as const) {
       if (astChildren != null) {
         el.innerHTML = toHtml(astChildren);
+        if (restoreViewportScroll) {
+          // force a layout(reflow)
+          void el.offsetHeight;
+        }
       }
     }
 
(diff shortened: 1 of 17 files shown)


packages/diffs/src/components/VirtualizedFile.ts
```diff
@@ -406,6 +406,27 @@ export class VirtualizedFile<
       : this.virtualizer.getContainerElement();
   }
 
+  public override getViewportScroll():
+    | { top: number; left: number }
+    | undefined {
+    const virtualizer = this.getSimpleVirtualizer();
+    return virtualizer != null
+      ? { top: virtualizer.getScrollTop(), left: virtualizer.getScrollLeft() }
```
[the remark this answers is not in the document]
amadeus: Yeah, this bot is getting a bit into the asymmetry I was worried about in that previous commit where I removed the viewport scroll. File and FileDiff don't really have a concept of a viewport, but the virtualized instances can reference a viewport. So fundamentally these feel like different types of data structures and I kinda wonder if they should be separate.

packages/diffs/src/components/FileDiff.ts
```diff
@@ -14,6 +14,7 @@ import {
   THEME_CSS_ATTRIBUTE,
   UNSAFE_CSS_ATTRIBUTE,
 } from '../constants';
+import { isSafari } from '../editor/platform';
```
amadeus, line 17: This feels like a check that has graduated to being relevant across all versions of the library. we should probably move this to the @pierre/diffs utils folder.

packages/diffs/src/components/FileDiff.ts
```diff
@@ -2610,12 +2614,31 @@ export class FileDiff<
...
+
     for (const [el, astChildren] of [
       [columns.gutter, gutterChildren],
       [columns.content, contentChildren],
     ] as const) {
       if (astChildren != null) {
         el.innerHTML = toHtml(astChildren);
+        if (restoreViewportScroll) {
+          // force a layout(reflow)
+          void el.offsetHeight;
```
amadeus, line 2640: Annoyingly this will really fuck with performance in Safari in CodeView, because all DOM writes and reads have been explicitly ordered to ensure we don't thrash.

So not sure I love this here. Is there a reason we need to call this after every file explicitly?

packages/diffs/src/components/VirtualizedFile.ts
```diff
@@ -406,6 +406,27 @@ export class VirtualizedFile<
...
+      : super.getViewportScroll();
+  }
+
+  public override setViewportScroll(position: {
```
amadeus, line 418: I'm worried that this API will cause problems when the CodeView is attempting to control scroll position via an animated scroll and fundamentally it creates an inverse relationship where the Files/Diffs themselves are attempting to control scroll position when they shouldn't have any knowledge of this.

packages/diffs/src/components/Virtualizer.ts
```diff
@@ -37,6 +37,7 @@ export interface VirtualizerConfig {
 
 export interface VirtualizerScrollOptions {
   top: number;
+  left?: number;
```
amadeus, line 40: Virtualizer doesn't support a horizontal scroll, so not sure we need this?

And while individual files can have horizontal scroll -- that would be a `left` value that ties to a specific file and wouldn't really be applicable to a viewport API like this.

amadeus (commented):
Some initial thoughts, I'm not sure this is the right way to fix this since it has the potential to vastly worsen WebKit's CodeView performance in general, and attempts to fix by hacking a scroll fix in the wrong direction (bottom up).

PR #1042, opened 2026-08-01
[diffs/edit] Misc Fixes

amadeus:
This PR fixes a bunch of scroll-jumping bugs that showed up while editing diffs, mostly in Safari.

- **Safari: the view jumped to the top while editing.** Pressing Enter (or expanding hidden lines) rebuilds the diff's DOM, and a WebKit bug makes the scroll position clamp to the top whenever that happens (related: webkit.org/b/308027). We now pin the diff's height while rebuilding so the scroll position can't get lost. Safari-only, edit-mode-only; virtualized views aren't affected by the bug so they skip the guard.
- **iOS Safari: creating a new line did an annoying animated focus scroll.** Rebuilds used to throw away the editable element and create a fresh one, which iOS treats as "you started editing again" and animates the caret back into view every time. Rebuilds now reuse the existing elements and just swap out their contents, so the editing session never restarts. This also fixed a double scroll-jump on desktop for the same edits.
- **The caret could end up floating way off the text on wrapped lines.** If the editor measured line wrapping at the wrong moment (mid-rebuild), it concluded "this line doesn't wrap", cached that, and kept placing the caret at the wrong spot — even forcing the code area to scroll sideways. Measurements taken in that broken state are no longer cached, so the next render self-corrects.
- **Small refactor:** browser detection helpers (`isSafari` etc.) moved into shared utils so non-editor code can use them too.
- **Tests:** updated the tests that relied on the old "rebuild = new element" behavior, added coverage that the editable element survives edits, and corrected one expectation that flipped now that focus never gets torn down in the first place.
- Made a few smol improvements to the homepage edit demo
  - Tweaked the trees horizontal padding to match vertical padding
  - Made file tree column narrower, felt a bit too wide for the content it was showing.

### DIFF
--- a/apps/docs/app/(diffs)/_home/agent-ui.css
+++ b/apps/docs/app/(diffs)/_home/agent-ui.css
@@ -151,7 +151,7 @@ button.aui-dot.is-zoom:hover {
 
 .aui-body {
   display: grid;
-  grid-template-columns: minmax(0, 1fr) 380px;
+  grid-template-columns: minmax(0, 1fr) 340px;
   min-height: 0;
 }
 
@@ -433,6 +433,7 @@ button.aui-dot.is-zoom:hover {
   min-height: 0;
   overflow: auto;
   padding-top: 6px;
+  --trees-padding-inline-override: 6px;
 }
 
 /* --- Fullscreen variant (/edit/live) ------------------------------------- */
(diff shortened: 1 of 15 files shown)


amadeus:
@codex review

Issue #1030, opened 2026-07-27
Bug: `CodeView` markdown remount scroll position jumping


aagrawal05:
### Prerequisites

- [x] I have [searched](https://github.com/pierrecomputer/pierre/issues?q=is%3Aissue) for duplicate or closed issues
- [x] I have read the [contributing guidelines](https://github.com/pierrecomputer/pierre/blob/main/.github/CONTRIBUTING.md)

### Describe the issue

When viewing diffs with richly rendered markdown, scrolling to bottom then scrolling back up, markdown stops rendering(?), the scroll position is lost jumping upwards to compensate for the change in scroll height, then markdown re-renders(?) — video shows this effect much more clearly.

### Reproduction

Can reproduce in [codiff](https://github.com/nkzw-tech/codiff) v1.9.1 very regularly with large md/source changes 

Preliminary AI audit of the DOM:
`CodeView` loses the scroll position when a tall asynchronous line annotation leaves the rendered window and remounts. In the reproduction, the annotation was initially measured at 1,395px, but its 180px loading placeholder was committed during remount. This reduced `scrollHeight` from 4,745px to 3,530px and moved scrollTop from 1,301px to 86px. The full height returned after loading, but the original scroll position did not.

https://github.com/user-attachments/assets/78fa4276-00fd-4b42-a7ba-1e301b4a8f55

### What browser(s) are you seeing the problem on?

Electron/Chromium

### What version of @pierre/diffs are you using?

@pierre/diffs@1.3.0-beta.9

aagrawal05:
More details of the AI traced repro at #1025 (apologies for the slop issue).
Repro'd on v1.9.1 and checkout of v1.9.2 of Codiff

Let me know if you'd like the exact same diff from the video/trace to repro on, though it should be pretty easy to replicate. Was encountering it often on docs related changes in code-bases.

amadeus:
Thanks for the new issue! In the process of getting `edit` released, so once that's out I'll poke around on this.

amadeus:
Alright, so I spent some time today digging into this, and I have a couple things to share about it, but the core issue here I am pretty confident is something you guys need to fix on the codiffs side, and a fix here would be untenable. That said, I did find an opportunity to strengthen a few things on our end to help alleviate some small occasional scroll jitters, so ultimately, thanks for reporting! (see #1054)

Oki, so to dig into it a bit more. 

[...]

PR #1047, opened 2026-08-03
fix(diffs): keep scroll anchor stable when CodeView removes items

fdarian:
> This PR is fully authored by Opus 5 and Sonnet 5 using Claude Code harness.
> <details>
> <summary>Here's a message from Opus regarding this change</summary>
> 
> `CodeView.reconcileItems` reassigns `this.items` (and rebuilds `idToItem`)
> without first calling `capturePendingLayoutAnchor()`, unlike `setOptions`,
> which always captures one before a layout-affecting change. The next render's
> `getScrollAnchor()` call has nothing pending, so it scans
> `renderState.firstIndex..lastIndex` against the array `reconcileItems` just
> reindexed. Every index at or after a removal now points at a different
> logical item, read through `.top`/`.height` that `recomputeLayout` hasn't
> refreshed yet, so the search either returns the wrong item or, once the
> window is far enough into stale territory, finds nothing at all and leaves
> `scrollTop` stuck at its old value while the content underneath it has moved.
> 
> Measured against a 355-file list, removing one file from the middle:
> 
> - Unpatched: `scrollTop` stays at `400000` while content shifts underneath
>   it, landing the viewport on an unrelated item; `scrollHeight` drops from
>   `738888` to `735704`.
> - Patched: `scrollTop` moves `400000` → `398256`, tracking the same anchor
>   line, with `scrollHeight` shrinking by exactly the removed item's height
>   plus gap in one update and no resettle.
> 
> I first hit this in [fdarian/nisi](https://github.com/fdarian/nisi), a diff
> review app built on `@pierre/diffs`, via its "hide reviewed files" action
> (removes a card from the rendered list) on a ~206-file list: `scrollTop`
> collapsed to `0`, `scrollHeight` dropped from `10088` to `502`, then
> resettled ~600-900ms later at `4956`, on unrelated files.
> 
> Same-length updates are unaffected; this only hits removals and reorders.
> 
> The fix is one line: call `capturePendingLayoutAnchor()` in
> `reconcileItems`, placed after the method's existing no-op early return
> (`if (firstDirtyIndex == null) return;`) and before `this.items = nextItems`.
> Placement matters: capturing unconditionally at the top of the method would
> strand a pending anchor on the instance after every no-op `setItems` call,
> since a same-length update that changes nothing returns early without
> rendering, and render is what consumes/clears the pending anchor. Some later,
> unrelated render would then "correct" scroll against that stale anchor. I hit
> exactly that while testing the fix, so it's a real trap and not hypothetical.
> 
> Added a regression test in `CodeView.scrollAnchoring.test.ts` that scrolls
> into a 60-item list, removes 10 earlier items in one `setItems` call, and
> asserts the anchored item's scroll offset shifts by exactly the removed
> items' combined height. It fails on `main` (scroll position stuck) and
> passes with the fix.
> </details>

Fixes a scroll-position bug in `CodeView`'s virtualized list: removing or reordering items in the controlled `items` array destroys the user's scroll position instead of keeping their place.

This one-line patch was discovered after Opus-5-observed Sonnet 5 did three iterations to solve the following bug. As shown in the video, removing an item will break the scroll position:

https://github.com/user-attachments/assets/2f63478d-6466-4496-9709-e78fa597f9d2

Then here's after it [added the patch](https://github.com/fdarian/nisi/commit/3e7f649348be06d08a996c882dc429b3aed91c11):

https://github.com/user-attachments/assets/b4839b94-2d6f-4844-baa0-d5e2ff95c20c

### Type of changes

- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would change existing functionality)
- [ ] Documentation update

### Checklist

### Related issues

Possibly related to #1030, which is the same symptom (scroll anchor failing
to recover) via a different trigger (markdown remount / transient estimated
heights) and a different code path; not claiming this fixes it.

### DIFF
--- a/packages/diffs/src/components/CodeView.ts
+++ b/packages/diffs/src/components/CodeView.ts
@@ -2535,6 +2535,18 @@ export class CodeView<LAnnotation = undefined> {
       return;
     }
 
+    // Capture the scroll anchor against the outgoing items/layout before
+    // reassigning this.items below. Otherwise the next render's
+    // getScrollAnchor() reads renderState.firstIndex/lastIndex against
+    // already-reindexed items through stale top/height metrics, which is
+    // exactly wrong for removals and reorders. This must stay below the
+    // early return above: capturing unconditionally at the top of the
+    // method would strand a pending anchor on the instance after a
+    // no-op reconcile, since render() (which consumes/clears it) never
+    // runs when nothing changed, and some later unrelated render would
+    // then "correct" scroll against that stale anchor.
+    this.capturePendingLayoutAnchor();
+
     this.items = nextItems;
     this.idToItem = nextIdToItem;
     this.instanceToItem = nextInstanceToItem;
--- a/packages/diffs/src/components/CodeView.ts
+++ b/packages/diffs/src/components/CodeView.ts
@@ -651,6 +651,11 @@ type PendingScrollTarget =
   | PendingRangeTarget
   | PendingItemTarget;
 
+type CodeViewItemMap<LAnnotation> = Map<
+  string,
+  CodeViewContextItem<LAnnotation>
+>;
+
 export class CodeView<LAnnotation = undefined> {
   static __STOP = false;
   static __lastScrollPosition = 0;
@@ -662,7 +667,7 @@ export class CodeView<LAnnotation = undefined> {
     resizeDebugging: false,
   };
   private items: CodeViewContextItem<LAnnotation>[] = [];
-  private idToItem: Map<string, CodeViewContextItem<LAnnotation>> = new Map();
+  private idToItem: CodeViewItemMap<LAnnotation> = new Map();
   private selectedLines: CodeViewLineSelection | null = null;
   // One editor per edit-mode item, created lazily via options.createEditor.
   // Entries survive virtualization unmounts so a remounted item re-attaches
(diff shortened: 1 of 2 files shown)


ije (approved):
LGTM

amadeus:
Hi! Yes, we should def fix this, but i think there's a few more areas of improvements. Your fix will only work if the top of your file is in the viewport so it can anchor to an earlier file. I'm take a few moments to try and see if i can get it to be more stable there and not cause a scroll jump if you're scrolled to the bottom of the file and remove it.

(the reason it will break is because you'll capture a pending anchor for the deleted file, and then when it attempts to rectify it wont exist and everything will be broked, so i'm looking into a way to fix pending anchor capture to not include removed files)

amadeus:
Oki, I added a commit that hardens this a bit more. 

So here's a before, that shows the existing fix, but also shows two other cases where the scroll anchoring breaks:

https://github.com/user-attachments/assets/126691f3-b089-4a14-8782-4f00c0e93174

Basically if you delete an item who's bottom is fully in view, the scroll anchoring will break. And if you remove an item who takes up the FULL viewport, then you'll just be arbitrarily somewhere else.

The fixes in action:

https://github.com/user-attachments/assets/d9fa0930-d8e6-48bb-a263-5dec97ebc0f9

Added a way to ensure we never anchor to an element that will be removed in the next render (this fixes the visible bottom item breaking scroll anchoring).

For items that fully take up the viewport, we just anchor the next item if it exists, to the top.

amadeus:
@codex review

PR #1049, opened 2026-08-03
feat(theme): Export palette and theme colors as Figma variables

mdo:
## Summary

`moonx theme:build` now also writes `packages/theme/figma/`, a Figma-importable copy of the same colors in [DTCG token format](https://www.designtokens.org/tr/2025.10/format/) — so designers can import the Pierre palette into Figma instead of copying hex values by hand.

Two collections, following Figma's [variable modes](https://help.figma.com/hc/en-us/articles/15343816063383-Modes-for-variables) import rules:

- **`figma/primitives.json`** — all 249 palette steps across the 21 scales, one mode. Tokens land as `blue/500`, `neutral/1040`, and so on.
- **`figma/semantic/*.json`** — 8 files, one per theme variant (light, dark, both `-soft`, and the four CVD variants). Figma turns each file into a **mode**, and every file carries the same 58 role tokens (`bg/editor`, `syntax/keyword`, `ansi/brightRed`, …).

Semantic values alias the primitives through the `com.figma.aliasData` extension, so editing a primitive in Figma updates every mode that uses it.

## Notes for reviewers

**Aliases carry a concrete color too.** Each semantic token has both a real hex `$value` and an alias, so a file still imports cleanly if the primitives collection is not present.

**Drift from the palette fails the build.** Aliases are resolved by matching role hexes back to palette steps, so a role value that is in no palette scale throws. That is deliberate — it catches roles drifting away from the palette. `#ffffff` (`bg.editor` and `accent.contrastOnAccent` in the four light variants) is the one intentional literal, allowlisted in `UNALIASED_ROLE_COLORS`.

**Why palette steps get sorted and serialized by hand.** JavaScript hoists integer-like object keys (`100`) ahead of leading-zero ones (`020`), so `Object.entries(neutral)` returns the lightest shades *last* and `JSON.stringify` preserves that. Since Figma creates variables in file order, the first draft listed every scale starting at `100` with `020`–`080` stranded at the bottom. Steps are now sorted numerically and groups serialized from `Map`s to hold that order.

**Vibrant variants are excluded.** Figma's importer supports only sRGB and HSL, and those themes use `color(display-p3 …)`.

**Packaging.** `figma/**` is added to `.vscodeignore` so the tokens do not ship inside the VSIX; npm was already covered by the `files` allowlist. The output is committed like `themes/*.json`, and `.oxfmtrc.json` ignores it the same way.

## Test plan

- [x] `moonx theme:build` — writes all 9 files
- [x] `moonx theme:test` — 376 pass, including 21 new in `test/figma.test.ts`
- [x] `moonx theme:typecheck`
- [x] `moon run root:format root:lint`
- [x] Deleting `figma/` and rebuilding reproduces byte-identical output
- [x] Mutation-checked the new tests: deleting a token from one mode and corrupting an alias target both fail with actionable messages
- [ ] **Needs a real Figma import.** The semantic files name `Pierre Primitives` as their alias target, so cross-collection linking only resolves if the collection is named exactly that (step 1 of the README instructions). Worth confirming aliases resolve and that the 8 files land as 8 modes.

### DIFF
--- a/.oxfmtrc.json
+++ b/.oxfmtrc.json
@@ -46,6 +46,7 @@
     "apps/docs/next-env.d.ts",
     "packages/theme/themes/*.json",
     "packages/theme/zed/themes/*.json",
+    "packages/theme/figma/**/*.json",
     "packages/tree-test-data/*-files.json",
     "*.patch",
     "README"
--- a/packages/theme/.vscodeignore
+++ b/packages/theme/.vscodeignore
@@ -6,6 +6,7 @@ preview/**
 .github/**
 .vscode/**
 zed/**
+figma/**
 
 ACCESSIBILITY.md
 CONTRIBUTING.md
--- a/packages/theme/CONTRIBUTING.md
+++ b/packages/theme/CONTRIBUTING.md
@@ -93,11 +93,25 @@ For visual proofing, `moonx theme:preview --ignore-ci-checks` writes
 `preview/*.html`: the palette scales, the Display-P3 conversions, and a
 normal-vs-simulated CVD proof sheet.
 
+## Figma variables
+
+`theme:build` also writes `figma/`, a Figma-importable copy of the same colors
+in DTCG token format (`src/createFigmaTokens.ts`). `figma/primitives.json` holds
+every palette step, and each `figma/semantic/*.json` is one theme variant that
+aliases those primitives — Figma turns each of those files into a mode. Commit
+the regenerated files alongside `themes/*.json`; see the README for the import
+steps.
+
+Semantic tokens are matched back to primitives by hex, so a role value that is
+not in any palette scale fails the build. That is deliberate: it catches roles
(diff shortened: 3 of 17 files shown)


amadeus:
CleanShot 2026-08-03 at 12 06 47

should we actually commit the built files?

mdo:
> should we actually commit the built files?

Just trying to pad my stats. And no, not likely. I'm playing with this in Figma for now, still not 100% solid yet.

amadeus:
word word!

mdo:
Okay this is pretty sweet. Works by importing base `Pierre Primitives` collection, then semantic themes use those tokens on import.

CleanShot 2026-08-03 at 12 32 07@2x

CleanShot 2026-08-03 at 12 32 16@2x

mdo:
hi mom

PR #1051, opened 2026-08-03
[chore] Update Shiki to 4.4.1

amadeus:
@SlexAxton had to break some min-age things, is this ok?


SlexAxton:
All good. Cc @evict

SlexAxton:
I occasionally go through and remove stuff that no longer needs an exception. In case you see it next week and wanna remove

amadeus:
Should i wait for Vincent to thumb up?

> I occasionally go through and remove stuff that no longer needs an exception. In case you see it next week and wanna remove

word, will do next week

SlexAxton:
Nah don’t need to wait

evict:
Thanks!

PR #1062, opened 2026-08-06
Update syntax of contrast-color() function on marker popovers, ensure e2e test works

mdo:


necolas:
@mdo This agent run wrote some helpers to fill the css feature gap in the test harness https://github.com/pierrecomputer/pierre/pull/1061. Might be worth incorporating?

mdo:
@necolas I'm down for either PR! Left a code comment on yours about the fallback intent here for the CSS vars.

necolas:
Yeah interesting that the model didn't infer that! I'd point your agent at the other branch, suggest it adds a similar fallback test that we keep while the pw chromium lacks support for that CSS. Not a blocker on merging this fix

necolas (approved):

amadeus (approved):
seems good to me! annoying lil syntax issue there. Not sure about the validity of the codex review though

mdo:
Will follow up with another iteration that addresses comments here.

PR #1070, opened 2026-08-10
[diffs/edit] tokenizer cache optimization

ije:
Previously, every edit discarded all cached bracket metadata from the edited line to the end of the document—even when tokenization proved that the downstream state was unchanged.

### Example

```ts
function foo() {
  return 1;
}
```

Changing `1` to `100` does not affect the TextMate state after that line. The highlighting below it remains valid, but the old implementation still deleted its cached string, comment, and regular-expression ranges.

A later bracket lookup on a distant line therefore had to invoke the grammar again to rebuild metadata that was still valid.

### New behavior

1. Re-tokenize the edited line.
2. Overwrite that line’s bracket metadata.
3. Compare its ending state with the cached state.
4. If the states match, preserve the unchanged downstream metadata.
5. If the state has not converged, invalidate from where background tokenization will continue.
6. For structural edits, invalidate from the edit because line indexes have shifted.

This is safe because unchanged text tokenized from an identical starting state produces identical bracket classifications.

### Benchmark

The benchmark primes a 100,000-line cache, edits line 0, and queries the final line.

| Metric | Before | After |
| --- | ---: | ---: |
| Grammar calls for the edit | 1 | 1 |
| Additional grammar calls for the final line | 1 | 0 |
| Median benchmark time | 0.198 ms | 0.104 ms |

The measured **47% improvement** applies specifically to this cache-invalidation scenario, not to overall tokenizer performance. The primary benefit is eliminating unnecessary grammar work after ordinary state-neutral edits.

### DIFF
--- a/packages/diffs/scripts/benchmarkEditorTokenizer.ts
+++ b/packages/diffs/scripts/benchmarkEditorTokenizer.ts
@@ -469,7 +469,7 @@ function createBenchmarkCases(
     {
       name: 'bracket-cache-invalidation',
       description:
-        'Edit one character at line 0 after caching bracket ranges for every line.',
+        'Edit one character at line 0, then query an unchanged cached line at the document tail.',
       run() {
         resetMessages();
         const counters = { grammarCalls: 0, setThemeCalls: 0 };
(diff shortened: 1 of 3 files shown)


ije:
@codex review

PR #1077, opened 2026-08-17
fix(diffs): ignore keydown during IME composition

Parassharmaa:
### Description

Skip editor and search panel keydown handling while an IME composition is open.

### Motivation & Context

Typing Japanese in the editor, `↓`/`↑` move the caret instead of walking the candidate list, `Enter` inserts a newline instead of confirming, and `Escape` runs `simplifySelection` instead of cancelling. The editor sets `#isComposing` but never reads it in keydown, so every key routes into `resolveEditorCommandFromKeyboardEvent`.

The search panel has the same gap. Its Search input has no guard at all. Its Replace input guards on `e.isComposing` alone, which is false when Safari fires `compositionend` before the keydown that ended the composition, so `keyCode === 229` is needed there too. That is also the case on the keydown that opens a composition, before `compositionstart`.

In the editor the guard sits after the two state resets, so a composing keydown keeps the same per-keydown state effects as any other and only dispatch is skipped.

Demo after the fix!

https://github.com/user-attachments/assets/2a0eb9f0-0cee-4009-bc8b-c9a8dd805beb

### Type of changes

### Checklist

- [x] I have read the contributing guidelines
- [x] My code follows the code style of the project
- [x] My code is formatted properly
- [x] I have added tests to cover my changes

4 tests added to `editorComposition.test.ts`. All 4 fail on `main` and pass here.

### DIFF
--- a/packages/diffs/src/editor/editor.ts
+++ b/packages/diffs/src/editor/editor.ts
@@ -1981,6 +1981,11 @@ export class Editor<LAnnotation> implements DiffsEditor<LAnnotation> {
         // typing, moving); let selectionchange sync #selections again.
         this.#suppressNativeSelectionSync = false;
 
+        // keyCode 229 opens the composition, before compositionstart fires.
+        if (e.isComposing || e.keyCode === 229) {
+          return;
+        }
+
         const command = resolveEditorCommandFromKeyboardEvent(
           e,
           this.#options.keymap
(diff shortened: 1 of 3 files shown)


necolas (approved):
lgtm, thanks!

ije (approved):
THANKS

PR #1086, opened 2026-08-19
fix(a11y): restore pinch-zoom on the docs sites

mdo:
## Summary
- The docs layout set `userScalable: false` in its `viewport` export, which fails Lighthouse's `meta-viewport` audit and blocks pinch-zoom for low-vision users on diffs.com and trees.software.
- Removing it re-enables pinch-zoom. This is a genuine accessibility defect independent of any ranking effect.
- DiffsHub deliberately keeps its locked viewport, so it is untouched.

## Test plan
- [ ] Pinch-zoom works on diffs.com and trees.software
- [ ] Lighthouse `meta-viewport` audit passes on the docs sites

Stacked on #1085 (3 of 4).

### DIFF
--- a/apps/docs/app/layout.tsx
+++ b/apps/docs/app/layout.tsx
@@ -61,7 +61,6 @@ const geistMono = Geist_Mono({
 export const viewport: Viewport = {
   width: 'device-width',
   initialScale: 1,
-  userScalable: false,
   themeColor: [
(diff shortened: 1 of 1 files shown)


necolas (approved):

amadeus:
you can def pinch to zoom on mobile safari, i assumed most browsers don't actually respect that flag. wild that we get docked for it.

Issue #1098, opened 2026-08-26
Bug: Editor selection highlights misaligned in split view with line wrap (Tauri WKWebView)


tywenk:
### Prerequisites

- [x] I have [searched](https://github.com/pierrecomputer/pierre/issues?q=is%3Aissue) for duplicate or closed issues
- [x] I have read the [contributing guidelines](https://github.com/pierrecomputer/pierre/blob/main/.github/CONTRIBUTING.md)

### Describe the issue

We have a Tauri app that runs on Macs and hooks into Webkit. We have a line selection bug in editable CodeView, specifically when in split view and with line wrapping turned on.

<img width="683" height="277" alt="Image" src="https://github.com/user-attachments/assets/02c7473d-b8d3-47be-bc04-7e8125fa95c6" />

The highlighted selection is not aligned with the line heights. Editing itself it not affected, this is just visual misalignment.

The following was generated by Claude, which can hopefully provide more details:
---

Environment

@pierre/diffs 1.3.6
Tauri (macOS WKWebView) — not Safari. Note the default WKWebView UA has no Safari token (Mozilla/5.0 (Macintosh; …) AppleWebKit/605.1.15 (KHTML, like Gecko)), so isSafari() returns false and any Safari-gated WebKit workarounds don’t engage in this embed.

Symptom

The editor’s selection highlight rects and caret don’t match the text. The overlay disagrees with WebKit’s own native selection: e.g. dragging a selection, getComposedRanges reports the anchor at char 9 while the rendered [data-selection-range] rects start at char 8 — the anchor appears to be latched from the first selectionchange after mousedown, before WebKit’s caret placement settles. We’ve also intermittently seen subsequent clicks on wrapped rows produce no selectionchange at all, leaving the caret stuck at its previous position.

Unified view and non-wrapped lines look fine; split + wrap is where lines actually wrap, which is where it shows.

Screenshots attached. Related: #1097.

### Reproduction

CodeView diff item with diffStyle: "split", overflow: "wrap", and edit: true, containing a long line that soft-wraps across several rows. Select text in the wrapped line with the mouse.

### What browser(s) are you seeing the problem on?

Other

### What version of @pierre/diffs are you using?

1.3.6

amadeus:
This is should be fixed in the next 1.4 beta release

PR #1101, opened 2026-08-27
[diffs/edit] Misc Webkit Bug Fixes

### LINKED ISSUE #1097: Bug: Edit caret renders one character behind at soft-wrap boundaries in WebKit
### Prerequisites

- [x] I have [searched](https://github.com/pierrecomputer/pierre/issues?q=is%3Aissue) for duplicate or closed issues

### Describe the issue

The following was generated with Claude 🤖 , but I (a human) was able to reproduce this:

---

## Summary

When using edit mode with `overflow: "wrap"` in WebKit/WKWebView, the visual caret can render exactly one character behind the document selection.

Typing still inserts text at the correct position, so this appears to affect only Pierre’s custom caret overlay.

Tested with `@pierre/diffs` 1.3.6.

## Reproduction

1. Render an editable file containing a long logical line.
2. Set `CodeView` to `overflow: "wrap"`.
3. Constrain the editor width so the line wraps.
4. Move the caret to the end of the logical line.

## Suspected cause

Wrap offsets are calculated by [`measureGrapheme`](https://github.com/pierrecomputer/pierre/blob/diffs-v1.3.6/packages/diffs/src/editor/editor.ts#L5894-L5898), which currently uses:

```ts
return range.getBoundingClientRect();
```

At a soft-wrap boundary, WebKit can return a bounding rectangle anchored to the previous visual row. This can cause the recorded wrap offset to be one grapheme late, so the segment-local caret measurement omits one character.

Using the final rectangle from `getClientRects()` preserves the rectangle belonging to the grapheme’s final visual row:

```diff
 const measureGrapheme = (index: number): DOMRect => {
   range.setStart(textNode, graphemeStart(index));
   range.setEnd(textNode, graphemeStart(index + 1));
-  return range.getBoundingClientRect();
+  const clientRects = range.getClientRects();
+  return (
+    clientRects.item(clientRects.length - 1) ??
+    range.getBoundingClientRect()
+  );
 };
```

We applied this patch locally and added a regression test that emulates the WebKit boundary geometry. Before the patch, the caret was displaced by exactly one glyph width; after the patch, its left edge aligns with the end of the final character.

This only reproduces when wrapping is enabled.

### Reproduction

For example:

```tsx
<EditProvider createEditor={createEditor}>
  <CodeView
    items={[
      {
        id: "file",
        type: "file",
        file: {
          name: "wrapped.txt",
          contents:
            "abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789",
        },
        edit: true,
        annotations: [],
        version: 0,
      },
    ]}
    editorOptions={editorOptions}
    options={{
      disableFileHeader: true,
      overflow: "wrap",
    }}
  />
</EditProvider>
```

Place the editor in a container around 240px wide, focus it, and move the caret to the document end.

## Actual behavior

The caret overlay appears one glyph behind the final character:

```text
WXYZ012345678|9
```

Typing still appends after `9`, confirming that the underlying selection position is correct.

## Expected behavior

The caret should render after the final character:

```text
WXYZ0123456789|
```

### What browser(s) are you seeing the problem on?

Safari

### What version of @pierre/diffs are you using?

1.3.6

amadeus:
Some smol Webkit specific fixes that I'm just adding on top of this larger edit workstream

Fixes #1097, #1098

### DIFF
--- a/packages/diffs/src/editor/editor.ts
+++ b/packages/diffs/src/editor/editor.ts
@@ -5966,7 +5966,13 @@ export class Editor<LAnnotation> implements DiffsEditor<LAnnotation> {
       const measureGrapheme = (index: number): DOMRect => {
         range.setStart(textNode, graphemeStart(index));
         range.setEnd(textNode, graphemeStart(index + 1));
-        return range.getBoundingClientRect();
+        // WebKit adds a zero-width fragment on the previous row at soft wraps;
+        // the final fragment belongs to the grapheme's rendered row.
+        const clientRects = range.getClientRects();
+        return (
+          clientRects.item(clientRects.length - 1) ??
+          range.getBoundingClientRect()
+        );
       };
 
       // A new visual line starts whenever a grapheme's top edge moves below
--- a/packages/diffs/src/editor/editor.ts
+++ b/packages/diffs/src/editor/editor.ts
@@ -2554,7 +2554,10 @@ export class Editor<LAnnotation> implements DiffsEditor<LAnnotation> {
   #computeContentOffset(contentEl: HTMLElement) {
     if (this.#isDiff && this.#diffSyle === 'split' && this.#isWrap) {
       this.#contentOffset = {
-        top: contentEl.offsetTop,
+        // #getLineY already includes the code column's block padding. Store
+        // only the grid displacement beyond it so split + wrap does not add
+        // the same top gap twice.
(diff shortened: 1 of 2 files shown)


ije (approved):
LGTM

amadeus:
@codex review

PR #1113, opened 2026-09-01
fix(diffs): avoid stack overflow when rendering massive diffs

Josh-Walker-GM:
### Description

Rendering a diff of a large file throws `RangeError: Maximum call stack size exceeded`, and the diff never appears. 

This occurred because the lines of a massive change were eventually splatted into a function. When the size is sufficiently large (V8 ~124k, JavaScriptCore ~639k) then this results in blowing the stack. Changing to calling the function repeated for each element fixes the issue and didn't appear to introduce any significant performance difference. 

A regression test that renders a single-hunk 700k-line diff, sized above the highest argument limit of the engines the package supports, has been added to demonstrate the issue and guard the fix.

### Motivation & Context

I observed this bug while using the product which motivated a fix for it. I believe I was visualising a diff of adding a new massive JSON file. 

### Type of changes

      being opened
      that this PR is being opened
- [ ] Documentation update

### Checklist

AI was used to debug my original issue and to assist in implementing the fix in the code. I have manually reviewed and refined the AI generated code. Claude usage is identified within the commit message too. 

### Related issues

I did not file an issue and proceeded straight to a PR with the fix.


amadeus:
Interesting find! A couple things

1) I'm not sure a test for this is necessary (given it's probably a lot of compute/work)
2) Are there aother places in the codebase that could also use a fix like this? I have fable 5.1 doing a bit of search at the moment.

Josh-Walker-GM:
Happy to ditch the test for sure. I mostly kept it because it's easy for someone to see a for loop like that and think "ah I can flatten that with a splat" but I guess as long as it's documented inline that's a bit of a non concern.

amadeus:
Fable came up with these potential spots:

| Site                                          | What is spread                                  | Verdict                                                                                                                                                                                        |
| --------------------------------------------- | ----------------------------------------------- | ---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
| `utils/trimPatchContext.ts:158`               | every line of one hunk                          | **Fix.** Same shape as the bug. A single massive hunk overflows. Public export.                                                                                                                |
| `utils/renderFileWithHighlighter.ts:104`      | all highlighted lines of a plain-text window    | **Fix for twin symmetry.** Virtualized windows are small, but `FileRenderer.render` and `asyncRender` accept any finite range, so a caller passing a large window on a massive file overflows. |
| `shiki-stream/tokenizer.ts:51`                | every token of every complete line in one chunk | **Caller-controlled.** Normal stream chunks are small. A consumer writing a whole large file as one chunk overflows. Public export.                                                            |
| `shiki-stream/tokenizer.ts:44`                | tokens of one line                              | Bounded per line.                                                                                                                                                                              |
| `editor/editor.ts:451`, `:3424`, `:3465`      | tokens of one line                              | Bounded. The editor tokenizer emits one token for lines over 1000 chars.                                                                                                                       |
| `editor/utils.ts:32`                          | `h()` children                                  | Bounded. All callers pass fixed small arrays or one line's tokens.                                                                                                                             |
| `utils/editSessionHunks.ts:588`               | blocks of a previous region plan                | Safe. The source plan's block list is always empty at merge time.                                                                                                                              |
| `utils/realignChangeContent.ts:47`            | replacement blocks                              | Bounded, at most 3.                                                                                                                                                                            |
| `components/FileDiff.ts:3115`, `:795`, `:798` | column elements, tuples                         | Bounded.                                                                                                                                                                                       |
| `editor/editor.ts:2064`                       | two listeners per annotation                    | Bounded by annotation count.                                                                                                                                                                   |
| `utils/trimPatchContext.ts:146`               | pending context lines                           | Bounded to twice the context size by the hunk-split logic.                                                                                                                                     |

amadeus:
This makes me wonder if we should just define your helper function in utils, then share it around all these scenarios.

Josh-Walker-GM:
You can take this over and solve these additional similar findings if you like. Whatever you'd prefer of course.

amadeus:
@codex review

amadeus:
Thanks for finding/reporting/opening this PR!

cannikin:
I work with @Josh-Walker-GM and we found this in our project, it'a amazing how fast this got merged, thank you! Any idea when it will be released? 👀

amadeus:
I'm in the final stages of prepping for the 1.4 beta release which would include this, best case later today, worst case by the end of the week.

PR #1118, opened 2026-09-05
[diffs] Virtualization FileDiff Fix

### LINKED ISSUE #1115: VirtualizedFileDiff.render: rendered a different diff than its prepared layout
When using Worker Pool + Virtualizer + FileDiff, replacing `fileDiff` with a new object results in an error:

```
Error: VirtualizedFileDiff.render: rendered a different diff than its prepared layout
    at VirtualizedFileDiff.render
    at VirtualizedFileDiff.onRender
    at render
```

In that scenario it's caught internally and printed to the console, but there are scenarios with edit involved where the same exception crashes the page altogether.

Minimal reproduction: [diffs-crash.tar.gz](https://github.com/user-attachments/files/31835403/diffs-crash.tar.gz)

amadeus:
Fixes #1115

### DIFF
--- a/packages/diffs/src/components/VirtualizedFile.ts
+++ b/packages/diffs/src/components/VirtualizedFile.ts
@@ -766,6 +766,11 @@ export class VirtualizedFile<
       latestFile == null
         ? undefined
         : this.fileRenderer.getFileForNextRender(latestFile);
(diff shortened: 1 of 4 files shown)


amadeus:
@codex review

PR #1119, opened 2026-09-07
Adds TSRX extension support

trueadm:
Adds [TSRX](https://github.com/tsrx-org/tsrx) extension support.

Verified with the Diffs test suite, typecheck, lint, and a built-package smoke test.

AI assistance: Codex implemented and tested this change.

### DIFF
--- a/packages/diffs/src/highlighter/languages/grammars/tsrx.tmLanguage.ts
+++ b/packages/diffs/src/highlighter/languages/grammars/tsrx.tmLanguage.ts
@@ -1,0 +1,7433 @@
+import type { LanguageRegistration } from 'shiki';
+
+// Vendored from tsrx-org/tsrx. See the package NOTICE.md for provenance and licenses.
+const grammar: LanguageRegistration & {
+  information_for_contributors: string[];
+  version: string;
+} = {
+  information_for_contributors: [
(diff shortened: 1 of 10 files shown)


packages/diffs/src/highlighter/languages/grammars/tsrx.tmLanguage.ts
ije, line 1: @trueadm  I wonder if you could contribute the grammar to https://github.com/shikijs/textmate-grammars-themes/tree/main/packages/tm-grammars, so that all Shiki users can benefit from it include diffs. Ideally, no language-specific grammars should be added directly to the package.

We’re also working on a fast highlighter that will include TSRX as a built-in grammar!

ije: Screenshot 2026-09-08 at 14 01 24

^preview

amadeus: Yeah I don’t think we want to bundle languages with diffs. Best is to upstream them with shiki

trueadm: Gotcha, thanks! Looking forward to playing around with the highlighter too! :)

PR #1121, opened 2026-09-09
Preserve native ranges when dragging selections upward

ije:
Preserve browser-native word and whole-line selections while dragging in either direction.

fix https://x.com/kebabyusuf/status/2097590039106490725?s=20


ije:
@codex review

